PR #1283, opened 2018-03-25
Fix the order of setup procedure

akihikodaki:
ATTENTION: I license my changes in AGPL-3.0. If you merge this, the users of this software will be required to conform to the license when modifying, redistributing, and hosting.


syuilo:
Thank you!

PR #1284, opened 2018-03-25
Pass hostname instead of host to vhost module

akihikodaki:


syuilo:
Thank you for your refactoring

akihikodaki:
It is not a refactoring. It is a fix. `vhost` accepts hostname but not host.

syuilo:
naruhodo

PR #1292, opened 2018-03-26
Define hostname and secondary_hostname in config

akihikodaki:
This is missing part of commit 0109e0811c4142153ae3f915295e62630653909e.


syuilo:
👍

PR #1313, opened 2018-03-27
Replace /:user endpoints with /@:user

### LINKED ISSUE #1310: Page for user named "search" is not accessible
Page for user named `search` is not accessible because the route collides to search endpoint. Possible solutions are:
1. Block `search`, and other names which collides to existing endpoints and, if we are going to create a new endpoint and it collides to an existing user, rename his username.
2. Prefix user endpoints with `@`.

akihikodaki:
Resolves #1310.


syuilo:
Thank you <3

PR #1384, opened 2018-04-03
Remove resolveRemoteUserObjects method of remote/activitypub/resolver

akihikodaki:
The value of the value returned by resolveRemoteUserObjects method of
remote/activitypub/resolver was inconsistent.


syuilo:
Thank you👍

PR #1420, opened 2018-04-08
Fix banner selection from user page and drives

unarist:
Previously, updateBanner takes a callback as a first argument of the function, but some calls passes a file object.

I've refactored it with Promise to avoid those misuses (although it should have type definitions :/)


syuilo:
Cool😎
ありがとうございます！

PR #1476, opened 2018-04-15
Specify ImageMagick version (7.0+)

Angristan:
Misskey needs the `magick` command, which is not available with ImageMagick <7

screenshot_15-04-2018_12-58-37


angristan:
https://github.com/syuilo/misskey/blob/37edacce44cb7d63c35fee88e0b43028a1c7765c/src/utils/dependencyInfo.ts#L14

syuilo:
Thank you!

PR #1491, opened 2018-04-16
fixed translate.en.md

mirro-san:
:heart:


TheKinrar:
@Conan-kun Would be great to not speak in French here, considering only a few people will understand you.

@mirro-san Your changes in section 3 were clearly wrong and had no actual sense. Please pay attention to what you write.

PR #1492, opened 2018-04-16
Fix translate.en.md

TheKinrar:
#1491 introduced some errors in translate.en.md.

This reverts some of its changes and tries to improve the original translation.


syuilo:
Thank you!

PR #1593, opened 2018-05-17
moar i18n

m4sk1n:
i'm testing it rn


mkljczk:
(will translate to English and Polish via crowdin)

PR #1685, opened 2018-06-08
fix: when text is null, bug can pass validation.

2vg:
fixed. (maybe?)


syuilo:
JavaScriptに or ってあったっけ

2vg:
@syuilo ごめん、ない！ｗ

syuilo:
Thank you!

PR #1892, opened 2018-07-13
Fix a bug

PRB0t:
Hi, I'm [@nasukh@misskey.xyz](https://misskey.xyz/@nasukh)'s owner. This is the bug which I used for the attack.
--
Automated submit by PRB0t


syuilo:
Thank you!
I made this fix, but I forgot to commit.

PR #1909, opened 2018-07-14
Update docs

skid9000:
I updated the setup docs to add some things and translated the management docs in English.

I don't speak/write Japanese so if anyone can translate my edits from the setup docs in jap, that would be nice ^^

### DIFF
--- a/docs/manage.en.md
+++ b/docs/manage.en.md
@@ -1,0 +1,13 @@
+# Management guide
+
+## Check the status of the job queue
+In the directory of Misskey:
+``` shell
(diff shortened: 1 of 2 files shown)


AyaMorisawa (approved):

AyaMorisawa:
> I don't speak/write Japanese so if anyone can translate my edits from the setup docs in jap, that would be nice ^^

I updated Japanese setup docs to reflect your changes: https://github.com/syuilo/misskey/commit/7276ec185b5e51c0d12ae8d6e999f99127ae8338
Thanks for suggestion!

PR #2008, opened 2018-07-27
memory usage excludes buffer and cache.

2vg:
ビルドしていないので要確認


syuilo:
Thanks!

src/daemons/server-stats.ts
```diff
@@ -1,5 +1,5 @@
 import * as os from 'os';
-const osUtils = require('os-utils');
+const sysUtils = require('systeminformation');
```
acid-chicken, line 2: I'm feeling the below code is better than that.
```js
import * as sysUtils from 'systeminformation';
```

src/daemons/server-stats.ts
```diff
@@ -18,25 +18,60 @@ export default function() {
...
+// TOTAL MEMORY STAT
+async function totalMem() {
+  try {
+    const data = await sysUtils.mem();
+    return data.total;
+  }
+  catch(error) {
+    console.error(error);
+  }
+}
```
acid-chicken, line 77: Please add newline here.
https://github.com/syuilo/misskey/blob/933255179151ca0ee6ea7fcc20299e54b2c1d0a2/.editorconfig#L7

src/daemons/server-stats.ts
AyaMorisawa: Done

src/daemons/server-stats.ts
AyaMorisawa: Done

PR #2053, opened 2018-08-01
Translated the totality of desktop views and common views, a great bunch of mobile views

ghost:


syuilo:
Wow, Thanks a lot @gutfuckllc!! 👍

PR #2509, opened 2018-08-27
Update update-avatar.ts to match update-banner.ts behaviour

Hiramiya:
This PR updates update-avatar.ts to match update-banner.ts behaviour, fixes the drive selection bug syuilo/misskey#2505


syuilo:
Thank you for your contribution!

PR #2511, opened 2018-08-27
Prevent setting non-image files as avatar/banner

Hiramiya:
Depends on/includes previous PR: syuilo/misskey/pull/2509

This PR prevents using file extensions other than commonly used image extensions as an avatar or banner. This does not stop files with incorrect extensions, but does prevent accidental changes.

Submitting as a fix for syuilo/misskey/issues/2096, however will require additional localisation

### DIFF
--- a/locales/en-US.yml
+++ b/locales/en-US.yml
@@ -411,6 +411,7 @@ desktop:
   uploading-avatar: "Uploading a new avatar"
   avatar-updated: "Successfully updated the avatar"
   choose-avatar: "Select an image for the avatar"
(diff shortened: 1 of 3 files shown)


syuilo:
👍👍👍

Hiramiya:
@syuilo can you add
```
desktop:
  invalid-filetype:
```
to CrowdIn? Originally added it directly in a [commit](https://github.com/syuilo/misskey/pull/2511/commits/14ad3af30d36dbc28d8e85973144aaa6f36efe77)

Thanks!

PR #2550, opened 2018-08-30
Fix broken invalid-filetype behaviour

Hiramiya:
After being merged in with other changes, my previous PR to prevent using non-images as Avatar/Banner is not working :cry: 

This PR uses the correct `return reject()` syntax instead, and works on my instance :+1:


syuilo:
👍

Issue #2829, opened 2018-10-06
Home/social timelines are broken in my server


ghost:
# Summary
I have set up my server at https://6.distsn.org, but its home/social timelines are broken.

# Expected Behavior
Home and social timelines are shown.

# Actual Behavior
"読み込みに失敗しました" is shown at the home and social timelines.

# Steps to Reproduce
1. Setup my server
2. Login

# Environment
* Server: Ubuntu 16, Misskey 9.7.1 release
* Client: Desktop Firefox

ghost:
This is the screenshot of my browser and its developer tool. `TypeError: t.announcements is undefined` is related to this issue?

![screenshot from 2018-10-06 17-00-53](https://user-images.githubusercontent.com/2155555/46569058-aa062780-c98a-11e8-9924-0269ed721d28.png)

mei23:
What's version mongoDB use ?
On MongoDB <3.6, home and social timeline may break.

ghost:
Oh...
```
MongoDB shell version: 2.6.10
connecting to: test
> db.version()
2.6.10
```

ghost:
https://docs.mongodb.com/manual/tutorial/install-mongodb-on-ubuntu/

ghost:
Thanks a lot! (Now its https://7.distsn.org)

PR #2837, opened 2018-10-06
#2829 Improve setup guide

ghost:


syuilo:
see https://github.com/syuilo/misskey/pull/2836#issuecomment-427619891

PR #3065, opened 2018-11-01
Change the language of instance to und by default

ghost:
This PR is changing the language of instance from *ja* to *und* by default.

The language code *und* is defined as *undefined* in the ISO 639-3 standard.

In the present implementation, if foreign admins do not set the language of their instance, the instance reports its language as Japanese. It is better that they report *undefined* or somehow.

### DIFF
--- a/src/server/api/mastodon.ts
+++ b/src/server/api/mastodon.ts
@@ -51,7 +51,7 @@ router.get('/v1/instance', async ctx => { // TODO: This is a temporary implement
 			status_count: originalNotesCount,
 			domain_count: domains.length
 		},
-		languages: config.languages || [ 'ja' ],
+		languages: config.languages || [ 'und' ],
 		contact_account: {
 			id: maintainer._id,
 			username: maintainer.username,


acid-chicken (changes requested):
Please use ISO 639-1.
FYI: https://docs.joinmastodon.org/api/entities/#instance

AyaMorisawa:
FYI: https://github.com/tootsuite/mastodon/issues/2130

ghost:
https://github.com/tootsuite/mastodon/issues/2130 is interesting, but Mastodon's default language detection is completely broken. Mastodon reads `DEFAULT_LOCALE` environment variable first, but usually `DEFAULT_LOCALE` is empty, and it is fall-backed into `en`. Finally, the language `en` does not means *we speak English in this instance*, but *the admin of this instance does not care about what language the users speak*. For Misskey, respectively, the default language *ja* indicates the *nothing*.

ghost:
The Mastodon's document just say *ISO6391*. Dose it means in our common sense only ISO-6391-1, not the -2 nor -3?

https://docs.joinmastodon.org/api/entities/#instance

acid-chicken:
> ISO 6391:1997
> Meat and meat products -- Enumeration of Escherichia coli -- Colony-count technique at 44 degrees C using membranes
> &mdash; https://www.iso.org/standard/24601.html

@hakaba-hitoyo Do you think that the Mastodon wants to mean this ISO? I don't think so.

AyaMorisawa:
@syuilo

ghost:
Now I see that the Mastodon's document says *ISO6391*, it should be *ISO-639-1*.

marihachi:
Probably, should set directly usable language for default.(und is special language code.)

ghost:
> Probably, should set directly usable language for default.(und is special language code.)

I am afraid that reporting the default code *ja* means just *the admin does not care about what language is spoken in this instance*, not *there is an instance for Japanese speakers here*. Usually what we need seems to be the latter.

marihachi:
As a premise, need to config that can set default language for misskey instance. (I could not find it...)
And should send it as default language.
Until then, I think it should be send `en`.

syuilo:
I just implemented language settings with GUI.
https://github.com/syuilo/misskey/commit/001bb7bbcd688510223b4f355057cf73d264a7c4

ghost:
https://github.com/syuilo/misskey/commit/001bb7bbcd688510223b4f355057cf73d264a7c4 を見るかぎりデフォルトは空配列になったようだ。

PR #3118, opened 2018-11-05
Make /api/v1/instance and /api/v1/custom_emojis better

acid-chicken:
***Depends #3117***
Closes #3066
Related to #1280 and #3116
From dca110ebaa78f64600429f812c238a07d2f1dc1d


syuilo:
👍👍👍

PR #3266, opened 2018-11-16
Add space.

incognico:
For some reason the missing space bothers me very much. Maybe a `<br>` would work too..


acid-chicken (approved):

syuilo:
Since this depends on the language, it seems better to improve that i18n. (I will do it)
Thanks anyway.

PR #3383, opened 2018-11-22
Fix #3343

### LINKED ISSUE #3343: 3rd party sign in broken because of secure cookie
Sign in with GitHub/Discord (and probably Twitter) not working on my configuration (Misskey http <-> apache proxy https <-> client https)

`Error: Cannot send secure cookie over unencrypted connection` when trying to access /api/signin/... (client: Internal Server Error)

Related #2923

incognico:
Setting "X-Forwarded-Proto: https" in the SSL proxy is the correct way to do this


syuilo:
Thank you!

PR #3472, opened 2018-12-02
Use if-statement instead of switch-statement

AyaMorisawa:


acid-chicken (commented):
I think that keeping to use switch statement in the codes of ActivityPub endpoints is good for our scalability, however, others are OK.

syuilo:
現状を保ちたいと思います
Thanks anyway.

PR #3697, opened 2018-12-20
Fix overlap of birthday label on datepicker

### LINKED ISSUE #3696: "Birthday" Overlaps the Date Picker
# Summary

image
the label overlaps the date picker

# Expected Behavior

Both are readable

# Environment

firefox, macos

ibrokemypie:
# Summary

Fix #3696

Adds a new "title" slot that works like the expanded label.
This means birthday will always be expanded out with no animation, so it cannot overlap the date picker
image


syuilo:
Great! Thank you 🙏🙏🙏

PR #3728, opened 2018-12-23
Fix mention links

l2dy:
# Summary
`canonical` already starts with `@`, remove the extra `@`.

```ts
computed: {
	canonical(): string {
		return `@${this.username}@${toUnicode(this.host)}`;
	},
```


syuilo:
Thank you!

PR #3732, opened 2018-12-23
Change italic syntax to fit with other markdown implementations

ibrokemypie:
# Summary

currently italics require an html <i> syntax, unlike every other formatter. this is very confusing, especially as there doesnt seem to be any documentation.
this change makes both _text_ and *text* work, both things that users would expect from other implementations of markdown such as pleroma, discord and github.

see:
https://github.com/adam-p/markdown-here/wiki/Markdown-Cheatsheet#emphasis
https://sourceforge.net/p/pleroma/wiki/markdown_syntax/#md_ex_text
https://support.discordapp.com/hc/en-us/articles/210298617-Markdown-Text-101-Chat-Formatting-Bold-Italic-Underline-

### DIFF
--- a/src/mfm/parser.ts
+++ b/src/mfm/parser.ts
@@ -224,7 +224,7 @@ const mfm = P.createLanguage({
 
 	//#region Italic
 	italic: r =>
-		P.regexp(/<i>([\s\S]+?)<\/i>/, 1)
+		P.regexp(/(\*|_)([\s\S]+?)\1/, 2)
 		.map(x => createTree('italic', P.alt(
 			r.bold,
 			r.strike,


AyaMorisawa (approved):

acid-chicken (changes requested):
Some of invalid syntaxes (e.g. `*italic_`) are allowed by this regex.

ibrokemypie:
Here we go, now it requires that the second match group be the same match as the first, preventing _word* etc
@acid-chicken

acid-chicken:
Oops! 

source: `*test* yay _foo_ *bar_ _wow*`
result: _*_ yay \_ _*_

image

AyaMorisawa:
* [ ] some test cases in ./test/mfm.ts should be changed or added

ibrokemypie:
How should I go about doing the test cases? Fixed the issue mentioned (needed to change ``P.regexp(/(\*|_)([\s\S]+?)\1/, 1)`` to ``P.regexp(/(\*|_)([\s\S]+?)\1/, 2)`` because there are two capture groups)
and the output is as expected:
image

I do not really understand how the tests work though, so don't know how to write such a test.

ibrokemypie:
Perhaps a toggle that defaults to off?

PR #3741, opened 2018-12-25
Make activity view in admin scrollable

ibrokemypie:
# Summary
I can't think of any reason why this isnt currently the case as the extra lines are still there, just not displayed, meaning theres no difference in performance/memory usage
Also means that sometimes entries are cut off which is weird
Also, sometimes there is reason to wish to view older entries that may have scrolled off the page/keep reading something which was pushed too far down.


syuilo:
👍

PR #3748, opened 2018-12-25
Add URI parameter to /api/meta

### LINKED ISSUE #3747: Return Uri With /api/meta
# Summary

Currently api/meta returns a bunch of info, but not the uri
eg, "misskey.xyz"

ibrokemypie:
# Summary
closes #3747

### DIFF
--- a/src/misc/fetch-meta.ts
+++ b/src/misc/fetch-meta.ts
@@ -2,6 +2,7 @@ import Meta, { IMeta } from '../models/meta';
 
 const defaultMeta: any = {
 	name: 'Misskey',
+	uri: '',
 	maintainer: {},
 	langs: [],
 	cacheRemoteFiles: true,
--- a/src/misc/fetch-meta.ts
+++ b/src/misc/fetch-meta.ts
@@ -2,7 +2,6 @@ import Meta, { IMeta } from '../models/meta';
 
 const defaultMeta: any = {
 	name: 'Misskey',
-	uri: '',
 	maintainer: {},
 	langs: [],
 	cacheRemoteFiles: true,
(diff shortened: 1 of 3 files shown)


src/models/meta.ts
```diff
@@ -30,6 +30,17 @@ if ((config as any).description) {
...
+	Meta.findOne({}).then(m => {
+		if (m != null && m.uri == null) {
+			Meta.update({}, {
+				$set: {
+					uri: (config as any).url
+				}
+			});
+		}
+	});
+}
```
mei23: Probably this section (L33-L43) is unnecessary.
This section to be used only by settings that need to be migrated from file to DB.

src/server/api/endpoints/meta.ts
```diff
@@ -42,6 +42,7 @@ export default define(meta, (ps, me) => new Promise(async (res, rej) => {
 		clientVersion: client.version,
 
 		name: instance.name,
+		uri: instance.uri,
```
mei23: So you can now directly pick from config file, like a `uri: config.url`.

mei23 (approved):

src/misc/fetch-meta.ts
```diff
@@ -2,6 +2,7 @@ import Meta, { IMeta } from '../models/meta';
 
 const defaultMeta: any = {
 	name: 'Misskey',
+	uri: '',
```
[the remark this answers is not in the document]
ibrokemypie: What else could it be?
It should never be able to fall back to this default because of https://github.com/syuilo/misskey/blob/8f714b5b126226346af850337566f0f70bd02d4f/src/config/load.ts#L30

config.url should always exist

syuilo:
I modified it more simply. Please review again.

AyaMorisawa (approved):

acid-chicken (approved):

syuilo:
🙏

PR #3795, opened 2018-12-29
Add id to return of users/relation

ibrokemypie:
# Summary
Adds the user id of the relevant user to the return of /api/users/relation, making it much easier to deal with the response programmatically or asynchronously


acid-chicken (approved):
:shipit:

syuilo:
Thank you!

PR #3916, opened 2019-01-18
Change italic syntax to fit with other markdown implementations

ibrokemypie:
currently italics require an html ``<i>`` syntax, unlike every other formatter. this is very confusing, especially as there doesnt seem to be any documentation.
this change makes both ``_text_`` and ``*text*`` work, both things that users would expect from other implementations of markdown such as pleroma, discord and github.

I changed the regex to require the first character in the match to be alphanumeric, addressing the issue of the syntax catching ascii emotes

also added testcases, though I cant seem to get the testing working in genneral on my system, any tips would be appreciated

### DIFF
--- a/src/mfm/parser.ts
+++ b/src/mfm/parser.ts
@@ -224,7 +224,7 @@ const mfm = P.createLanguage({
 
 	//#region Italic
 	italic: r =>
-		P.regexp(/<i>([\s\S]+?)<\/i>/, 1)
+		P.regexp(/(\*|_)([a-zA-Z0-9]+?[\s\S]*?)\1/, 2)
 		.map(x => createTree('italic', P.alt(
 			r.bold,
 			r.strike,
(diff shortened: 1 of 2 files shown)


acid-chicken (changes requested):
The tests were failed on my computer.

```diff
      + expected - actual

       [
         {
      -    "children": []
      +    "children": [
      +      {
      +        "children": []
      +        "node": {
      +          "props": {
      +            "text": "foo"
      +          }
      +          "type": "text"
      +        }
      +      }
      +    ]
           "node": {
      -      "props": {
      -        "text": "<i>foo</i>"
      -      }
      -      "type": "text"
      +      "props": {}
      +      "type": "italic"
           }
         }
       ]

      at Context.it (test/mfm.ts:864:12)
```

AyaMorisawa:
https://github.com/syuilo/misskey/pull/3732#issuecomment-450160403

I request a review from @syuilo

syuilo:
Thank you!
I will continue to make HTML syntax available. I will also modify the test.

PR #3934, opened 2019-01-19
Repack note for each follower it is published to

### LINKED ISSUE #3928: Reply targets not fully shown when streamed
# Summary
Reply targets are sometimes incorrectly shown as "this post is private" on followers-only notes, but only when streamed to the browser in realtime.

# Expected Behavior
Reply targets should be displayed if the user is allowed to see the note.

# Actual Behavior
When streamed in realtime, the reply target is shown as "this post is private". It is shown correctly if the page is reloaded.

# Steps to Reproduce
1. Create a followers-only note on a remote instance, followed by the misskey user.
2. Reply to that note from the remote instance.
3. The reply target will be shown as "this post is private"
4. Reload the page and see the reply is shown normally.

# Environment
This appears to occur in the webui only. I didn't try any other apps but I don't have any issue like this in Subway Tooter so I think the API is okay.

# Fix?
I believe I fixed this by generating `detailPackedNote` for each follower the note is published to in `src/services/note/create.ts`, `publishToFollowers`, and specifying `follower.followerId` instead of `null` for the user. I can create a PR for this if desired, but I don't know if it causes any other problems.

# Screenshots
Current behavior:
screenshot_2019-01-18_07-21-58
After the fix mentioned above:
screenshot_2019-01-18_07-21-41

natarii:
# Summary
This is an attempt to fix #3928. It re-generates `detailPackedNote` for each follower the note is published to, specifying the follower's user so the reply target visibility is correct.

### DIFF
--- a/src/services/note/create.ts
+++ b/src/services/note/create.ts
@@ -525,11 +525,6 @@ async function publishToUserLists(note: INote, noteObj: any) {
 }
 
 async function publishToFollowers(note: INote, user: IUser, noteActivity: any) {
-	const detailPackedNote = await pack(note, null, {
-		detail: true,
-		skipHide: true
-	});
-
 	const followers = await Following.find({
 		followeeId: note.userId
 	});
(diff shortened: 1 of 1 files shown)


syuilo:
Thank you!
But I think that this approach has performance problems. If there are 1000 followers, it will be packed 1000 times. pack is a relatively expensive process because it has multiple database accesses. I will try to solve this problem with another approach.
Thanks anyway.

natarii:
Thanks, I was also concerned about performance issues.

PR #4088, opened 2019-02-02
Fix "About" url error

fz0000:
Currently there are only English and Japanese About pages. In other languages user will see 404 error.
So I change the url "/docs/${lang}/about" into 2 urls: "/docs/en-US/about" and "/docs/ja-JP/about"


fz0000:
A better solution is adding the About pages on crowdin to localize them.
However, I'm not sure if it's necessary, since English should be enough for most developers...

syuilo:
I fixed that issue by [this commit](https://github.com/syuilo/misskey/commit/2a50997a7586976888826e2f8760010dad68eccb).
Thanks anyway!

Issue #1821, opened 2018-06-28
Increase font size a little in the Deck view


mareklach:
I feel like the font size could be made slightly bigger in the Deck view? It's just a little too small right now I think?
<img width="1279" alt="screen shot 2018-06-28 at 23 03 31" src="https://user-images.githubusercontent.com/10089099/42063037-c7938fba-7b27-11e8-9148-bebd8b0a5aa8.png">

marek-lach:
While Misskey in its primary view is already pretty nice, I feel that the Deck view, which is cool, could be much improved if there was a user setting to increase font-size in the Deck view, or zoom in a little.
Especially for Roman alphabet used in the West I think the ability to enlarge the Deck's font would improve readability of the timeline a great deal.

@syuilo Perhaps implementing a font-size, or a zoom level for the Deck could be considered as important for wider adoption. I would use Misskey significantly more often if I could set a bigger fontsize for the Deck.
At least for me, a slightly bigger font, like the one Mastodon uses, significantly contributes to preventing eye strain:

![bigger_font](https://user-images.githubusercontent.com/45913977/50525515-c0c4a200-0ad3-11e9-8146-ebc662068e8a.PNG)

For people with poorer eyesight, this could be a great usability improvement overall, I feel.
Thank you for your efforts,

PR #4136, opened 2019-02-04
Attempt to increase font size of notes in Deck for latin script

marek-lach:
It was too small to see well, especially for latin script languages, like English...

# Summary
This increases the font-size of notes in Deck view, which was too small for latin script languages.


acid-chicken (approved):
↓ Before (12px)



<p align="right">After (15px) ↑</p>

syuilo:
Related to #1821

syuilo:
I just fixed this issue by [this commit](https://github.com/syuilo/misskey/commit/ef618b2431a70df25c892f05795fd23bc3191c4e). so I close this PR. Thanks anyway!
Ultimately, I would like to allow users to set the font size.

PR #4386, opened 2019-02-28
Increase font size of some components for better readability

marek-lach:
# Summary


syuilo:
I made it possible for users to set the font size with [this commit](https://github.com/syuilo/misskey/commit/d83efecc94871633ccac804b4435327a0be72432).
Thanks anyway 👍

PR #4932, opened 2019-05-16
Change primary list to be selected by default

ghost:
## Summary

Currently, when adding a user to a list, it's always necessary to select a list.  
But to simplify this, _the primary list_ will be the **default**.

This also avoids the problem of not being able to successfully select the list **twice** if you do the same with pwa.

I'm waiting for your review.  
Thanks


ghost:
> the problem of not being able to successfully select the list twice if you do the same with pwa.

src/client/app/common/views/components/user-menu.vue
```diff
@@ -77,7 +77,8 @@ export default Vue.extend({
 				select: {
 					items: lists.map(list => ({
 						value: list.id, text: list.name
-					}))
+					})),
+					selected: lists[0].id
```
acid-chicken: I think that the TypeError is thrown when list is empty.

src/client/app/common/views/deck/deck.vue
```diff
@@ -190,7 +190,8 @@ export default Vue.extend({
 							select: {
 								items: lists.map(list => ({
 									value: list.id, text: list.name
-								}))
+								})),
+								selected: lists[0].id
```
acid-chicken: Same as above.

src/client/app/common/views/components/dialog.vue
```diff
@@ -67,7 +67,8 @@ export default Vue.extend({
 			required: false
 		},
 		select: {
-			required: false
+			required: false,
+			selected: null
```
syuilo: ここには required とか type とかのプロパティしか置けないですね
https://vuejs.org/v2/api/index.html#props

ghost: 修正しました

ghost:
Should I choose a primary list or make the list order changeable?

syuilo:
I have resolved this issue by https://github.com/syuilo/misskey/commit/50251955cdeb5fbb8ff279bbcbea68d2bf606c7f.
Thanks anyway!

PR #5054, opened 2019-06-14
Use halfwidth space instead of fullwidth space in code

AyaMorisawa:
## Summary
We don't need fullwidth spaces! :fist_oncoming:


syuilo:
👍

PR #5088, opened 2019-06-27
Implement Webauthn 🎉

### LINKED ISSUE #3660: Support Web Authentication API
https://www.w3.org/TR/webauthn/

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67687242-support-web-authentication-api?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Mstrodl:
## Summary
Implements [Webauthn](https://webauthn.io) for most security keys.
Somewhat WIP, but everything _should_ be working!

Resolves #3660

Settings page
Registration
Registration Save
Logging in with key
Timed out login


Mstrodl:
Only issue is the need for Japanese translations and probably better typings (I don't use Typescript so I did my best)

src/client/app/common/views/components/settings/2fa.vue
```diff
@@ -1,11 +1,55 @@
...
 	<p v-if="!data && !$store.state.i.twoFactorEnabled"><ui-button @click="register">{{ $t('register') }}</ui-button></p>
 	<template v-if="$store.state.i.twoFactorEnabled">
+		<h2 class="heading">{{ $t('totp-header') }}</h2>
 		<p>{{ $t('already-registered') }}</p>
 		<ui-button @click="unregister">{{ $t('unregister') }}</ui-button>
+
+
+		<hr v-if="supportsCredentials" class="totp-method-sep">
+
+		<template v-if="supportsCredentials">
```
rinsuki, line 11: Is it possible to move the above `<hr>` into `<template>`?

src/client/app/common/views/components/signin.vue
```diff
@@ -27,6 +45,12 @@ import i18n from '../../../i18n';
 import { apiUrl, host } from '../../../config';
 import { toUnicode } from 'punycode';
 
+function hexifyAB(buffer) {
+	return Array.from(new Uint8Array(buffer))
+		.map(item => item.toString(16).padStart(2, 0))
+		.join('');
+}
```
rinsuki: If you use this function from multiple files, consider moving the code to a dedicated file rather than copying and pasting.

rinsuki (commented):
thanks for implement awesome feature!

src/client/app/common/views/components/settings/2fa.vue
Mstrodl: @rinsuki Yes, but I moved it outside because it separates the two different types of credentials which was why I thought it made more sense outside

src/client/app/common/views/components/signin.vue
Mstrodl: @rinsuki Where should I put the file? I'm not really sure where that sort of thing fits within the structure of the app :/

Mstrodl:
I've also created a small daemon to clean up `AttestationChallenges` that were never used to make sure the DB doesn't end up getting filled with them

acid-chicken (dismissed):
Thank you for your great contribution!
Please fix these codes.

mei23:
lintはこちらでなおしてあげてもいいかと思います。
contributorにいきなりlint100箇所以上直せってというのはどうかと思います。

acid-chicken:
batch でワンポチじゃダメですか
できる限り負担を軽減した方が良いと考え `suggestion` 形式にしておりますが

mei23:
> batch でワンポチじゃダメですか

私はその方法を知らないし、100箇所以上Resolveを確認する必要はあるのではと

Mstrodl:
@acid-chicken Sorry, I didn't know there was a proper linter until just now and my own linter (prettier) ended up breaking a lot of your guidelines there. I've applied your suggestions before seeing that there was a linter..
The one thing I'm confused about is how best to deal with the `try {} catch() {}` with a `var` inside it. I think in this case, it makes sense to use `var` because it will expose it to the context of the function itself and there cannot be any conditions where `data` would be undefined

I would also really appreciate if someone with a `fido-u2f` key (Yubikeys and other older keys do this) could make sure this works for them because my keys only do `packed` attestation.

src/server/api/endpoints/i/2fa/key-done.ts
```diff
@@ -0,0 +1,139 @@
...
+	if (publicKey.get(3) != -7) {
+		throw new Error('alg mismatch');
+	}
+
+	if (!procedures[attestation.fmt]) {
+		throw new Error('unsupported fmt');
+	}
+
+        try {
+            var verificationData = procedures[attestation.fmt].verify({
```
acid-chicken:
```suggestion
		// tslint:disable-next-line:no-var-keyword
		var verificationData = procedures[attestation.fmt].verify({
```

acid-chicken (commented):
@Mstrodl Thank you for applying those suggestions.

> I think in this case, it makes sense to use var because it will expose it to the context of the function itself and there cannot be any conditions where data would be undefined

Yes, you are right. Linting is just a guide for shaking out the bugs, so we can ignore them when we can guarantee the code is safe.
Therefore, I drop the change request and send a new suggest for ignoring lint. :)

syuilo:
ありがとうございます。
AttestationChallenge テーブルが UserProfile テーブルと一対一なら、テーブルを分けずに UserProfile テーブルに新しくカラムを追加した方が良いと思うのですが、新しく AttestationChallenge テーブルを追加した理由は何でしょうか？

Thank you very much.
If the AttestationChallenge table is one-to-one with the UserProfile table, I think it would be better to add a new column to the UserProfile table without splitting the table. What is the reason for adding the new AttestationChallenge table?

Mstrodl:
@syuilo It is possible that a user may have more than one challenge available to them. For example, if the user starts registration of a key and also logs in with a different key at the same time or similar. Perhaps an attacker could still lock someone out of their account with just their username and password by generating new challenges every 10 seconds so the real user's challenge was always invalid by the time they went to use their security key which would largely defeat the purpose of 2FA.

Mstrodl:
I'm also currently working on supporting some more attestations on my local copy (android-safetynet, android-key, none) as well as some yubikey u2f fixes (I found a friend with a yubikey to test with)

syuilo:
> It is possible that a user may have more than one challenge available to them.

understood. Thank you!

@acid-chicken @mei23 @rinsuki OneToOneをManyToOneに変更した方が良さそう？
あと UserProfile と紐づけるより User と直接紐づけた方が自然そう

Mstrodl:
The above commits should add support for android N and later's built-in security keys and 'anonymous mode' on Yubikeys (which appears to actually be outside of spec, but they do it anyways)

syuilo:
~~Is it possible for one user to have multiple UserSecurityKeys?~~
Resolved

Mstrodl:
@syuilo Yes, this is useful if you have a backup key, a wireless key for a phone, or just an extra key. (People got mad at Twitter for not allowing more than one key to be registered, so I figured this was the best option)

syuilo:
I simplified some ID column names. Seems good?

Mstrodl:
@syuilo your commit above will break the frontend, (also: `credentialId` in this case is generated by the security key, so I'm not sure moving it to `id` is a good plan anyways)

Mstrodl:
@syuilo Login uses `credentialId` also to tell the browser which security key(s) to give the challenge to

Mstrodl:
@syuilo Any idea why has CircleCI has been stuck pending since I opened the PR? Is it just not set up properly? or did I do something wrong?

acid-chicken:
@Mstrodl CI was broken in the past and the patch #4909 is waiting for merge. So you can ignore these statuses for now.

syuilo:
I added a migration.

syuilo:
I added a japanese translations.
間違ってたら教えて

Mstrodl:
Currently, the only unsupported attestations are Windows Hello / TouchID on macOS (`tpm` attestation is super complicated) and `android-key`, both of which are just due to lack of testing resources. In theory `android-key` should work, but I haven't been able to test (my device uses `android-safetynet`, which seems to work well!)

aveao:
I can try and help with testing android-key and Windows Hello (also possibly touch id).

src/server/api/2fa.ts
```diff
@@ -1,6 +1,6 @@
 import * as crypto from 'crypto';
 import config from '../../config';
-import * as jsrsasign from 'jsrsasign';
+const jsrsasign = require('jsrsasign');
```
acid-chicken: #3713

acid-chicken: @syuilo よければやりましょうか？

syuilo: お願いします 🙏

acid-chicken: はい

syuilo:
二段階認証の設定が完了してない状態でログインしようとすると二段階認証を求められるバグがありますね

There is a bug that requires two-step authentication if you try to login without completing the setting of two-step authentication.

src/server/api/2fa.ts
Mstrodl: Uh this should really not be done like this, it breaks the whole mantra of ESM modules :/

Mstrodl:
@syuilo Apologies! This is due to a missing check for the status of 2FA on the frontend. This should hopefully be fixed now! (Along with the error on the janitor daemon caused by a missing cast to Date)

This was, in fact, reproducible even if the account had never touched 2FA settings. Any time a user without 2FA tried to login, they would be prompted for a 2FA token

package.json
```diff
@@ -105,7 +105,7 @@
 		"autosize": "4.0.2",
 		"autwh": "0.1.0",
 		"bcryptjs": "2.4.3",
-		"bootstrap": "^4.3.1",
+		"bootstrap": "4.3.1",
```
Mstrodl, line 108: This was here because your dependency on bootstrap pulls in `^4.3.1`, so I assumed it would make sense to have matching versions

syuilo:
Is it okay to merge?

acid-chicken:
jsrsasign.d.ts を待ってほしい

Mstrodl:
@acid-chicken [jsrsasign.d.ts](https://gist.github.com/aneilbaboo/2e0e90cdbfb44505405fd07c29382225)?

acid-chicken:
@Mstrodl No, that is too small. I've written almost 500 lines of code, but it's still a work in progress.

Mstrodl:
@acid-chicken oh, I assumed you'd just implement the things our code used at first

syuilo:
@acid-chicken any progress?

acid-chicken:
@syuilo almost done

acid-chicken:
@syuilo done so please review

syuilo:
とりあえずマージします
皆様ありがとうございました 🙏 

Thank you guys 🙏

PR #5532, opened 2019-10-20
boot: remove setAttribute() calls and translate reload msg

Mstrodl:
## Summary
Cleans up some of the code in boot. Uses actual properties instead of `setAttribute()` because it's faster and translates the `alert()` popup.

### DIFF
--- a/src/client/app/boot.js
+++ b/src/client/app/boot.js
@@ -103,7 +103,7 @@
 	// If mobile, insert the viewport meta tag
 	if (isMobile) {
 		const viewport = document.getElementsByName("viewport").item(0);
-		viewport.content = `${viewport.getAttribute('content')},minimum-scale=1,maximum-scale=1,user-scalable=no`;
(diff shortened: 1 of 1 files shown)


src/client/app/boot.js
```diff
@@ -99,8 +103,7 @@
 	// If mobile, insert the viewport meta tag
 	if (isMobile) {
 		const viewport = document.getElementsByName("viewport").item(0);
-		viewport.setAttribute('content',
-			`${viewport.getAttribute('content')},minimum-scale=1,maximum-scale=1,user-scalable=no`);
+		viewport.content = `${viewport.getAttribute('content')},minimum-scale=1,maximum-scale=1,user-scalable=no`;
```
acid-chicken: I think you can also use getter of content propery.

```suggestion
		viewport.content = `${viewport.content},minimum-scale=1,maximum-scale=1,user-scalable=no`;
```

syuilo:
Thank you!

PR #5810, opened 2020-02-02
Add ZWSP when federating emojis for Mastodon compatibility

chocological00:
## Summary
When inserting two custom emojis simultaneously, Misskey specification does :custom_1::custom_2:. This causes a problem on some other ActivityPub implements (especially Mastodon), due to some weird RegEx issues. (They expect a space to be in between them, like :custom_1: :custom_2:)

This patch resolves that issue by adding a zero-width space before and end of the colon (:) when federating. This does not change how emojis in a note is stored in Misskey's DB.

### DIFF
--- a/src/mfm/toHtml.ts
+++ b/src/mfm/toHtml.ts
@@ -87,7 +87,7 @@ export function toHtml(tokens: MfmForest | null, mentionedRemoteUsers: IMentione
 		},
 
 		emoji(token) {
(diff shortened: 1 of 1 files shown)


rinsuki (commented):
can you use unicode escape sequence (because we can't see ZWSP)?

syuilo:
Thank you guys 🥰🥰🥰

PR #5811, opened 2020-02-02
Allow wider window for totp verification

chocological00:
When using two-factor authentication, the token sometimes fail due to time synchronization issues. This patch makes the allowed window wider by 1 minute (30 seconds before and after the server time).


rinsuki (approved):

syuilo:
Thank you 👍

PR #5812, opened 2020-02-02
Fix for CASCADE DELETE not being federated

chocological00:
## Summary
Misskey's DB deletes notes using CASCADE DELETE. While this itself could be a specification, notes deleted by CASCADE not being federated to other instances is a huge problem - it leaves a copy of notes in other instances that is already deleted on the original instance, leaving them undeletable. 

This patch fetches all candidate notes for deletion whenever CASCADE is invoked, and send tombstones to other instances before the actual deletion is occured. Only notes composed by local users are affected.

This issue has been brought up several times. I will add the related issues in comment later.

### DIFF
--- a/src/services/drive/delete-file.ts
+++ b/src/services/drive/delete-file.ts
@@ -1,11 +1,17 @@
 import { DriveFile } from '../../models/entities/drive-file';
 import { InternalStorage } from './internal-storage';
-import { DriveFiles, Instances, Notes } from '../../models';
+import { DriveFiles, Instances, Notes, Users } from '../../models';
 import { driveChart, perUserDriveChart, instanceChart } from '../chart';
 import { createDeleteObjectStorageFileJob } from '../../queue';
 import { fetchMeta } from '../../misc/fetch-meta';
 import { getS3 } from './s3';
 import { v4 as uuid } from 'uuid';
+import { Note } from '../../models/entities/note';
+import { renderActivity } from '../../remote/activitypub/renderer';
+import renderDelete from '../../remote/activitypub/renderer/delete';
+import renderTombstone from '../../remote/activitypub/renderer/tombstone';
+import config from '../../config';
+import { deliverToFollowers } from '../../remote/activitypub/deliver-manager';
 
 export async function deleteFile(file: DriveFile, isExpired = false) {
 	if (file.storedInternal) {
@@ -63,7 +69,7 @@ export async function deleteFileSync(file: DriveFile, isExpired = false) {
 	postProcess(file, isExpired);
 }
 
-function postProcess(file: DriveFile, isExpired = false) {
+async function postProcess(file: DriveFile, isExpired = false) {
 	// リモートファイル期限切れ削除後は直リンクにする
 	if (isExpired && file.userHost !== null && file.uri != null) {
 		DriveFiles.update(file.id, {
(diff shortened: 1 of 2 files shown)


chocological00:
Related: #4863
FYI: This patch has been running live in https://madost.one/ for last 1 month. No special problems observed.

src/services/note/delete.ts
```diff
@@ -64,4 +69,30 @@ export default async function(user: User, note: Note, quiet = false) {
...
+		id: note.id,
+		userId: user.id
+	});
+}
+
+async function findCascadingNotes(note: Note) {
+	const cascadingNotes: Note[] = [];
+
+	const recursive = async (noteId: string) => {
+		const replies = await Notes.find({ replyId: noteId });
```
mei23: It would be better to get `user` with JOIN.
```suggestion
		const query = Notes.createQueryBuilder('note')
			.where('note.replyId = :noteId', { noteId })
			.leftJoinAndSelect('note.user', 'user');

		const replies = await query.getMany();
```

src/services/drive/delete-file.ts
```diff
@@ -106,3 +126,35 @@ export async function deleteObjectStorageFile(key: string) {
...
+
+async function findRelatedNotes(fileId: string) {
+	// NOTE: When running raw query, TypeORM converts field name to lowercase. Wrap in quotes to prevent conversion.
+	const relatedNotes = await Notes.createQueryBuilder('note').where(':id = ANY("fileIds")', { id: fileId }).getMany();
+	for (const relatedNote of relatedNotes) {
+		const user = await Users.findOne({ id: relatedNote.userId });
+		if (user)
+			relatedNote.user = user;
+	}
+	return relatedNotes;
```
mei23, line 139: Seems replaceable using JOIN
```ts
async function findRelatedNotes(fileId: string) {
	// NOTE: When running raw query, TypeORM converts field name to lowercase. Wrap in quotes to prevent conversion.
	const query = Notes.createQueryBuilder('note')
		.where(':id = ANY("fileIds")', { id: fileId })
		.leftJoinAndSelect('note.user', 'user');

	return await query.getMany();
}
```

src/services/drive/delete-file.ts
```diff
@@ -106,3 +126,35 @@ export async function deleteObjectStorageFile(key: string) {
...
+}
+
+async function findCascadingNotes(note: Note) {
+	const cascadingNotes: Note[] = [];
+
+	const recursive = async (noteId: string) => {
+		const replies = await Notes.find({ replyId: noteId });
```
mei23: It would be better to get `user` with JOIN.
```suggestion
		const query = Notes.createQueryBuilder('note')
			.where('note.replyId = :noteId', { noteId })
			.leftJoinAndSelect('note.user', 'user');

		const replies = await query.getMany();
```

src/services/note/delete.ts
chocological00: Fixed :>

src/services/drive/delete-file.ts
mei23: Probably looks unfixed.
I think the whole `findRelatedNotes` function will be replaced by the above.

mei23 (approved):
findRelatedNotes JOIN にしたいのは残ってるけど
挙動は大丈夫そう

syuilo:
ありがとうございます 🙏

PR #5909, opened 2020-02-10
Update en-US.yml

okpierre:
## Summary

@syuilo fix typo in en-US.yml


tamaina:
Thank you for fixing but you must not edit that file.

See: https://github.com/syuilo/misskey/blob/develop/locales/README.md

PR #6116, opened 2020-02-29
Fix 404 unread svg

okpierre:
## Summary

I commented on the issue https://github.com/syuilo/misskey/issues/5871 but forgot to create a pull request so here it is @syuilo


syuilo:
Thank you!

PR #6188, opened 2020-03-27
Update fr-FR.yml

kazukyakayashi:
## Summary

Sentence correction for greater readability


tamaina:
Thank you for fixing but you must not edit that file.

See: https://github.com/syuilo/misskey/blob/develop/locales/README.md

PR #6209, opened 2020-03-31
Option to hide revealed sensitive media

yitsushi:
## Summary

This PR commit adds a button on sensitive images and videos
to to hide them without refreshing the page.

Resolve #3994

Note: This is the first time I touched any kind of TypeScript or Vue code, please give me advice if something looks bad or anything, thanks.

PR #6210, opened 2020-04-01
Ability to set header image for a Page

### LINKED ISSUE #6185: Adding an image to the page header does not work.
## 💡 Summary
When I create a page and I want to add a header image to it, nothing happens.

## 🙂 Expected Behavior

Opens the Drive to be able to add or select an image.

## ☹️ Actual Behavior

Adding an image to the page header doesn't work, nothing happens when I click on the button to add the image.

image-header

## 📌 Environment

Firefox 74.0 Windows and GNU/Linux (Manjaro)

yitsushi:
## Summary

 - Add header image to Page
 - Show it on Page view
 - Show correctly it on Page list view
 - On the Page list view, pages have a light border
   to make it easier to see an image belongs to a page

Fix #6185

### DIFF
--- a/src/client/components/page/page.image.vue
+++ b/src/client/components/page/page.image.vue
@@ -22,12 +22,10 @@ export default Vue.extend({
 		};
 	},
 	created() {
-		if (this.value.hasOwnProperty('fieldId')) {
-			// media
-			this.image = this.page.attachedFiles.find(x => x.id === this.value.fileId);
-		} else {
-			// plain image
-			this.image = this.value
+		this.image = this.page.attachedFiles.find(x => x.id === this.value.fileId);
+
+		if (typeof this.image === "undefined") {
+			this.image = this.value;
 		}
 	}
 });
--- a/src/client/components/page/page.image.vue
+++ b/src/client/components/page/page.image.vue
@@ -22,7 +22,13 @@ export default Vue.extend({
 		};
 	},
 	created() {
-		this.image = this.page.attachedFiles.find(x => x.id === this.value.fileId);
+		if (this.value.hasOwnProperty('fieldId')) {
+			// media
+			this.image = this.page.attachedFiles.find(x => x.id === this.value.fileId);
+		} else {
+			// plain image
+			this.image = this.value
+		}
 	}
 });
 </script>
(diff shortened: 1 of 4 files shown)


src/client/components/page/page.image.vue
```diff
@@ -22,7 +22,13 @@ export default Vue.extend({
 		};
 	},
 	created() {
-		this.image = this.page.attachedFiles.find(x => x.id === this.value.fileId);
+		if (this.value.hasOwnProperty('fieldId')) {
```
yitsushi, line 25: I couldn't find an example how to convert an Image to a Media when I call `<x-image>`

yitsushi: And actually it's not good. :(

yitsushi:
### How it looks like

Page List:
image

Page List :hover
image

Page view with header image:
image

yitsushi:
`yarn lint` error
```
src/client/store.ts:133:1
ERROR: 133:1  no-trailing-whitespace  trailing whitespace
ERROR: 137:1  no-trailing-whitespace  trailing whitespace
ERROR: 142:1  no-trailing-whitespace  trailing whitespace
ERROR: 151:1  no-trailing-whitespace  trailing whitespace
ERROR: 161:1  no-trailing-whitespace  trailing whitespace
ERROR: 163:1  no-trailing-whitespace  trailing whitespace
```
is fixed in #6209

src/client/components/page/page.image.vue
```diff
@@ -23,6 +23,10 @@ export default Vue.extend({
 	},
 	created() {
 		this.image = this.page.attachedFiles.find(x => x.id === this.value.fileId);
+
+		if (typeof this.image === "undefined") {
+			this.image = this.value;
+		}
```
syuilo: Is this required?

yitsushi: I did it because I did not find a way to convert an Image to a Media or something. When I pass an Image from `page.eyeCatchingImage` it does not have `fileId`, but it's a file and has an `id`, but not `fileId`. If it has a `fileId` does not matter because in that case it's not even a page attachment just a property of a page, so it's not in the attachment array.

Any idea how can I resolve this problem without duplicating `x-image`? I would like to fix it better and learn from it as possible.

syuilo: I see.
You can simply use `` tag instead of `<x-image>` in https://github.com/syuilo/misskey/pull/6210/files#diff-b81443d697613ee9cb5072d734e28f73R8

yitsushi: Addressed and I don't know why I did not even consider to use raw `img` tag.

syuilo:
Thank you🥰

PR #6290, opened 2020-04-22
Implement Unrenote Functionality

chocological00:
## Summary
Re-add the missing "undo renote" function.

I added the button on the top bar like this as the previous (v11) implementation gave users a false impression of interacting with the original note:
image

NOTE: Unfortunately I don't speak or write Japanese, so someone would have to add the corresponding locale to `ja-JP.yml` :<


src/client/components/note.vue
```diff
@@ -21,6 +21,12 @@
 			</router-link>
 		</i18n>
 		<div class="info">
+			<span v-if="isMyRenote">
+				<fa :icon="faTimes"/>
+				<span>&nbsp;</span>
+				<a href="#" @click="(e) => { e.preventDefault(); delRenote() }">{{ $t('unrenote') }}</a>
```
acid-chicken: You can also write here like this, maybe.

```suggestion
				<a href="#" @click.prevent="delRenote()">{{ $t('unrenote') }}</a>
```

u1-liquid:
Actually the 'undo renote' function is already implemented as same way with v11 😳
image

I know this is not so friendly UI / UX.
How about add 'undo renote' menu on renote button? I think this is more reasonable.
image

And also, I think 'Renoteを解除しますか？' will be enough localization string for `ja-JP.yml`

chocological00:
Oops, I did mot know that :P 
I personally dont think adding it to the renote menu is a good UX as it might give users a false impression that s/he is interacting with the original note, not the renoted one. (Which is why I wanted to keep the button at the top bar in this PR)
However, current timestamp button is indeed pretty hard to find. It also doesnt look very interact-able, as it has no special color or any button surrounding it. How about adding a little trash can icon next to the timestamp that opens a delete confirmation modal? (and removing the functionality from timestamp link)
If any maintainer thinks this is a good idea, I will add more commits accordingly. :>

locales/ko-KR.yml
```diff
@@ -79,6 +79,7 @@ followRequestPending: "팔로우 허가 대기중"
 enterEmoji: "이모지 입력"
 renote: "Renote"
 unrenote: "Renote 취소"
+unrenoteConfirm: "정말로 Renote를 취소하시겠습니까?"
```
tamaina, line 82:
```suggestion
```

[DO NOT edit locale files except ja-JP.yml.](https://github.com/syuilo/misskey/blob/develop/locales/README.md)

It would be good to add the following to ja-JP.yml

```yaml
unrenoteConfirm: "このノートを削除しますか？"
````

File: src/client/components/note.vue [hunk 3]
```diff
@@ -21,6 +21,12 @@
 			</router-link>
 		</i18n>
 		<div class="info">
+			<span v-if="isMyRenote">
+				<fa :icon="faTimes"/>
+				<span>&nbsp;</span>
+				<a href="#" @click.prevent="delRenote()">{{ $t('unrenote') }}</a>
+				<span>&nbsp;</span>
+			</span>
 			<button class="_button time" @click="showRenoteMenu()" ref="renoteTime"><mk-time :time="note.createdAt"/></button>
```
tamaina, line 30: It would be better if the change is integrated with the existing `showRenoteMenu()` in order not to make the UI patchy.

syuilo:
時間の横に「...」アイコンを付けるだけで良いと思います
それなら時間の部分をクリックすると何らかのメニューが表示されることが分かるようになる

chocological00:
PR no longer relevant.

PR #6366, opened 2020-05-14
Add an indicator to make the undo renote button look more interactable

chocological00:
## Summary
image

Related discussion on #6290 

Indicator is on the left to keep the timestamp text aligned.


syuilo:
Thank you🥰

PR #6373, opened 2020-05-16
Only show ellipsis if the button's interactable

chocological00:
## Summary
Sorry for duplicate entry. I have noticed that I forgot to check if it's the user's renote. :<
This PR fixes that issue and show the ellipsis icon only if it is the user's renote.


syuilo:
👍👍👍

PR #6374, opened 2020-05-16
Fix CASCADE-related problems

chocological00:
## Summary
This PR fixes two problems:
1. Quoted renote deletions not being federated to remote servers when CASCADE is invoked. 
2. CASCADE-invoked deletions not being federated to relays.


mei23:
It seems necessary to branch Activity in Renote and Quote, so I will test it later.

mei23 (approved):

syuilo:
Thank you guys🥰

PR #6456, opened 2020-06-06
Update english nyaize

robflop:
## Summary

日本語の語句がしりません、ですから英語でかいます、すみません。

I've updated the english nyaize functionality. I had some other ideas, but decided to only implement the "na" to "nya" parts. The RegEx case insensitive flag ``i`` does not work here because it would replace e.g. "Naturally" with "nyaturally", which would be incorrect. Therefore two separate replacements have been implemented.

In addition, I fixed the existing replacements. For example, "Morning everyone!" was replaced with "mornyan everynyan!" when it should have been "Mornyan everynyan!" (see above problem).

### DIFF
--- a/src/misc/nyaize.ts
+++ b/src/misc/nyaize.ts
@@ -3,9 +3,9 @@ export function nyaize(text: string): string {
 		// ja-JP
 		.replace(/な/g, 'にゃ').replace(/ナ/g, 'ニャ').replace(/ﾅ/g, 'ﾆｬ')
 		// en-US
-		.replace(/Na/g, 'Nya').replace(/na/g, 'nya')
-		.replace(/Morning/g, 'Mornyan').replace(/morning/g, 'mornyan')
-		.replace(/Everyone/g, 'Everynyan').replace(/everyone/g, 'everynyan')
+		.replace(/(?<=n)a/gi, x => x === 'A' ? 'YA' : 'ya')
+		.replace(/(?<=morn)ing/gi, x => x === 'ING' ? 'YAN' : 'yan')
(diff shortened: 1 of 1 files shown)


src/misc/nyaize.ts
```diff
@@ -3,7 +3,9 @@ export function nyaize(text: string): string {
 		// ja-JP
 		.replace(/な/g, 'にゃ').replace(/ナ/g, 'ニャ').replace(/ﾅ/g, 'ﾆｬ')
 		// en-US
-		.replace(/morning/gi, 'mornyan').replace(/everyone/gi, 'everynyan')
+		.replace(/Na/g, 'Nya').replace(/na/g, 'nya')
+		.replace(/Morning/g, 'Mornyan').replace(/morning/g, 'mornyan')
+		.replace(/Everyone/g, 'Everynyan').replace(/everyone/g, 'everynyan')
```
acid-chicken: Thank you for the contribution! How about look like this?

```suggestion
		.replace(/(?<=n)a/gi, x => x === 'A' ? 'YA' : 'ya')
		.replace(/(?<=morn)ing/gi, x => x === 'ING' ? 'YAN' : 'yan')
		.replace(/(?<=every)one/gi, x => x === 'ONE' ? 'NYAN' : 'nyan')
```

It can replace all of an uppercase word, a titlecase word, and a lowercase word per one call.

robflop: That's great, though if you have "nAme" (or similar cases with the other replacements) you will get a result of "nYAme". If that's okay then I'd commit your changes.

acid-chicken (approved):
:shipit:

syuilo:
Thank you guys!

PR #6598, opened 2020-07-28
Mapping files folder outside the container

Candinya:
# Summary

In order to prevent the loss of files uploaded by users when upgrading Misskey deployed with Docker.
**But** it might be necessary to create an accessible folder before `docker-compose up -d` (Not fully tested)

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -13,6 +13,8 @@ services:
     networks:
       - internal_network
       - external_network
+    volumes:
+      - ./files:/misskey/files
 
   redis:
     restart: always


syuilo:
cc: Dockerに詳しい人
(@acid-chicken @mei23 @rinsuki ？)

rinsuki (approved):

mei23 (approved):
うちも同じ変更してたので大丈夫のはず

mei23:
Resolve #6455

syuilo:
あざます！

PR #6601, opened 2020-07-29
added aria label and alt tag

okpierre:
## Summary

added aria label and alt tag


syuilo:
Thank you.

ありがとうございます👀
私の考えでは、アイコン画像にalt付ける必要性は無さそう。
aria-labelについても必要性がよくわかっていない

okpierre:
> Thank you.
> 
> ありがとうございますeyes
> 私の考えでは、アイコン画像にalt付ける必要性は無さそう。
> aria-labelについても必要性がよくわかっていない

@syuilo 

alt attribute improves SEO and accessibility (example: users with disabilities ... also see aria-label)
aria-label describe the action to anyone using a screen reader (example: Android phones talkback for blind or visually impaired users or live transcribe for deaf users https://www.youtube.com/watch?v=jLCwjIaPXwA)

src/client/components/autocomplete.vue
```diff
@@ -2,7 +2,7 @@
 <div class="swhvrteh" @contextmenu.prevent="() => {}">
 	<ol class="users" ref="suggests" v-if="type === 'user'">
 		<li v-for="user in users" @click="complete(type, user)" @keydown="onKeydown" tabindex="-1" class="user">
-			<img class="avatar" :src="user.avatarUrl"/>
+			<img class="avatar" :src="user.avatarUrl" :alt="image.name"/>
```
rinsuki, line 5: I think `image.name` is file name of image (it is maybe "Screen Shot - 20200730012345.png", "IMG_1234.jpg" or other), it is not alt text.
in this context, I think we must use "title" attribute instead of "alt" attribute.

do you think about this?

src/client/components/page/page.image.vue
```diff
@@ -1,6 +1,6 @@
 <template>
 <div class="lzyxtsnt">
-	<img v-if="image" :src="image.url"/>
+	<img v-if="image" :src="image.url" :alt="image.name"/>
```
rinsuki, line 3: same reason

okpierre: @rinsuki 

Yes you are correct. Title or description is much better for alt. 

I used `image.name` because it was previously used in Misskey code
https://github.com/syuilo/misskey/blob/c6837b9fdfbc30d8e3ede381293281db2d6364f0/src/client/components/image-viewer.vue#L3

What is the correct way to display title or description for Misskey? Is it `image.title`?

syuilo:
ふーむ
Misskeyをスクリーンリーダーで使うユースケースはあるのかしら
SEOについても、Misskeyはどうでもいいし

src/client/components/page/page.image.vue
rinsuki: actually correct data of title (or description) is not available.
in database schema, we have `image.comment`, but we can't set any value from API, so this is dead property.

(but anyways we must provide alt-text from mastodon or other servers, so i will create issue about this)

okpierre:
> ふーむ
> Misskeyをスクリーンリーダーで使うユースケースはあるのかしら
> SEOについても、Misskeyはどうでもいいし

@syuilo screen reader is for users with disabilities

src/client/components/page/page.image.vue
rinsuki: created: #6606

src/client/components/autocomplete.vue
Johann150: I'm not sure if user avatars have alt text in the first place. I would assume most do not.

src/client/pages/user/index.timeline.vue
```diff
@@ -1,9 +1,9 @@
 <template>
 <div class="kjeftjfm" v-size="[{ max: 500 }]">
 	<div class="with">
-		<button class="_button" @click="with_ = null" :class="{ active: with_ === null }">{{ $t('notes') }}</button>
-		<button class="_button" @click="with_ = 'replies'" :class="{ active: with_ === 'replies' }">{{ $t('notesAndReplies') }}</button>
-		<button class="_button" @click="with_ = 'files'" :class="{ active: with_ === 'files' }">{{ $t('withFiles') }}</button>
+		<button class="_button" @click="with_ = null" :class="{ active: with_ === null }" aria-label="$t('notes')">{{ $t('notes') }}</button>
+		<button class="_button" @click="with_ = 'replies'" :class="{ active: with_ === 'replies' }" aria-label="$t('notesAndReplies')">{{ $t('notesAndReplies') }}</button>
+		<button class="_button" @click="with_ = 'files'" :class="{ active: with_ === 'files' } aria-label="$t('withFiles')">{{ $t('withFiles') }}</button>
```
Johann150, line 6: I'm not sure why these text labels are necessary when they are already the text content of the button.

src/client/components/notes.vue
```diff
@@ -8,7 +8,7 @@
 	<mk-error v-if="error" @retry="init()"/>
 
 	<div v-show="more && reversed" style="margin-bottom: var(--margin);">
-		<button class="_panel _button" ref="loadMore" :disabled="moreFetching" :style="{ cursor: moreFetching ? 'wait' : 'pointer' }">
+		<button class="_panel _button" ref="loadMore" :disabled="moreFetching" :style="{ cursor: moreFetching ? 'wait' : 'pointer' } aria-label="$t('loadMore')">
```
Johann150, line 11: seems more sensible to add this to the `mk-loading` component to me

src/client/components/notes.vue
```diff
@@ -19,7 +19,7 @@
 	</x-list>
 
 	<div v-show="more && !reversed" style="margin-top: var(--margin);">
-		<button class="_panel _button" ref="loadMore" :disabled="moreFetching" :style="{ cursor: moreFetching ? 'wait' : 'pointer' }">
+		<button class="_panel _button" ref="loadMore" :disabled="moreFetching" :style="{ cursor: moreFetching ? 'wait' : 'pointer' } aria-label="$t('loadMore')">
```
Johann150, line 22: same as above

Johann150 (commented):
I think parts of this PR are unnecessary, and deleting `services/drive/add-file.ts` does not make sense.

I will open a separate PR for refactoring the `page.image.vue` component, but I think the rest of this PR is not needed.

PR #6604, opened 2020-07-30
Simplified Chinese Install & Setup Guides Added

Candinya:
## Summary

Simplified Chinese Install & Setup Guides Added


syuilo:
Thank you🥰🥰🥰

PR #6743, opened 2020-10-18
Update switch.vue

okpierre:
## Summary

Removes hard coded margin and moves save button to align with rest of page

misskey-instance-settings-after

misskey-instance-settings-before


syuilo:
Close this for the same reason as https://github.com/syuilo/misskey/pull/6744#issuecomment-729370583. (It has already been fixed!)
Thank you🥰

PR #6744, opened 2020-10-18
Update input.vue

okpierre:
## Summary

change to misskey settings page to give more spacing for save button content above

here's video of the change
misskey-settings-input

@syuilo I added video to show the change


syuilo:
Removing the style will cause problems in a number of other areas. example:

Before
image

After
image

I solved the problem by modifying the style of the settings page.
image

It is generally best to avoid changing the style of commonly used components such as ui/input.
Thanks anyway🙏

PR #6745, opened 2020-10-18
Update style.scss

okpierre:
## Summary

This gives the styling for the footer button

misskey-instance-settings-after

misskey-instance-settings-before


syuilo:
Fixed it by using _card instead of _section in the admin panel.
Thank you anyway🙏

PR #6776, opened 2020-10-28
TOOLS: Created demote tool based on mark-admin.ts

wiredsis:
## Summary

Since demoting admins for now requires using SQL (which is not very fast for people who need fast tools) I created demote-admin.ts based mark-admin.ts. This was because I was dealing with a badly typed password on my admin account and needed to change it.


syuilo:
👍

PR #7015, opened 2020-12-24
Showusers order by updateAt NULL considered as max

### LINKED ISSUE #7014: On admin user page, the order by update time, null should be the last
## 💡 Summary

In the admin user page, when select order by last update time, the null value is conceded as latest update.

## 🙂 Expected Behavior

Show users with recent update

## ☹️ Actual Behavior

full of never post users

## 📝 Steps to Reproduce

1. go to admin
2. user tab
3. order by last update

## 📌 Environment

v11

CGsama:
## Summary

In the admin user page, when select order by last update time, the null value should conceded as oldest update.

Resolve #7014


acid-chicken (approved):

syuilo:
👍👍👍

PR #7042, opened 2020-12-30
Rename ht-HT.yml to th-TH.yml

okpierre:
## Summary

Just a simple and quick rename to fix typo

https://github.com/syuilo/misskey/issues/7030


syuilo:
That file is generated automatically, so there is no point in editing it manually. I fixed it by updating crowdin settings.
Thanks anyway!

okpierre:
Ok thanks!

PR #7112, opened 2021-01-23
Parse followers audience from followers URI

InspectorCaracal:
Adds followersUri to user model and fetches from followers attribute in ActivityPub actor model.

This allows parsing the audience to use the actor's defined followers collection instead of the hardcoded path.


src/remote/activitypub/audience.ts
```diff
@@ -88,6 +88,6 @@ function isPublic(id: string) {
 function isFollowers(id: string, actor: IRemoteUser) {
 	return [
 		`${actor.uri}/followers`,
-		// actor.followerUri,	// TODO
+		actor.followersUri,
```
mei23: If it's null, I don't want to use it.

src/remote/activitypub/models/person.ts
```diff
@@ -330,6 +331,7 @@ export async function updatePerson(uri: string, resolver?: Resolver | null, hint
 		lastFetchedAt: new Date(),
 		inbox: person.inbox,
 		sharedInbox: person.sharedInbox || (person.endpoints ? person.endpoints.sharedInbox : undefined),
+		followersUri: person.followers,
```
mei23: I think it's better to do the same as createPerson above.

InspectorCaracal: good catch, thank you!

src/remote/activitypub/audience.ts
InspectorCaracal: good point!

to fix this i have redone `isFollowers` to compare `id` to either `followersUri` or the default

an account should only have one followers collection, so the default should only be as a backup

src/remote/activitypub/audience.ts
```diff
@@ -86,8 +86,7 @@ function isPublic(id: string) {
 }
 
 function isFollowers(id: string, actor: IRemoteUser) {
-	return [
-		`${actor.uri}/followers`,
-		// actor.followerUri,	// TODO
-	].includes(id);
+	return (
+		id == (actor.followersUri ? actor.followersUri : `${actor.uri}/followers`)
+	);
```
mei23: Please use `===`
And, maybe below is enough.
```suggestion
  return id === (actor.followersUri || `${actor.uri}/followers`);
```

InspectorCaracal: `||` will give a result of either True or False, not the URI, but I will use `===`

mei23: > `||` will give a result of either True or False

That's no problem
image

mei23: Perhaps its specification is only equality operators.
image

InspectorCaracal: oh! I learned something, thank you!

mei23 (approved):
カッコがもう1段不要だけどめんどくさいからいいわ

syuilo:
Thank you guys🥰

PR #7126, opened 2021-01-25
using set instead of array for search

### LINKED ISSUE #6905: using set instead of array for search
In websocket channels arrays are used to keep followings and then be searched for a match with `includes()`.
https://github.com/syuilo/misskey/blob/e6cb07bef4a82ef83c6b80c4d178a0218a8d3a22/src/server/api/stream/channels/home-timeline.ts#L25

I think if we use sets it will be thousands of time faster when the followings of a user increases.
I created a sample gist to show the difference:
https://gist.github.com/ehsanjavadynia/5e744c7dc354d9f22b946509d3642df2

ehsanjavadynia:
## Summary

Resolve #6905 
- replaced all string arrays in connection with string sets (to improve performance)
- change "includes" to "has" function where sets are used


syuilo:
Thank you!

PR #7132, opened 2021-01-29
Update recaptcha's host

TennousuAthena:
Changed to recaptcha.net because google.com are blocked in some region

## Summary

   Hi there. "google.com" is blocked in some regions like China mainland, as a result, users from those area cannot access Misskey instance if Recaptcha is on. Alternatively, "recaptcha.net" is available.

  Thanks.

### DIFF
--- a/src/client/components/captcha.vue
+++ b/src/client/components/captcha.vue
@@ -57,7 +57,7 @@ export default defineComponent({
 		src() {
 			const endpoint = ({
 				hcaptcha: 'https://hcaptcha.com/1',
-				grecaptcha: 'https://www.google.com/recaptcha',
+				grecaptcha: 'https://www.recaptcha.net/recaptcha',
 			} as Record<PropertyKey, unknown>)[this.provider];
 
 			return `${typeof endpoint == 'string' ? endpoint : 'about:invalid'}/api.js?render=explicit`;


rinsuki:
for reviewers: https://developers.google.com/recaptcha/docs/faq#can-i-use-recaptcha-globally

mei23:
Can this really fix it? 
https://www.recaptcha.net/recaptcha/api.js seems to refer to www.google.com inside.

This solution seems to be broken. 
https://groups.google.com/g/recaptcha/c/XBhetkVgh84/m/pQTC5Sx6AgAJ

rinsuki:
> Our system will pick the appropriate url to use. If you are testing it outside of china, it will use www.google.com.

TennousuAthena:
> Can this really fix it?
> https://www.recaptcha.net/recaptcha/api.js seems to refer to [www.google.com](http://www.google.com) inside.
> 
> This solution seems to be broken.
> https://groups.google.com/g/recaptcha/c/XBhetkVgh84/m/pQTC5Sx6AgAJ

According to [Greatfire](https://en.greatfire.org/https/www.recaptcha.net/recaptcha/api.js), it usually works in China

mei23:
Oh, then OK?

The server side still references www.google.com, is that okay?
It may be OK as long as the server is outside mainland China.

mei23 (approved):
I tested it, but it seems that there is no problem outside the existing mainland China. 
The server-side host doesn't seem to need to be changed for now, but I'll do the same in the future.

syuilo:
Thank you guys!

PR #7221, opened 2021-02-17
Improve image view by using viewerjs(Need Help About iOS )

### LINKED ISSUE #7010: Image Zooming in picture view
## Summary

When viewing a picture in timeline at misskey, if the picture is large enough, I can't see any details on it(especially via mobile).Because of that, I hope to add a Zoom Button that can help me adjust the size myself.
Thank you very much!

ybw2016v:
## Summary

Improve the image view to add the function that  the pictures can be resized by both touch and wheel by using [viewerjs](https://fengyuanchen.github.io/viewerjs/).

Here is the demo on  PC and mobile.

PC

Mobile Phone

Resolve #7010


src/client/components/image-viewer.vue
```diff
@@ -31,11 +33,41 @@ export default defineComponent({
 	},
 
 	emits: ['closed'],
-
+	mounted ()
+	{
+		this.createViewer();
+	},
+	destroyed (){
```
syuilo, line 44:
```suggestion
	unmounted() {
```

src/client/components/image-viewer.vue
```diff
@@ -31,11 +33,41 @@ export default defineComponent({
...
+					size: 'large',
+				},
+				next: 0,
+				rotateLeft: 4,
+				rotateRight: 4,
+				flipHorizontal: 4,
+				flipVertical: 4,};
+			this.$viewer = new Viewer(this.$el, {title: [1, ()=>title],navbar:0,toolbar:toolbar});
+			this.$viewer.show();
+			this.$refs.modal.close(); //Fix the bug that one same image show twice
```
syuilo, line 72: この image-viewer.vue を削除し、親の media-image.vue で viewer.js を呼び出すようにすると良さそうです。

syuilo:
iOS でテストしましたが、動作が不安定だったり、画像を保存できないなど、viewer.js のユーザビリティが悪いようです。

src/client/components/image-viewer.vue
ybw2016v: Thanks. I will try it soon.

ybw2016v:
> iOS でテストしましたが、動作が不安定だったり、画像を保存できないなど、viewer.js のユーザビリティが悪いようです。

This problem seems like this [issue](https://github.com/fengyuanchen/viewerjs/issues/348), I will try to fix it but unfortunately I don't  have any iOS device to test it.

ybw2016v:
**need help**
to find out what cause it failed on iOS device and test if the way to fix it work.

In the last commit ,  `slideOnTouch` is set `false` , but I don't konw whether it works or not.

ybw2016v:
The `Viewer.js` can't work well in the environment of iOS and it is reported that `Viewer.js` can't resize images by touch  for iOS browser even if in the  `Viewer.js`'s official demo. The reason that cause this problem might be that different browsers especially in different types of device have different reaction to the touch event. But I can't open an issue in `Viewer.js`'s repository because of lack of the detail , such as  the version of browser and the device information. `Viewer.js` is still in the status of active developing , which means that an appropriate issue with key information may accelerate the fix of this problem.

It is also found that `Viewer.js` can not  deal with the pictures with blurhash when I try to replace ImageViewer to viewerjs. The reason might be that `Viewer.js` can not change the url of the image dynamically. In order to resolve this problem, the `media-image.vue` need to rewrite almostly which might cause more bugs.(I tried but failed )

So a temporary and compromise way to deal with these problems is to keep the `image-viewer` to compat iOS and use different way to show the image according to the `user-agent` .
  
Anyway to compat different browsers and devices is one of the most difficult work in front-end development. Thanks.

Johann150:
I think this has been solved a different way now, see #7696.

syuilo:
PhotoSwipeを使用する実装を行いました。
Thanks anyway 🙏

PR #7241, opened 2021-02-21
 feature: add video.js suport for inline video playing

Candinya:
## Summary

Add [video.js](https://github.com/videojs/video.js) support for inline video playing.
Inline styles and some display logics are rewritten for better performance.
Feel free to let me know if there's any questions. Always willing to improve. 😃 

( this is the changed PR branch, sorry for the inconvenience)

### DIFF
--- a/src/client/components/media-video.vue
+++ b/src/client/components/media-video.vue
@@ -1,29 +1,28 @@
 <template>
-<div class="icozogqfvdetwohsdglrbswgrejoxbdj" v-if="hide" @click="hide = false">
-	<div>
-		<b><Fa :icon="faExclamationTriangle"/> {{ $ts.sensitive }}</b>
-		<span>{{ $ts.clickToShow }}</span>
+<div class="kkjnbbplepmiyuadieoenjgutgcmtsvu">
+	<div class="icozogqfvdetwohsdglrbswgrejoxbdj" v-show="hide" @click="onShow">
+		<div>
+			<b><Fa :icon="faExclamationTriangle"/> {{ $ts.sensitive }}</b>
+			<span>{{ $ts.clickToShow }}</span>
+		</div>
 	</div>
-</div>
-<div class="kkjnbbplepmiyuadieoenjgutgcmtsvu" v-else>
-	<i><Fa :icon="faEyeSlash" @click="hide = true"/></i>
-	<a
-		:href="video.url"
-		rel="nofollow noopener"
-		target="_blank"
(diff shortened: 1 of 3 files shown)


rinsuki:
Is there any reason to prefer video.js rather then `<video controls>`?

I think, video.js works fine in desktop, but in smartphone, experience may not good.

Candinya:
> Is there any reason to prefer video.js rather then `<video controls>`?
> 
> I think, video.js works fine in desktop, but in smartphone, experience may not good.

It has more customizable options than original `<video>` tag, like a thumbnail pic (generated by ffmpeg for first glance), pause operation API (which I've used to prevent video from playing after the element visibility is set to hidden), community plugins support and so on.

Awful experienes on smartphone could be caused by default skin (like those built-in tiny-tiny buttons), maybe it's recommend to make a misskey-styled third-party skin for the player 🤔

rinsuki:
actually thumbnail picture and play/pause API is available with `<video controls>` tag too (you can use from `this.$refs.VIDEO_TAG_REF_NAME`)

src/client/components/media-video.vue
```diff
@@ -35,71 +34,93 @@ export default defineComponent({
...
 	},
 	created() {
 		this.hide = (this.$store.state.nsfw === 'force') ? true : this.video.isSensitive && (this.$store.state.nsfw !== 'ignore');
 	},
+	mounted() {
+		this.$nextTick(() => {
+			this.player = videojs(this.$refs.videoCast, {
+				autoplay: false,
+				controls: true,
+				preload: "metadata",
```
rinsuki: I think this may incrases traffic of media server because browser preloading all videos (including user dont want to play), so I suggest `preload: "none"`

src/client/components/media-video.vue
```diff
@@ -35,71 +34,93 @@ export default defineComponent({
...
+				playbackRates: [0.5, 0.7, 1.0, 1.5, 2.0, 2.5, 3.1],
```
rinsuki: `3.1` is may `3.0`?

Candinya:
Mentioned issues resolved, thanks for the suggestions.
Also I'd like to test the original `video` tag, maybe it's a more ideal solution for users considering both light-weight and multiplatform friendly support. Working in progress 🚧

Candinya:
Changed to #7242 using original HTML `<video>` tags. Thanks @rinsuki for the advice!

PR #7335, opened 2021-03-12
Use integrity for scripts hosted on the CDN

### LINKED ISSUE #7333: /api-doc: pin redoc bundle version and add integrity attribute
## 💡 Summary

Including a `<script>` with `@next` from a third-party service is potentially dangerous. We should pin it to a specific version and make use of [SRI](https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) to validate its integrity.

https://github.com/syuilo/misskey/blob/5b965349a5abed481ef0d6649c02184f7b774036/assets/redoc.html#L22

## 🙂 Expected Behavior

`redoc.standalone.js` is pinned and guaranteed to be the expected version.

## ☹️ Actual Behavior

`redoc.standalone.js` could change at any time and may become malicious if the npm package or jsDelivr get compromised.

mei23:
## Summary
Fix #7333 

https://www.jsdelivr.com/package/npm/redoc?path=bundles&version=2.0.0-rc.50

### DIFF
--- a/assets/redoc.html
+++ b/assets/redoc.html
@@ -19,6 +19,6 @@
 	</head>
 	<body>
 		<redoc spec-url="/api.json" expand-responses="200" expand-single-schema-field="true"></redoc>
-		<script src="https://cdn.jsdelivr.net/npm/redoc@next/bundles/redoc.standalone.js"> </script>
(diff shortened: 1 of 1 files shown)


syuilo:
🆒

PR #7463, opened 2021-04-19
Update index.ts

### LINKED ISSUE #7461: robots.txt is missing
## 💡 Summary

Misskey used to have a robots.txt. Since a couple of updates the robots.txt doesn't exist anymore.

## 🙂 Expected Behavior

Building Misskey should create a robots.txt at `/built/assets/robots.txt`.

## ☹️ Actual Behavior

When visiting for example https://misskey.de/robots.txt you'll get an error message... (1)
`ENOENT: no such file or directory, stat '/home/misskey/misskey/built/assets/robots.txt'`

When I check the file path I see that there is no robots.txt.

It could be that you only get the error when you `yarn clean` before you build the latest version.

(1) As a workaround a have restored the file from a backup.

## 📝 Steps to Reproduce

0. maybe you need to run `yarn clean`
1. then build the latest Misskey version
2. check if a `robots.txt` has been created

## 📌 Environment

Misskey v12.77.1
Node.js v14.16.1
PostgreSQL v12.6 (Ubuntu 12.6-0ubuntu0.20.04.1)
Redis v5.0.7

okpierre:
## Summary

Fix https://github.com/misskey-dev/misskey/issues/7461

### DIFF
--- a/src/server/web/index.ts
+++ b/src/server/web/index.ts
@@ -99,7 +99,7 @@ router.get('/manifest.json', require('./manifest'));
 
 router.get('/robots.txt', async ctx => {
 	await send(ctx as any, '/robots.txt', {
-		root: assets
+		root: staticAssets
 	});
 });
 


syuilo:
Thanks!

PR #7509, opened 2021-05-09
Fix hide button of image

Candinya:
## Summary

Fix hide button (of image) which has been covered by image layer.
By changing the element's order.


syuilo:
👍

PR #7528, opened 2021-05-22
Add Indonesian language to locales index

sandycorzeta:
## Summary

Adding Indonesian language into locales index to make it available in the settings page


syuilo:
👍

PR #7560, opened 2021-06-08
Add MRFs

nullobsi:
## Summary

Inspired by Pleroma's [MRFs](https://docs-develop.pleroma.social/backend/configuration/mrf/).

Using JS files under `.config/mrf`, incoming activities can be more finely filtered and even modified. It could be used to force post visibility, or mark all incoming media as NSFW from certain instances.

I have set it as a draft, as I feel like there are some improvements and feedback that could be made.


nullobsi:
# TODO
- [ ] 管理インターフェイス（プラグインみたいのconfigオプション)
- [ ] APIを追加
- [ ] resolveで使う
- [ ] configはDBで保存する
- [ ] workerでconfigをキャッシュする

src/queue/processors/inbox.ts
```diff
@@ -144,5 +154,6 @@ export default async (job: Bull.Job<InboxJobData>): Promise<string> => {
 
 	// アクティビティを処理
 	await perform(authUser.user, activity);
+	await perform(authUser.user, activity);
```
sousuke0422, line 157: Is there any significance to the duplication here?

nullobsi: don't think so, i wonder if it was added when i merged the updates

Johann150:
I think this is related to #4529

nullobsi:
I'd have to rework this, but here are my ideas for MRFs:

- Allow the .JS files to be uploaded/managed from the Control Panel UI
  - Security issues? It would need a big warning that these files can access all node.js APIs
  - Text file editor for these scripts
- Management of MRF options
  - JS scripts expose a `properties` object that defines the various options to be displayed in UI
  - (filtered words, allowed users, blocked users, etc...)
  - Stored in DB? Or stored in the config folder?
  - These should probably be under /api/meta when requesting as admin
  - toggle on/off
- Performance
  - Needs to be quick and easy to hook.
  - AiScript could make it easier (similar to plugins on Frontend); but is it too slow for server-side use
  - Needs to be redundant, if an MRF has an error it should not stop all activities from being accepted
- API
  - js scripts expose properties object for details like name, description, configurable options
  - Also for where the hooks should run
  - seperate function possibility for each activitypub object type; Create, Update, Announce, or just one generic hook with its own logic (as this PR worked before)
  - Expose useful utilities like resolver, isNote(), etc? Require paths could change, these could be added in a global object `mrf.resolve()`, `mrf.isOrderedCollection()`, etc
  - properties.onRecieveCreate, properties.onActorUpdate, hook functions

Johann150:
If you have to rework this anyway I think it makes sense to close this PR and have further discussion in the related issue linked above. We can keep this implementation idea in mind. When the discussion has matured you can open another PR.

PR #7588, opened 2021-06-27
Rich welcome content

Skehmatics:
## Summary

Add rich content (ie. media, polls, and mfm) into the featured notes shown on the welcome screen. Many posts that may make it into featured may be very confusing when this context is removed.

Since images especially can take up quite a lot of space on the screen, a scrolling animation is also added if the featured section is longer than the browser window.

Peek 2021-06-26 22-48

This is my first time touching vue, let me know if there's a better way to do this!

### DIFF
--- a/src/client/components/poll.vue
+++ b/src/client/components/poll.vue
@@ -10,7 +10,7 @@
 			</span>
 		</li>
 	</ul>
-	<p>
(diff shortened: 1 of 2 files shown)


syuilo:
👍👍👍

PR #7618, opened 2021-08-11
Adds ServiceWorker section to manage.en.md

0hlov3:
Adds short description for activating ServiceWorker on misskey in the manage.en.md.

# What

# Why

# Additional info (optional)

### DIFF
--- a/docs/manage.en.md
+++ b/docs/manage.en.md
@@ -12,3 +12,10 @@ e.g.
 ``` shell
 node built/tools/mark-admin @syuilo
 ```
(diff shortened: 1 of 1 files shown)


mei23:
Thank you for your contribution.
Because the operation of docs/ changes and localization is not clear. It may take some time. 

Below is my feelings 

```
# non Docker
npx web-push generate-vapid-keys

# Docker
docker-compose run --rm web npx web-push generate-vapid-keys
```
may be better.

Issue #7627, opened 2021-08-13
Error: invalid Actor: summary invalid-range


ghost:
Sorry, I can't read Japanese, so I don't know if this error has already been reported.

## 💡 Summary

Sometimes I will get the error "Error: invalid Actor: summary invalid-range" in logs when a new remote post is received.

## 🙂 Expected Behavior

No error occurs.

## ☹️ Actual Behavior

```js
04:17:17 INFO *	[remote ap]	Creating the Note: https://poa.st/objects/44cc7ccb-f458-428f-b6c6-80638c468c1b
04:17:17 WARN *	[remote ap]	Error in announce target https://poa.st/objects/44cc7ccb-f458-428f-b6c6-80638c468c1b - Error: invalid Actor: summary invalid-range
Error: invalid Actor: summary invalid-range
    at validate (/misskey/built/remote/activitypub/models/person.js:48:19)
    at validateActor (/misskey/built/remote/activitypub/models/person.js:54:5)
    at createPerson (/misskey/built/remote/activitypub/models/person.js:102:20)
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async Object.resolvePerson (/misskey/built/remote/activitypub/models/person.js:333:12)
    at async createNote (/misskey/built/remote/activitypub/models/note.js:78:19)
    at async Object.resolveNote (/misskey/built/remote/activitypub/models/note.js:252:16)
    at async Object.default_1 [as default] (/misskey/built/remote/activitypub/kernel/announce/note.js:35:22)
```

## 📝 Steps to Reproduce

I don't know how to reproduce this, maybe it is possible to reproduce if you try to fetch the note with the url above.

## 📌 Environment

Misskey 12.85.1 in Docker

mei23:
Fixed with #7629

PR #7629, opened 2021-08-13
fix: truncate user information if it is too long

### LINKED ISSUE #6769: Cant open profile. Error [object Object]
## 💡 Summary

When i try to open profile https://gentoo.live/sathariel i have error [object Object]

levels.error

`[api] Internal error occurred in ap/show: invalid json response body at https://gentoo.live/sathariel reason: Unexpected token < in JSON at position 0`

levels.warning

`[api] failed to resolve remote user: Error: invalid person: invalid name`

Nickname - :unicursalpink: Sathy de los Muertos 3: La venganza :chaos_pink:

## 🙂 Expected Behavior

Profile opened.

## ☹️ Actual Behavior

When i search @sathariel@gentoo.live

When i search https://gentoo.live/sathariel

## 📝 Steps to Reproduce

1. Click Search button.
2. Paste https://gentoo.live/sathariel or @sathariel@gentoo.live

## 📌 Environment

Debian GNU/Linux 10 (buster) x86_64
Misskey v12.50.0
Node.js v14.13.1
PostgreSQL v13.0 (Debian 13.0-1.pgdg100+1)
Redis v5.0.3

Johann150:
# What
Some AP software allows for user names or summaries to be very long.
Misskey can not handle this and the profile page can not be opened and
no activities from such users can be seen.

Instead, the user name and summary are cut off after the maximum length.
so misskey can still process the activities of the profile.

fixes #6769 

# Why
Without this, the profiles and the user notes, reactions can not be federated.

Instead of failing completely, this allows to federate other information.

# Additional info (optional)
This can remove some information from the user summary, but Misskey already shows a warning for profiles from remote instances. 

This is also [part of `misskey-extra`](https://github.com/JakeMBauer/Misskey-Extras/blob/main/patches/README.md#Truncate-Profiles), and is already used on some instances.

### DIFF
--- a/src/remote/activitypub/models/person.ts
+++ b/src/remote/activitypub/models/person.ts
@@ -56,7 +56,7 @@ function validateActor(x: IObject, uri: string): IActor {
 	};
 
 	const truncate = (input: string | undefined, size: number) => {
-		if(!input || input.length <= size){
+		if (!input || input.length <= size) {
 			return input;
 		} else {
 			return input.substring(0, size);
--- a/src/remote/activitypub/models/person.ts
+++ b/src/remote/activitypub/models/person.ts
@@ -31,6 +31,9 @@ import { normalizeForSearch } from '@/misc/normalize-for-search';
 
 const logger = apLogger;
 
+const nameLength = 128;
+const summaryLength = 2048;
+
 /**
  * Validate and convert to actor object
  * @param x Fetched object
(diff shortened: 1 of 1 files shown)


Johann150:
also solves #7627

src/remote/activitypub/models/person.ts
```diff
@@ -31,6 +31,9 @@ import { normalizeForSearch } from '@/misc/normalize-for-search';
 
 const logger = apLogger;
 
+const nameLength = 128;
+const summaryLength = 2048;
```
Johann150, line 35: maybe these should be stored in `misskey/src/misc/hard-limits.ts` ?

syuilo: I will to do so!

syuilo:
🙏

mei23:
Because the DB column is unextended.
https://github.com/misskey-dev/misskey/blob/b5cadeca2e02f88d46625f27c3fcc796e45c2526/src/models/entities/user-profile.ts#L32-L36

Not only must it pass validate, but it must also be truncate when inserting/updating.
https://github.com/misskey-dev/misskey/blob/b5cadeca2e02f88d46625f27c3fcc796e45c2526/src/remote/activitypub/models/person.ts#L173

PR #7657, opened 2021-08-18
fix: api/users

### LINKED ISSUE #7656: Browser showing Internal Server Error /api/users
## 💡 Summary

Internal Server Error /api/users

## 🙂 Expected Behavior

No error when viewing the `https://www.example.com/explore` as user

## ☹️ Actual Behavior

Showing "An error occurred" error message in browser
error code 500 for `https://www.example.com/api/users`

## 📝 Steps to Reproduce

1. Login first 
2. Visit `https://www.example.com/explore`
3. An error occurred message with icon appears

## 📌 Environment

Misskey 12.88.0 PostgreSQL 10.15

Johann150:
# What

the wrong function is used here https://github.com/misskey-dev/misskey/blob/881b914c6a7bdf73ed16194e71aa779b27de074e/src/server/api/endpoints/users.ts#L93
Instead `generateBlockQueryForUsers` should be used. This leads to an SQL query error if a logged in user uses the `api/users` endpoint.

# Why

fix #7656 (caused by <https://github.com/misskey-dev/misskey/commit/b690c410a37b129e42fb35d759dd0dedc2dca1f7#diff-44772bc244023699b872cf80bb8ce51cd97c2c771d2c435ab7cfd8f07f5caedc>)


imashen:
Yeah！ I agree, absolutely right

syuilo:
Thanks!

imashen:
Wuhoooo~
I'm glad you finally fixed this bug. I publicized the misskey project in China. Recently, many programmers and friends have contacted this good project. I will always support you and sponsor it in the future. By the way, when is the next version scheduled to be updated

syuilo:
I expect it to be released in the next few days.

Issue #7632, opened 2021-08-15
Unable to load more Mentions


okpierre:
## 💡 Summary

Unable to load @ Mentions page when loading more
`https://www.example.com/my/mentions`

## 🙂 Expected Behavior

Page should retrieve more mentions from user

## ☹️ Actual Behavior

Page keeps showing spinning icon and then load more is shown again

## 📝 Steps to Reproduce

1. Go to menu and select Mentions
2. First page loads ok
3. Scroll to the bottom of page to activate load more 
4. Spinning icon shows and the page keeps loading and nothing happens

## 📌 Environment

Misskey 12.87.0 PostgreSQL 10.15

cliqsocial:
Our community is experiencing the same problem -> https://cliq.social/notes/8pibdqhgzb

Timeline = read more ok
Drive = read more ok

Mentions = `[error] upstream prematurely closed connection while reading response header from upstream request: "POST /api/notes/mentions HTTP/2.0", upstream: http://127.0.0.1:3000/api/notes/mentions"`

@syuilo that was the error from the log file. Anything else we can provide to track down this issue?

Johann150:
I suspect this could be some sort of timeout (e.g. <https://stackoverflow.com/a/36488757>). Could you try increasing the proxy timeout in nginx?

cliqsocial:
> I suspect this could be some sort of timeout (e.g. https://stackoverflow.com/a/36488757). Could you try increasing the proxy timeout in nginx?

Thanks for the link @Johann150 we already have it set for 120s which I think is already too long. 

Do you know why [timeline] and [drive] **does not** time out? What's special about the [Mention] functionality?

Johann150:
This was just the first idea that came to mind, I am not intimately familiar with the differences.
One difference is that the timeline and drive come from their own database tables, while mentions have to be filtered from the `note` table.

P.S. its unnecessary to @ mention if you are replying to people, it only creates an annoying notification

tamaina:
The PostgreSQL version may be the cause. Please try a newer version.

The setup.md says ">= 10", but 10 might not work.  
syuilo also says it needs 12.   
https://misskey.io/notes/8pm4bc1aar

I'm running 13.4 that works well.

syuilo:
12 is installed on Docker, but 13 is preferred.

archaeme:
I'm using 13 on my server and I have the same error.

normikoto:
[...]

PR #7675, opened 2021-08-21
Update setup.en.md

okpierre:
# What
Updated PostgreSQL version to 12  

# Why
See https://github.com/misskey-dev/misskey/issues/7632#issuecomment-901653826


syuilo:
👍

PR #7682, opened 2021-08-22
enhance: add setting to keep content warning

### LINKED ISSUE #6605: Retain CW title when reply on a CW note
## Summary

When there is a note with CW comment/title, it has a reason why the note is behind a CW mark and most of the cases the reply should be in CW too. To add the CW with the same comment/title to keep the thread we have to copy the previous one manually, but removing it if we don't want to CW the reply would be much easier if that would be the default behavior.

So, expected behavior: Reply on CW note automatically turns on CW for the reply note and populates the Comment field with the original CW Comment field value. If we don't want to reply with CW, deleting/turning off the CW flag is much easier.

Johann150:
# What

Add a setting to use `keepCw` when replying to a post.

# Why

Keeping the content warning was already implemented and working, but there was no setting to enable it.
fix #6605


syuilo:
👍

ThatOneCalculator:
Can we have this be on by default for all users? Or let admins force it on by default for all users?

PR #7685, opened 2021-08-23
feat: Added a toggle for showing replies in timelines

### LINKED ISSUE #3712: Option to show replies in timelines
# Summary

Currently it seems that replies do not show in the global timeline and there is no option for them to do so. It would be helpful and appreciated to have a toggle to be able to see them in the global timeline, as without that you miss conversations that are not on your local instance unless you follow all people involved.

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67847655-option-to-show-replies-in-timelines?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github).
</bountysource-plugin>

jigglypuff-fan:
# What

This PR adds a user-level toggle for showing all replies on all timelines. What this means is if the user is following user A, they will see every post user A makes, including replies to other users.

# Why

This behavior is designed to mimic Pleroma's default timeline behavior, and is intended for the comfort of the large amount of users switching to Misskey from Pleroma.

# Additional info (optional)

Couldn't run tests, unfortunately. Looks like `test.yml` is missing from the repository.

Closes #3712


locales/en-US.yml
```diff
@@ -138,6 +138,8 @@ cacheRemoteFilesDescription: "When this setting is disabled, remote files are lo
 flagAsBot: "Mark this account as as bot"
 flagAsBotDescription: "Enable this option if this account is controlled by a program. If enabled, it will act as flag for other developers to prevent endless interaction chains with other bots and adjust Misskey's internal systems to treat this account as a bot."
 flagAsCat: "Mark this account as a cat"
+flagPleroma: "Show replies in timeline"
+flagPleromaDescription: "Show users replying to other users in the timeline"
```
Johann150: I think the ja-JP file is used as the base for translations so these lines should be added there too.

src/server/api/common/generate-replies-query.ts
```diff
@@ -1,22 +1,24 @@
 import { User } from '@/models/entities/user';
 import { Brackets, SelectQueryBuilder } from 'typeorm';
 
-export function generateRepliesQuery(q: SelectQueryBuilder<any>, me?: { id: User['id'] } | null) {
-	if (me == null) {
+export function generateRepliesQuery(q: SelectQueryBuilder<any>, me: any) {
+	const userId = me['id'] ? me['id'] : null;
+	const plTimeline: boolean = me['pleromaTimeline'] ? me['pleromaTimeline'] : false;
```
Johann150: I think using `||` or `??` (null coalescing) instead of the ternary operator might be cleaner / less duplication.

Johann150 (commented):
Not sure if there is a particular reason why this can not sufficiently be done with an Antenna with source "Notes from followed users" and "include replies" enabled.

jigglypuff-fan:
@Johann150 the antenna route is a workaround. It does not include your own posts, for one. And it doesn't affect other timelines. There is no way currently to see replies on Social or Local timelines. In addition to the solution being uncomfortable, using an antenna as a timeline versus using the actual timelines.

locales/en-US.yml
robflop: This is correct. Keys must be included in ja-JP.

robflop:
I feel like it would better to name this neutrally instead of refering to Pleroma. "Pleroma-style" is what it could be compared to when casually talking, but the actual setting and such should be neutral descriptions, imo.

locales/en-US.yml
jigglypuff-fan: Thanks! I just copied them over as-is to the JP file, though, as I don't know a word of Japanese.

src/server/api/common/generate-replies-query.ts
jigglypuff-fan: Thank you! Sorry I'm not really a JS/TS dev, so I wasn't aware it had coalesce operators. Changed.

src/models/entities/user.ts
```diff
@@ -232,6 +232,12 @@ export class User {
 	})
 	public token: string | null;
 
+	@Column('boolean', {
+		default: false,
+		comment: 'Whether to use the pleroma-style timelines'
```
robflop: Pleroma reference remaining here

migration/1629659466000-AddPleromaTL.ts
```diff
@@ -0,0 +1,15 @@
+import {MigrationInterface, QueryRunner} from "typeorm";
+
+export class addPleromaTL1629659466000 implements MigrationInterface {
```
robflop: Pleroma reference remaining here (and in filename)

migration/1629659466000-AddPleromaTL.ts
```diff
@@ -0,0 +1,15 @@
...
+    name = 'addPleromaTL1629659466000'
```
robflop: Pleroma reference remaining here

robflop (commented):
Found some more places with references to Pleroma.

src/models/entities/user.ts
jigglypuff-fan: Resolved in latest commit

migration/1629659466000-AddPleromaTL.ts
jigglypuff-fan: Resolved in latest commit

migration/1629659466000-AddPleromaTL.ts
jigglypuff-fan: Resolved in latest commit

locales/ja-JP.yml
```diff
@@ -139,6 +139,8 @@ flagAsBot: "Botとして設定"
 flagAsBotDescription: "このアカウントがプログラムによって運用される場合は、このフラグをオンにします。オンにすると、反応の連鎖を防ぐためのフラグとして他の開発者に役立ったり、Misskeyのシステム上での扱いがBotに合ったものになります。"
 flagAsCat: "Catとして設定"
 flagAsCatDescription: "このアカウントが猫であることを示す場合は、このフラグをオンにします。"
+flagShowTimelineReplies: "Show replies in timeline"
```
robflop:
```suggestion
flagShowTimelineReplies: "タイムラインにノートへの返信を表示する"
```

locales/ja-JP.yml
```diff
@@ -139,6 +139,8 @@ flagAsBot: "Botとして設定"
...
+flagShowTimelineRepliesDescription: "Show users replying to other users in the timeline"
```
robflop:
```suggestion
flagShowTimelineRepliesDescription: "オンにすると、タイムラインにユーザーのノート以外にもそのユーザーの他のノートへの返信を表示する。"
```

robflop (commented):
Here's a suggestion for the japanese text. I'm not fluent in japanese, but these should be okay.

locales/ja-JP.yml
robflop: In english: "When turned on, will not only show notes of a user, but also replies of that user to other notes".

locales/ja-JP.yml
robflop: In english: "Show replies to notes on the timeline".

locales/en-US.yml
robflop: See below comment, i've suggested japanese strings.

locales/ja-JP.yml
jigglypuff-fan: Added both, thanks :heart:

locales/ja-JP.yml
```diff
@@ -139,6 +139,8 @@ flagAsBot: "Botとして設定"
 flagAsBotDescription: "このアカウントがプログラムによって運用される場合は、このフラグをオンにします。オンにすると、反応の連鎖を防ぐためのフラグとして他の開発者に役立ったり、Misskeyのシステム上での扱いがBotに合ったものになります。"
 flagAsCat: "Catとして設定"
 flagAsCatDescription: "このアカウントが猫であることを示す場合は、このフラグをオンにします。"
+flagShowTimelineReplies: "タイムラインにノートへの返信を表示する"
+flagShowTimelineRepliesDescription: "オンにすると、タイムラインにユーザーのノート以外にもそのユーザーの他のノートへの返信を表示する。"
```
rinsuki:
```suggestion
flagShowTimelineRepliesDescription: "オンにすると、タイムラインにユーザーのノート以外にもそのユーザーの他のノートへの返信を表示します。"
```

in "description" context, "〜します" looks nice than "〜する".

src/server/api/common/generate-replies-query.ts
```diff
@@ -1,22 +1,24 @@
 import { User } from '@/models/entities/user';
 import { Brackets, SelectQueryBuilder } from 'typeorm';
 
-export function generateRepliesQuery(q: SelectQueryBuilder<any>, me?: { id: User['id'] } | null) {
-	if (me == null) {
+export function generateRepliesQuery(q: SelectQueryBuilder<any>, me: any) {
```
rinsuki: you shouldn't use TypeScript's `any` in normal case.

```suggestion
export function generateRepliesQuery(q: SelectQueryBuilder<any>, me?: Pick<User, 'id' | 'showTimelineReplies'> | null) {
```

src/server/api/common/generate-replies-query.ts
```diff
@@ -1,22 +1,24 @@
...
+	const userId = me['id'] ?? null;
+	const showReplies: boolean = me['showTimelineReplies'] ?? false;
+	if (userId == null) {
```
rinsuki: in TypeScript, `else` cause of `if (me == null)`, you can use `me` as non-null types. it means dont need this variables.
```suggestion
	if (me == null) {
```

src/server/api/common/generate-replies-query.ts
```diff
@@ -1,22 +1,24 @@
...
-	} else {
+	} else if (!showReplies) {
 		q.andWhere(new Brackets(qb => { qb
 			.where(`note.replyId IS NULL`) // 返信ではない
-			.orWhere('note.replyUserId = :meId', { meId: me.id }) // 返信だけど自分のノートへの返信
+			.orWhere('note.replyUserId = :meId', { meId: userId }) // 返信だけど自分のノートへの返信
 			.orWhere(new Brackets(qb => { qb // 返信だけど自分の行った返信
 				.where(`note.replyId IS NOT NULL`)
-				.andWhere('note.userId = :meId', { meId: me.id });
+				.andWhere('note.userId = :meId', { meId: userId });
```
rinsuki: after remove `userId` and `showReplies` variables, you also need to update this codes with `me.id` and `me.showTimelineReplies` (sorry, I cant create suggestions by GitHub's limitation, it saying `Applying suggestions on deleted lines is not supported.`)

rinsuki (changes requested):

src/server/api/common/generate-replies-query.ts
jigglypuff-fan: Yeah, sorry about that :sweat_smile: I assumed this wasn't good but TypeScript isn't a language I work with a lot. Thank you!

jigglypuff-fan:
@rinsuki I've applied your suggested changes. If you have any more ideas I'd be happy to hear them :heart:

locales/ja-JP.yml
```diff
@@ -139,6 +139,8 @@ flagAsBot: "Botとして設定"
 flagAsBotDescription: "このアカウントがプログラムによって運用される場合は、このフラグをオンにします。オンにすると、反応の連鎖を防ぐためのフラグとして他の開発者に役立ったり、Misskeyのシステム上での扱いがBotに合ったものになります。"
 flagAsCat: "Catとして設定"
 flagAsCatDescription: "このアカウントが猫であることを示す場合は、このフラグをオンにします。"
+flagShowTimelineReplies: "タイムラインにノートへの返信を表示する"
+flagShowTimelineRepliesDescription: "オンにすると、タイムラインにユーザーのノート以外にもそのユーザーの他のノートへの返信を表示します"
```
robflop: Could you add a full stop at the end here? Specifically, a japanese one, so this: 

```suggestion
flagShowTimelineRepliesDescription: "オンにすると、タイムラインにユーザーのノート以外にもそのユーザーの他のノートへの返信を表示します。"
```

rinsuki (approved):

jigglypuff-fan:
Bumped the version of the migration as to not break instances that ran an earlier version of this patch.

tamaina:
設定項目はProfileではなくClient settings - Generalの中にしたほうがわかりやすい気がする

robflop:
Noticed this since I just got the notif: Remember not to change localization files other than ja-JP.yml - so best if you could revert the changes to the english locale file.

PR #7699, opened 2021-08-24
fix: truncate image descriptions

Johann150:
# What
Properly handle the length restriction on `drive_file.comment` (絵のキャプション) which is a `VARCHAR(512)`.
- Description longer than 512 chars is truncated by the API. (like #7629)
- Show the limited length in the UI, when writing the image description: <details><summary>screenshot: new dialog for editing the image caption</summary></details>

# Why
Long descriptions lead to a fatal error. The file, the note and any replies to that note can not be saved correctly. This means that these notes can not be received.

This is necessary because Mastodon for example allows **1500 chars** in image descriptions (see https://github.com/mastodon/mastodon/issues/11658). In my opinion, the current limit of 512 is okay and does not have to be changed. If the limit gets larger, it can be difficult for people using screen readers (see also https://github.com/mastodon/mastodon/issues/12268).

### DIFF
--- a/src/remote/activitypub/models/image.ts
+++ b/src/remote/activitypub/models/image.ts
@@ -5,6 +5,8 @@ import { fetchMeta } from '@/misc/fetch-meta';
 import { apLogger } from '../logger';
 import { DriveFile } from '@/models/entities/drive-file';
 import { DriveFiles } from '@/models/index';
+import { truncate } from '@/misc/truncate';
+import { DM_MAX_IMAGE_COMMENT_LENGTH } from '@/misc/hard-limits';
 
 const logger = apLogger;
 
@@ -27,6 +29,8 @@ export async function createImage(actor: IRemoteUser, value: any): Promise<Drive
 
 	const instance = await fetchMeta();
 	const cache = instance.cacheRemoteFiles;
+
+	image.name = truncate(image.name, DB_MAX_IMAGE_COMMENT_LENGTH);
 
 	let file = await uploadFromUrl(image.url, actor, null, image.url, image.sensitive, false, !cache, image.name);
 
--- a/src/remote/activitypub/models/image.ts
+++ b/src/remote/activitypub/models/image.ts
@@ -30,9 +30,7 @@ export async function createImage(actor: IRemoteUser, value: any): Promise<Drive
 	const instance = await fetchMeta();
 	const cache = instance.cacheRemoteFiles;
 
-	image.name = truncate(image.name, DB_MAX_IMAGE_COMMENT_LENGTH);
-
-	let file = await uploadFromUrl(image.url, actor, null, image.url, image.sensitive, false, !cache, image.name);
+	let file = await uploadFromUrl(image.url, actor, null, image.url, image.sensitive, false, !cache, truncate(image.name, DB_MAX_IMAGE_COMMENT_LENGTH));
 
 	if (file.isLink) {
 		// URLが異なっている場合、同じ画像が以前に異なるURLで登録されていたということなので、
--- a/src/client/components/media-caption.vue
+++ b/src/client/components/media-caption.vue
@@ -3,10 +3,13 @@
 		<div class="container">
 			<div class="fullwidth top-caption">
 				<div class="mk-dialog">
-					<header v-if="title"><Mfm :text="title"/></header>
+					<header>
+						<Mfm v-if="title" class="title" :text="title"/>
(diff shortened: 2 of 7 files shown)


nullobsi:
It seems the use of the `length` function breaks when there is no image description. (the description is null and the function throws because null is not a string.)
here is a small patch that fixes the issue
```
Index: src/client/components/media-caption.vue
IDEA additional info:
Subsystem: com.intellij.openapi.diff.impl.patch.CharsetEP
<+>UTF-8
===================================================================
diff --git a/src/client/components/media-caption.vue b/src/client/components/media-caption.vue
--- a/src/client/components/media-caption.vue	(revision 089668a2c2bbb24256eb47e260cb64e5793a974b)
+++ b/src/client/components/media-caption.vue	(revision a72362db858c8393ea9bbdf316346ccb59742791)
@@ -86,6 +86,7 @@
 
 	computed: {
 		remainingLength(): number {
+			if (typeof this.inputValue != "string") return DB_MAX_IMAGE_COMMENT_LENGTH;
 			return DB_MAX_IMAGE_COMMENT_LENGTH - length(this.inputValue);
 		}
 	},
```

Johann150:
Thanks. I also noticed that my implementation of truncate was not Unicode aware, so I fixed that too.

mei23 (commented):
For uploads from **local (not from remote)**, I think it should be error instead of truncate.
For example, **local** Note text and user descriptions are also behaviors. 

Possible with validater.
```ts
comment: {
	validator: $.optional.nullable.str.max(DB_MAX_IMAGE_COMMENT_LENGTH),
```

mei23:
https://github.com/misskey-dev/misskey/pull/7629 is processing only remote objects.

mei23:
@syuilo これどうしたいのだわ？カウントいるのかしら
まず、stringz, cafy, PostgreSQL の文字数のカウント方法が全部違う。
だから、同じ制限値を使い回すことに無理がある。
あまり正確でないカウンターをクライアントで出すことに意味があるのか。

syuilo:
あれ、stringzとPostgreSQLのカウント方法違うのか
PostgreSQLはUnicodeの文字数単位でカウントするから一緒かと思った

mei23:
stringzは絵文字とかの合字 (サロゲートペアではない) を1文字扱いにする
cafyはサロゲートペアを2文字にするはず

mei23:
APIの制限は、Note.textみたいにcafyのカウンタ使わないほうがいいのかも
https://github.com/misskey-dev/misskey/blob/cc756209764bef11709c71352b1402514174162a/src/server/api/endpoints/notes/create.ts#L51

src/services/drive/add-file.ts
```diff
@@ -389,7 +391,7 @@ export default async function(
 	file.userId = user ? user.id : null;
 	file.userHost = user ? user.host : null;
 	file.folderId = folder !== null ? folder.id : null;
-	file.comment = comment;
+	file.comment = truncate(comment, DB_MAX_IMAGE_COMMENT_LENGTH);
```
mei23: This is also used outside of ActivityPub, so it may be better to limit `image.name` below. 
https://github.com/misskey-dev/misskey/blob/cc756209764bef11709c71352b1402514174162a/src/remote/activitypub/models/image.ts#L31

Johann150: I think the current way is better, see the commit message:
> Truncating could also be done in `src/remote/activitypub/models/image.ts` or `src/services/drive/upload-from-url.ts` but both of those will use the code in `src/services/drive/add-file.ts` in the end. So this seems like the safest approach.

mei23: `src/services/drive/add-file.ts` also works for local uploads.
I don't think truncate will occur for local uploads.
I think it is unsafe to ignore unexpected patterns and continue.

Johann150: I see, local changes will now definitely cause a hard error if trying to insert into the database. This is a double check from the validator.

src/remote/activitypub/models/image.ts
```diff
@@ -28,6 +30,8 @@ export async function createImage(actor: IRemoteUser, value: any): Promise<Drive
 	const instance = await fetchMeta();
 	const cache = instance.cacheRemoteFiles;
 
+	image.name = truncate(image.name, DB_MAX_IMAGE_COMMENT_LENGTH);
```
mei23: If possible, I want you to copy it without changing the original.

mei23 (approved):
とりあえずはたぶん大丈夫

PR #7709, opened 2021-08-26
feat: Secure Mode and Private Mode

nullobsi:
# What

他のインスタンスからの投稿、リアクション、などがもっとコントロールしたり、Mastodonみたいな機能「セキュアモード」を追加しました。
セキュアモードで、ほかのインスタンスからのリクエストがHTTP Signatureを付ける必要にします。ブロックされたインスタンスで表示させません。
それとも、非公開モードも追加しました。非公開モードで、**世界にも**見えないようにします。allowedInstancesだけに連合、リアクション、返信したりできます。

# Why

小さなインスタンス管理者が嫌がらせなんてから守ることがときどき難しいですから、ホワイトリストで連合できると、もっと管理しやすいようにします。

# Additional info (optional)


mei23:
For verified AP responses, please set
```
ctx.set('Cache-Control', 'private, max-age=0, must-revalidate');
```

For anonymous responses, leave as below
```
ctx.set('Cache-Control', 'public, max-age=180');
```

nullobsi:
I did more testing on a test instance and everything seems to be fine. Secure mode works well in keeping posts reaching blocked instances, and I think limited federation keeps most info about a post hidden.

mei23:
Maybe the code is great though. I have to explain how the function should be, so please wait for a while.

packages/backend/src/remote/activitypub/check-fetch.ts
```diff
@@ -0,0 +1,78 @@
...
+
+		// それでもわからなければ終了
+		if (authUser == null) {
+			return 403;
+		}
+
+		// publicKey がなくても終了
+		if (authUser.key == null) {
+			return 403;
+		}
```
Johann150: I think it could be written like this to make it more concise?

```suggestion
		if (authUser?.key == null) {
			return 403;
		}
```

Johann150 (changes requested):
Please also check the files tab for linter annotations.

Johann150:
related to #2318?

nullobsi:
Yes, with Private Mode enabled, AP becomes disabled. But it doesn't let users browse the pages; you have to be logged in to view posts when Private Mode is enabled. Perhaps they should be two seperate options?
- Private Mode (as before)
- Limited Federation (allowlist of instances, posts are still viewable)
- Secure Mode (require signing of AP get)
- Normal (as Misskey default)

nullobsi:
Perhaps a UI like this:
```
連合の制限：
[パブリック]  [セキュア]  [許可されたインスタンス]  [非公開]
|-----O--------------------------------------------|
```
with a slider to select the federation limits.

packages/backend/src/server/web/views/channel.pug
```diff
@@ -1,18 +1,20 @@
 extends ./base
 
 block vars
-	- const title = channel.name;
+	- const title = privateMode ? '非公開インスタンス' : channel.name;
 	- const url = `${config.url}/channels/${channel.id}`;
 
 block title
 	= `${title} | ${instanceName}`
```
acid-chicken: Including in the title whether the instance is private or not will not make much sense to the end users, so instead the title itself should be only the `instanceName`.

nullobsi: That makes sense. I changed it to just be the instancename.

itzteajay-glitch:
Any luck on this getting deployed? I was disappointed to see I could not lock down my instance for use with just my family and friends.

nullobsi:
this is a 3yr old pull request and would have to be re-done for the new misskey version, and is probably useless. i don't develop Fedi software anymore

PR #7748, opened 2021-08-31
refactor: decode punycode domains for the UI

Johann150:
# What
The function `getAcct` is improved by a `getAcctUi` function. It returns a Unicode version of the domain name, instead of the punycode version. Also the `acct` function, which was only an alias for `getAcct` is refactored and removed. It was left over from the Vue3 migration.

There are a few places where `getAcct` is not replaced (yet):
- [ ] `src/client/components/post-form.vue` & `src/client/ui/chat/post-form.vue`
    uses account name in a post, will MFM decode IDNs correctly? if so `getAcct` should also be replaced with `getAcctUi` here because the user name is show to the user.
- [ ] `src/client/pages/my-antennas/editor.vue`
    will antennas correctly recognize IDNs?
- [ ] `src/server/api/limiter.ts`
    This could put IDNs in log files. Does the database, log and syslog protocol all support Unicode for this? And should IDNs be used here?

# Why
Punycode was not always decoded in the user interface. For example:
<details><summary>old screenshot</summary>

</details>

<details><summary>new screenshot</summary>

</details>

# Additional info
related: #5826


syuilo:
Thank you!
One concern is that when a user copies acct, the punycode will still be decoded when pasted, and will not be recognized correctly when mentioning, etc.

Johann150:
Yes I see. That is also why I did not change anything in `post-form.vue` yet, so autocompletion will currently still insert the punycoded name.

But as a user I would expect that an IDN would be correctly recognized if I type it in manually or copy it from somewhere else like Mastodon or a web page. So I think this is another issue in MFM, that we should solve too. Do you agree?

syuilo:
We do not want to interpret unencoded punycodes, because MFM should be parsed according to the specification. (As well as not interpreting URLs that are not percent encoded)

Perhaps we need to make it so that when a punycode-decoded acct is pasted into the submission form, it is automatically encoded. However, for users who are not familiar with punycode, they may misunderstand that it was not pasted correctly.
The extent to which Misskey will support punycode may need to be considered.
So, I'm going to put this PR on hold for now🙏

Johann150:
Sure 👍

I think I will open a separate issue for discussion?

Johann150:
some UI elements already decode punycode and some do not, making it even more confusing. The profile and the mentioned profile are on the same instance:
image

Johann150:
When https://github.com/misskey-dev/mfm.js/pull/86 gets merged, MFM will be able to parse IDN mentions correctly, which will check the first item off the list.

The implementation of `acct` has been changed so that part now has to be solved as https://github.com/misskey-dev/misskey.js/issues/32

PR #7777, opened 2021-09-06
add resolver check for blocked instance

### LINKED ISSUE #7776: resolver ignores blocked instances
## 💡 Summary

When using the resolver, the blocked instance list is ignored. This leads to the database becoming populated with potentially unwanted emojis, files, etc from blocked remote instances. 

## 🙂 Expected Behavior

Ideally, no requests should be made to blocked instances. 

## ☹️ Actual Behavior

Blocking an instance can still lead to logs like
```
register emoji host = blocked.instance
Creating the Person: https://blocked.instance/users/one
```
etc

## 📝 Steps to Reproduce

1. Block an instance
2. Attempt to resolve a reply to a blocked instance, or otherwise create a call to the resolver.
3. Instance data is still downloaded into the database

nullobsi:
# What
close #7776 by adding check for blocked instances

# Why

Ideally, no requests should be made to blocked instances

# Additional info (optional)

### DIFF
--- a/src/remote/activitypub/models/note.ts
+++ b/src/remote/activitypub/models/note.ts
@@ -301,6 +301,10 @@ export async function extractEmojis(tags: IObject | IObject[], host: string): Pr
 
 	if (!tags) return [];
 
+	// ブロックしてたら中断
+	const meta = await fetchMeta();
+	if (meta.blockedHosts.includes(host)) throw new Error('blocked instance emoji');
+
 	const eomjiTags = toArray(tags).filter(isEmoji);
 
 	return await Promise.all(eomjiTags.map(async tag => {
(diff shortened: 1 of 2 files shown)


src/remote/activitypub/models/note.ts
```diff
@@ -301,6 +301,10 @@ export async function extractEmojis(tags: IObject | IObject[], host: string): Pr
 
 	if (!tags) return [];
 
+	// ブロックしてたら中断
+	const meta = await fetchMeta();
+	if (meta.blockedHosts.includes(host)) throw new Error('blocked instance emoji');
+
```
mei23: You can't attach an emoji from another instance, so it may not be necessary if it's blocked by resolver.

mei23 (approved):

syuilo:
🙏

PR #7808, opened 2021-09-18
fix(client): fix #6811 - Fix image height when uses CW to hide it

### LINKED ISSUE #6811: CWで画像が隠されたとき、画像の高さがおかしいことになる
## 💡 Summary
CWで画像が隠されたとき、画像の高さがおかしいことになる

https://github.com/syuilo/misskey/commit/21b6e23e986b5f43de89927d675f310055d83d97 でこれを不完全ながら防いでいたんだけど、メモリリークの問題があるのかぁ…

## 🙂 Expected Behavior
image

## ☹️ Actual Behavior
image

## 📝 Steps to Reproduce

1. CWで画像が添付されている投稿を見つける
2. CWを開く

## 📌 Environment
Firefox 83.0b8 (64 ビット); Misskey v12.53.0

null31:
# What

Replace `const parent = this.$parent.$el;` by `const parent = this.$parent.getElementsByClassName('main');`
Will use div (class main) as reference to calculate the image height instead of whole note box (this.$parent.$el).

# Why

Try to fix UI glitch that affects default, deck and pope when uses CW (#6811).
So, this fix an annoying UI glitch when CW is used to hide images, because those images stay behind of polls and footer bar (overlapping the image).

# Additional info (optional)

The easiest way to reproduce the problem are using deck UI and resizing the timeline box.

- Before apply this:

before

- After apply this:

after

### DIFF
--- a/src/client/components/media-list.vue
+++ b/src/client/components/media-list.vue
@@ -64,7 +64,7 @@ export default defineComponent({
 
 				if (this.$refs.gridOuter) {
 					let height = 287;
-					const parent = this.$parent.getElementsByClassName('main');
+					const parent = this.$parent?.$el.getElementsByClassName('main')[0];
 
 					if (this.$refs.gridOuter.clientHeight) {
 						height = this.$refs.gridOuter.clientHeight;
--- a/src/client/components/media-list.vue
+++ b/src/client/components/media-list.vue
@@ -64,7 +64,7 @@ export default defineComponent({
 
 				if (this.$refs.gridOuter) {
 					let height = 287;
-					const parent = this.$parent?.$el.getElementsByClassName('main')[0];
+					const parent = this.$parent.getElementsByClassName('main')[0];
 
 					if (this.$refs.gridOuter.clientHeight) {
 						height = this.$refs.gridOuter.clientHeight;
--- a/src/client/components/media-list.vue
+++ b/src/client/components/media-list.vue
@@ -64,7 +64,7 @@ export default defineComponent({
 
 				if (this.$refs.gridOuter) {
 					let height = 287;
-					const parent = this.$parent.getElementsByClassName('main')[0];
+					const parent = this.$parent?.$el.getElementsByClassName('main')[0];
 
 					if (this.$refs.gridOuter.clientHeight) {
 						height = this.$refs.gridOuter.clientHeight;
--- a/src/client/components/media-list.vue
+++ b/src/client/components/media-list.vue
@@ -64,7 +64,7 @@ export default defineComponent({
 
 				if (this.$refs.gridOuter) {
 					let height = 287;
-					const parent = this.$parent.$el;
+					const parent = this.$parent.getElementsByClassName('main');
 
 					if (this.$refs.gridOuter.clientHeight) {
 						height = this.$refs.gridOuter.clientHeight;
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 - UIの改善
 
 ### Bugfixes
+- fix image height when uses CW to hide it (#6811)
 - アカウントデータのエクスポート/インポート処理ができない問題を修正
 - popupで設定ページを表示すると、アカウントの削除ページにアクセスすることができない問題を修正
 


tamaina:
これはエラーで高さ変更処理を中断させた結果、高さが更新されていないだけなのでは？

たぶんiOSに適用するとバグると思う

tamaina:
（iOSを優先させたがために他のブラウザでバグが出ているのはどうかと思っているけど）

syuilo:
> （iOSを優先させたがために他のブラウザでバグが出ているのはどうかと思っているけど）

iOSを優先しないとなると、画像を**全く**見れなくなるので、それは回避しなければならない

src/client/components/media-list.vue
```diff
@@ -64,7 +64,7 @@ export default defineComponent({
 
 				if (this.$refs.gridOuter) {
 					let height = 287;
-					const parent = this.$parent.$el;
+					const parent = this.$parent.getElementsByClassName('main');
```
tamaina:
```suggestion
					const parent = this.$parent?.$el.getElementsByClassName('main')[0];
```

tamaina:
I doubt that there will always be a main class, but it seems better than no change.

null31:
> これはエラーで高さ変更処理を中断させた結果、高さが更新されていないだけなのでは？
> たぶんiOSに適用するとバグると思う

Maybe. From what I could see is that `article class=article` is used to calculate the height of images, but when you use CW, enters here `else if (parent) { height = parent.getBoundingClientRect().width * 9 / 16; }`.

It's don't seems to be caused by a error in the processing, it's just uses an element that doesn't fit when you expand CW, that is why you should use the major blue square instead the green one.

Screenshot_2021-09-19_10-56-43

>iOSを優先させたがために他のブラウザでバグが出ているのはどうかと思っているけど

There is.
I tested in Firefox v91, Chromium v91 and Bromite v91 (Android), they all have the same behavior.
Also, I don't have a iOS to test it.

src/client/components/media-list.vue
null31: Just tested this change. Works only when you don't change the timeline box width.

null31: Well... I messed a bit with the above "commit suggestion" button.
But what I want to suggest are using then `this.$parent.getElementsByClassName('main')[0];`, works fine how you expect when changes width.

null31:
Anyway, I don't know the purpose of `$el` in that context, but seems to be the reason why images height get messed.

null31:
> Anyway, I don't know the purpose of `$el` in that context, but seems to be the reason why images height get messed.

- `const parent = this.$parent?.$el.getElementsByClassName('main')[0];`
Screenshot_2021-09-19_11-42-52

- `const parent = this.$parent.getElementsByClassName('main')[0];`
Screenshot_2021-09-19_11-36-25

tamaina:
> `article class=article` is used to calculate the height of images

This is something I hadn't noticed, and I noticed it during when I was reviewing.

You should learn about some of the APIs:

- $parent https://v3.vuejs.org/api/instance-properties.html#parent
- $el https://v3.vuejs.org/api/instance-properties.html#el
- getElementsByClassName() https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByClassName

$parent returns `Component instance`, so we can't use getElementsByClassName().

It is the reason for my remark:

> これはエラーで高さ変更処理を中断させた結果、高さが更新されていないだけなのでは？
> (Isn't this just a result of the height change process (size()) being interrupted by an error, and the height not being updated?)

We should use $el so that we use getElementsByClassName().

`const parent = this.$parent.getElementsByClassName('main')[0];` muse works well on Chromium and Firefox because the height is determined by CSS.

tamaina:
Anyway, it is never a good idea to merge a PR that is sure to throw errors.

`const parent = this.$parent?.$el.getElementsByClassName('main')[0];` works well in my Firefox and Edge.

image

null31:
Thanks for explaining about vue stuff.

The last code fix the problem where timeline don't change width value. Although UI deck and desktop doesn't work because they allow you change width later.

So, what you said is that height don't change because doesn't uses CSS (calculate and then never update the value)?

tamaina:
I didn't explain it well.

"CSS" I said is the padding-top set ( https://github.com/misskey-dev/misskey/blob/develop/src/client/components/media-list.vue#L92-L96 ) and you're not aware of this.

In windows, pope UI and chat UI where the resize event is not fired even though the width of the parent element has changed, padding-top set is likely to be a bad thing.

tamaina:
たしかに一部でバグる

image

tamaina:
iOS15正式リリースされたし、高さ固定処理自体を削除する方針( 558d288 の復活)でいいでしょうか？

syuilo:
> iOS15正式リリースされたし、高さ固定処理自体を削除する方針( [558d288](https://github.com/misskey-dev/misskey/commit/558d288e7bc129fa92ee7dd1dc08d5a9f3615f8f) の復活)でいいでしょうか？

どれくらいのユーザーがアプデするかに依りそうだけど、遅かれ(最長で半年くらい？)早かれそうするのは確実

tamaina:
雰囲気的にiOS 15未満は切っちゃっていいかも

https://p1.a9z.dev/notes/8s8rf2ql2y

syuilo:
KIRUZO

PR #7871, opened 2021-10-15
chore: remove unnecessary imports

Johann150:
# What

Remove unused imports of `userPage` and `acct`.

# Why

The imports are unnecessary and clutter up the code.


syuilo:
🙏

PR #7875, opened 2021-10-16
use high resolution asset for splash screen

### LINKED ISSUE #7873: loading screen uses low resolution favicon
The loading screen is shown when a user is logged in, but before the main web app is shown. It can also be seen when the page is reloaded.
Currently, the loading screen uses `favicon.ico`. This file is usually only small (recommended 32px or 64px square image). Instead, one of the more high resolution pictures could be used for the loading screen, which are already available in the `assets` directory. These files are larger and should look better at the size they are shown in the loading screen.

Johann150:
# What
Use the 512px size icon asset for the splash screen instead of the favicon.

# Why
fix #7873

### DIFF
--- a/packages/backend/src/server/web/views/base.pug
+++ b/packages/backend/src/server/web/views/base.pug
@@ -56,5 +56,5 @@ html
 			br
 			| Please turn on your JavaScript
 		div#splash
-			img(src='/favicon.ico')
+			img(src='/static-assets/icons/512.png')
 		block content


Johann150:
Maybe the 512x512 picture is too big to load quickly on some connections. But the image is displayed at quite a large size, so a high resolution picture would be good in those circumstances.
The optimal solution would probably be using `@media` queries to load differen images depending on screen size.

syuilo:
Thanks!
In the future, it might be nice to be able to display a logo image of an instance.

PR #7927, opened 2021-10-28
fix: stop context menu handling for videos

### LINKED ISSUE #7926: context menu triggers on video
## 💡 Summary

When right clicking on a video, the Misskey context menu is triggered. The behaviour is different from right clicking on images.

Additionally, right clicking on a video multiple times shows the context menu multiple times, which makes it seem like this is not intended.

## 🙂 Expected Behavior

Native browser context menu opens. (Same behaviour as images)

## ☹️ Actual Behavior

Misskey custom context menu opens.

## 📝 Steps to Reproduce

1. Go to post with attached video.
2. Right click on video.
3. (optional) right click on video again.

## 📌 Environment

Misskey 12.94.1

Johann150:
# What
The context menu handler is set to stop. The native context menu will be displayed instead of the Misskey custom context menu.

# Why
fix #7926

### DIFF
--- a/src/client/components/media-video.vue
+++ b/src/client/components/media-video.vue
@@ -11,6 +11,7 @@
 		:title="video.name"
 		preload="none"
 		controls
+		@contextmenu.stop
 	>
 		<source 
 			:src="video.url" 


syuilo:
🙏

PR #7942, opened 2021-10-31
fix(client): reaction viewer layout

Johann150:
# What
The design of the reaction viewer is changed so that the situation like in the picture can not occur:
image
Note how the profile picture of the 2nd person is on the 1st line, but the name is on the 2nd line.

Instead the design now looks like this: (sorry the background was different, the square/round profile pictures are different because i took the picture on another misskey instance)
image
The profile picture is always above the name, and different users are separated vertically.

# Why
The profile picture and name should be grouped together as they belong, and it should be clear which picture belongs to which name.

# Additional info
What seemed to be the error in the old code is that `display: flex` is not inherited by `.users`.

The intendation of one line was also fixed.


syuilo:
👍

PR #7950, opened 2021-11-02
fix: limit menu list size

### LINKED ISSUE #7946: Missing scroll functionality on Place furniture dropdown.
## 💡 Summary

When selecting a new item to place in the Misskey room, the dropdown menu stretches the whole height of the screen with no way to scroll when on Desktop.

## 🙂 Expected Behavior

I'd expect the select to have a `max-height` with vertical scrolling so that options further down can be selected.

## 📝 Steps to Reproduce

1. Go to your Misskey Room
2. Click "Place furniture"

## 📌 Environment

Windows 10, Google Chrome (Not sure if it's browser or OS specific)

Johann150:
# What
Limit the size of a selection menu. This can be an issue for example when adding an object to the room.
Also use `overflow: auto` to make sure a scroll bar appears if necessary.

# Why
fix #7946

### DIFF
--- a/src/client/components/ui/menu.vue
+++ b/src/client/components/ui/menu.vue
@@ -147,6 +147,8 @@ export default defineComponent({
 .rrevdjwt {
 	padding: 8px 0;
 	min-width: 200px;
+	max-height: 90vh;
+	overflow: auto;
 
 	&.center {
 		> .item {


syuilo:
Thanks!

PR #7975, opened 2021-11-11
fix: mention local users in replies

### LINKED ISSUE #7870: Mention local users in replies
## Summary

Replying to local accounts uses just a "reply", and not a mention. This causes issues with notifications, leaving replies unseen.

Instead, replying to local users should also include a mention so further replies will include the user.

nullobsi:
# What

Close #7870

When replying to local users, mentions will be used.
```
user1: Hello!
-->user2: How are you? (notification)
-->-->user2: I hope you are doing good. (no mention, so no notification)
```
new:
```
user1: Hello!
-->user2: @user1 How are you? (notification)
-->-->user2: @user1 I hope you are doing good. (mention, so there is notification)
```

# Why

No mentions causes issues when replying in threads; the users may not get notifications because they do not get mentioned.
Also, it may confuse remote users replying, and the local users will not get notified.
# Additional info (optional)

### DIFF
--- a/packages/client/src/components/post-form.vue
+++ b/packages/client/src/components/post-form.vue
@@ -279,8 +279,8 @@ export default defineComponent({
 			this.text += ' ';
 		}
 
-		if (this.reply && this.reply.user.host != null) {
-			this.text = `@${this.reply.user.username}@${toASCII(this.reply.user.host)} `;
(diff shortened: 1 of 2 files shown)


Johann150:
This is not the UI component used for the normal UI. You have to change it here too:
https://github.com/misskey-dev/misskey/blob/ea97d6a028b817923dd650754456c9905f0f89ea/packages/client/src/components/post-form.vue#L282-L284

syuilo:
👍

PR #7979, opened 2021-11-13
limit depth of reply posts

### LINKED ISSUE #6343: New threaded replies break on long threads
## 💡 Summary

I love the new threaded replies, but it appears that they don't stop even when outside the viewport. An example would be https://catgirl.life/notes/83msoy88v8

## 🙂 Expected Behavior

Stop threading when thread starts leaving the view port with a indication that the thread continues. Maybe allow users to change size of the viewport. 

Screenshot 2020-05-10 13 19 22

## ☹️ Actual Behavior

image

## 📝 Steps to Reproduce

1.Make a really long thread

## 📌 Environment

Misskey v12.37.0

Johann150:
# What
Replies to a note are not shown after a specific depth. I arbitrarily chose 20. After that only a link is displayed. The link goes to the note above it so all replies can be viewed from there and only one link is needed (instead of making multiple links to each separate reply).

This is in addition to the `max-width_450px` class etc. being applied.

# Why
fix #6343

This makes it easier to view different replies if there are long threads. It also reduces network load for long threads because not everything is immediately fetched from the server.

# screenshot
image


syuilo:
👍

PR #7980, opened 2021-11-13
feat: Undo Accept

nullobsi:
# What

Add a "Break Follow" feature.
image

It will remove the other person from following you.

# Why

For people with follower-approval accounts, they may want to remove a person from access. Blocking and then unblocking can be annoying, or send the wrong message. Therefore, "breaking follow" is a good solution.

# Additional info (optional)

- [ ] needs tests


Johann150:
There has been some discussion on whether to introduce something like this in #7899, although for all users, not only those with manual approved follows.

nullobsi:
This works for accounts with and without approve followers on

nullobsi:
Ah, it seems this is already a thing but under "Reject Follow" for Mastodon. So I'll update the AP implementaton to use that for better compatibility

syuilo:
🆒

PR #8001, opened 2021-11-18
enhance: Forward report

### LINKED ISSUE #8000: forwarding reports
## Summary

It should be possible to forward reports to the other instance if the post was made on another instance.

This is important so the responsible moderator team is also notified of Spam or bad behaviour because they might have to intervene and suspend and/or delete the account in severe cases.

Johann150:
# What
Add the ability for users to choose to forward a report to the other instance for remote users.

# Why
fix #8000 

# Additional info (optional)
The API is extended in a backward-compatible manner.

In the future, the API should probably be further expanded to allow to pass in URLs of notes, so the note link does not have to be in the report text.


Johann150:
The locale strings are still missing because I don't know Japanese well enough. In English it would be:
```
forwardReport: "Forward the report to the other instance."
forwardReportIsAnonymous: "The other instance will see a system account instead of your name."
```
What should it be in Japanese?

syuilo:
```
forwardReport: "リモートインスタンスに通報を転送する"
forwardReportIsAnonymous: "リモートインスタンスからはあなたの情報は見れず、匿名のシステムアカウントとして表示されます。"
```

mei23:
リモートに転送するか選択するスイッチは、ユーザーの通報画面ではなく管理画面の通報一覧にあるべきと思ってる。

Johann150:
I put it on the users screen because that is how [Pleroma](https://api.pleroma.social/#operation/ReportController.create) and [Mastodon](https://docs.joinmastodon.org/methods/accounts/reports/) do it.

mei23:
I know that. I think Mastodon is strange.

Johann150:
Trying to find out why it is implemented this way in Pleroma and Mastodon, I found that Pleroma copied it from Mastodon. And Mastodon has this pull request: https://github.com/mastodon/mastodon/pull/6570

Looking further I found this comment that actually suggested that only the admin can forward a report: <https://github.com/mastodon/mastodon/issues/2176#issuecomment-310358890>. So I agree it would probably be a good idea to instead add the ability to the reports panel so the admin can select it. It should be selectable for each report separately.

I am not sure how to best implement this though. I could imagine
1. A button that immediately forwards the report. We would have to save whether the report has been forwarded already so that it can not be forwarded multiple times.
2. A button that forwards and closes the report.
3. A switch that forwards the report when the moderator clicks on "mark the issue as resolved". Then we would not have to save whether the report was forwarded already.

Johann150:
I chose option 3, but still saving whether the report was forwarded so that this can be seen for resolved reports.

A moderator can change the switch if they want to forward the report (off by default), and then click on the mark as resolved button.
<details>
<summary>screenshot of a new report</summary>

image
</details>

An admin on the other instance will see something like the next screenshot. Note that the toggle switch is off and disabled because the report can not be forwarded again.
<details>
<summary>screenshot of a forwarded report on the other instance</summary>

image
</details>

I changed the UI of resolved reports slightly. I use the existing `$ts.moderator` i18n string for asignee and added a `$ts.reporter` i18n string. I moved the assignee information to the same position as the reporter information to make it easier to see at a glance.

The switch to forward a report will still be visible for resolved reports, but of course disabled because it can not be changed after resolving the report. It will be on or off to show whether the report was forwarded. In this screenshot it was forwarded so it is on.
<details>
<summary>screenshot of a forwarded report on the instance the report was created on</summary>

image
</details>

For local users, forwarding the report will of course not be available. It will look like this (this time in Japanese):

<details>
<summary>screenshot of a report concerning a local user</summary>

image
</details>

Johann150:
Rebased to resolve the merge conflict. I don't think the test failure is caused by this code, the test passed earlier.

syuilo (approved):
YOSASOU

Johann150:
Installed this on my own instance yesterday and tested with someone else that it is working with Mastodon. 🎉

Although they noted that the report looks like it is regarding the whole account and not a specific note, which may cause issues with the way moderation tools work in Mastodon: Moderators would only be able to do things with the account and not with the specific post. I am pretty sure this is due to the value of `object` in the `Flag` activity that is sent (2nd argument to `renderFlag`). This should also include the URL of the specific note, but that would require changing of `@/components/abuse-report-window.vue` to pass the URL of the note along.

But since reporting works in general I think this is something that can be fixed later. We can create a separate issue when this is merged.

packages/backend/migration/1637320813000-forwarded-report.js
```diff
@@ -0,0 +1,13 @@
+const { QueryRunner } = require('typeorm');
+
+module.exports = class forwardedReport1637320813000 {
+	name = 'forwardedReport1637320813000';
```
Johann150, line 4: Will this migration apply correctly if there is another migration with a higher number? `1639325650583-chart-v3.js` was merged yesterday

Johann150:
It seems this is currently broken 😞

Johann150:
With the latest commit, it will also work correctly with Pleroma. 🥳

syuilo:
🆒

PR #8016, opened 2021-11-23
Implemented a rotate MFM

EpicnessTwo:
# What

Implemented a rotate MFM tag

# Why

Missing feature of MFM that would be really helpful! It also helps lower the amount of variations of emoji required on an instance that have different rotations.

# Additional info (optional)


syuilo:
👌

PR #8030, opened 2021-11-30
fix: mentions to remote local user

nullobsi:
# What

when replying to a mention to another instance's local user, the host part is added.

20211130_09h33m26s_grim

# Why

20211130_09h29m54s_grim

before the fix, replying to another user's post with a local mention breaks and tries to mention a potentially nonexistent local user. this is jarring and not proper behavior
# Additional info (optional)


syuilo:
👍👍👍

PR #8036, opened 2021-12-04
fix: integration links

outloudvi:
# What

Pick the `meta.uri` out and put it in the right places.

# Why

This will fix the callback URL display in:
* `/admin/integrations/discord`
* `/admin/integrations/github`
* `/admin/integrations/twitter`

Instead of:

Callback URL: undefined/api/gh/cb


syuilo:
🙏

PR #8062, opened 2021-12-16
feat: remove file count limit for a note

### LINKED ISSUE #8054: Allow changing the number of files in a note
## Summary

(Sorry for not being able to search if there are Japanese duplicates.)

Currently there is a fixed number of maximum 4 files in a note:

https://github.com/misskey-dev/misskey/blob/d95fafb5b35729638f0c5f9f26dd2be78c822170/packages/backend/src/server/api/endpoints/notes/create.ts#L81

https://github.com/misskey-dev/misskey/blob/d95fafb5b35729638f0c5f9f26dd2be78c822170/packages/client/src/components/post-form-attaches.vue#L13

However, 4 is a pretty small number. Probably the number can be increased or [even be configurable](https://github.com/outloudvi/misskey/tree/feat/maxFileNumber).

outloudvi:
# What & Why
Resolves #8054.

# Additional info (optional)
This also changed the class name of `.remain` for semantic reasons, please check if it creates breakages elsewhere.

Also, We probably also need some UX changes to optimize for too many images in a note.

### DIFF
--- a/packages/backend/src/server/api/endpoints/notes/create.ts
+++ b/packages/backend/src/server/api/endpoints/notes/create.ts
@@ -78,11 +78,11 @@ export const meta = {
 		},
 
 		fileIds: {
-			validator: $.optional.arr($.type(ID)).unique().range(1, 4),
+			validator: $.optional.arr($.type(ID)).unique(),
 		},
 
 		mediaIds: {
-			validator: $.optional.arr($.type(ID)).unique().range(1, 4),
+			validator: $.optional.arr($.type(ID)).unique(),
 			deprecated: true,
 		},
 
--- a/packages/client/src/components/post-form-attaches.vue
+++ b/packages/client/src/components/post-form-attaches.vue
@@ -10,7 +10,7 @@
 			</div>
 		</template>
 	</XDraggable>
-	<p class="remain">{{ 4 - files.length }}/4</p>
+	<p class="count">{{ files.length }}</p>
 </div>
 </template>
 
@@ -41,7 +41,6 @@ export default defineComponent({
 	data() {
 		return {
 			menu: null as Promise<null> | null,
-
 		};
 	},
 
@@ -179,7 +178,7 @@ export default defineComponent({
 		}
 	}
 
-	> .remain {
+	> .count {
 		display: block;
 		position: absolute;
 		top: 8px;


packages/backend/src/server/api/endpoints/notes/create.ts
```diff
@@ -78,11 +78,11 @@ export const meta = {
 		},
 
 		fileIds: {
-			validator: $.optional.arr($.type(ID)).unique().range(1, 4),
+			validator: $.optional.arr($.type(ID)).unique(),
 		},
 
 		mediaIds: {
-			validator: $.optional.arr($.type(ID)).unique().range(1, 4),
+			validator: $.optional.arr($.type(ID)).unique(),
```
Johann150, line 85: Maybe we should validate that the array has at least one element if it is present, but from what I can see it should also work without that validation, since it will be replaced by an empty array if it is not present: https://github.com/outloudvi/misskey/blob/32ec42a89f0ae6d40e047f84e1bef2c0f6820551/packages/backend/src/server/api/endpoints/notes/create.ts#L185-L194

packages/client/src/components/post-form-attaches.vue
```diff
@@ -179,7 +178,7 @@ export default defineComponent({
 		}
 	}
 
-	> .remain {
+	> .count {
```
Johann150, line 181: This shouldn't cause a problem because the style is scoped.
https://github.com/outloudvi/misskey/blob/32ec42a89f0ae6d40e047f84e1bef2c0f6820551/packages/client/src/components/post-form-attaches.vue#L134

Johann150 (approved):

tamaina:
サーバーサイドでもバリデーションがある

Johann150:
`packages/backend/src/server/api/endpoints/notes/create.ts`はサーバーサイドですね?

tamaina:
> packages/backend/src/server/api/endpoints/notes/create.tsはサーバーサイドですね?

yes

Johann150:
> ノートの文字数には制限があるなら、ファイルの数に制限がある方が自然そう
>（そもそもノートの文字数制限が要るかどうか怪しいところではあるけど

I think the character limit is to discourage giant notes that take up too much disk space. But for files this is already solved with the Misskey drive quota.

rinsuki:
I think 4 files limit are too small, but I also think we should have a useful limit (like 16 files), for prevent some user attaches TONS of files to one post and makes server slow/down

Johann150:
> and makes server slow down

I don't understand how attaching more files slows down the server. How much the server has to work to create the note does not depend on the number of files, but on the number of federating instances.

Or do you mean the server slows down because many files will be requested? In that case, how would this be different from someone creating many notes with a few attachments?

syuilo:
We chose to increase the limit to 16, rather than eliminate it completely.
Thanks anyway 🙏🏻

Johann150:
It would have been nice if somebody explained this decision more in detail why it was not removed completely. I think with a UI redesign for displaying images this could have been a good idea.

Also, how does this relate to the linked issue #8054? I think since other Fediverse software can still send more notes than Misskey, a UI change would still be good.

PR #7943, opened 2021-10-31
fix(mfm): correct code blocks conversion

Johann150:
# What
Another newline is added in front of code blocks so their MFM syntax can be recognized correctly.

# Why
fix #7919

Since the conversion for other block elements like especially `<p>` puts the newline in front of the element, this also has to put the newline in front so the MFM syntax gets recognized correctly.

PR #8069, opened 2021-12-19
fix: HTML->MFM blockquote

Johann150:
# What
Fix the conversion of blockquotes from HTML to MFM, because a quote in MFM has to start at the first position.

# Why
```
<blockquote><p>quote</p></blockquote>
<p>text</p>
<blockquote><p>another quote</p></blockquote>
<p>more text</p>
```

got displayed as
```
> quote
text> another quote
more text
```


Johann150:
this is #7943 but for blockquotes

syuilo:
🙏

PR #8112, opened 2022-01-02
enhance: disable animations on more transitions

Johann150:
# What
Enabling the "Reduce UI animation" setting leaves many animations enabled. With this change, all transition animations will be disabled when the "Reduce UI animation" setting is enabled.

# Why
The UI animations can be resource intensive and drain a lot of charge, especially on mobile devices (phones & laptops).


syuilo:
🙏

PR #8117, opened 2022-01-05
fix: code url in documentation

Johann150:
# What
Change the base URL used to generate the links to source code in the documentation.

# Why
It seems this was not changed while refactoring the modules apart.


syuilo:
🙏🏻

PR #8131, opened 2022-01-09
fix: image description does not save locally

nullobsi:
# What
Captioning an image now saves your caption locally so it can be edited.

# Why
Before, when saving a caption, the server would save it. But it would not save in your immediate draft, which could cause confusion.

# Additional info (optional)


syuilo:
🙏

PR #8186, opened 2022-01-25
enhance: Improve poll-editor UI + composition port

Skehmatics:
# What
Adds some (imo, much needed) whitespace to the poll creation UI, and takes better advantage of screen width when available by making expiration options one line. In addition, this ports the component to the composition API (setup sugar)

image
image

# Why
The current design feels a bit cramped, especially on mobile, and has some alignment issues.

# Additional info (optional)
Feel free to cherry-pick just the composition API port if you don't like the design (last 2 commits). praise syulio~


syuilo:
Thank you!
The due date specification does not seem to be reflected in the submission.

syuilo:
> The due date specification does not seem to be reflected in the submission.

fixed

syuilo:
🙏

PR #8230, opened 2022-01-30
feat: Federate clips

nullobsi:
# What
Public clips will now be exposed via AP OrderedCollections under /users/:id/clips and /clips/:id.
Misskey will also now accept Add, Remove, and Update activities for clips.
# Why
Many features in Misskey do not federate. To be a good member of the Fediverse, many of these features should be federating. It makes for a better experience.

# Additional info (optional)
The pull request is a bit large, but I have done local testing and it does work for the most part. I don't know how to write tests.

### DIFF
--- a/packages/backend/src/remote/activitypub/kernel/delete/clip.ts
+++ b/packages/backend/src/remote/activitypub/kernel/delete/clip.ts
@@ -1,0 +1,35 @@
+import { IRemoteUser } from '@/models/entities/user';
+import { apLogger } from '../../logger';
+import DbResolver from '../../db-resolver';
+import { getApLock } from '@/misc/app-lock';
+import { Clips } from '@/models';
+
+const logger = apLogger;
+
+export default async function(actor: IRemoteUser, uri: string): Promise<string> {
+	logger.info(`Deleting the Clip: ${uri}`);
+
+	const unlock = await getApLock(uri);
+
+	try {
+		const dbResolver = new DbResolver();
+		const clip = await dbResolver.getClipFromApId(uri);
+
+		if (clip == null) {
+			return 'clip not found';
+		}
+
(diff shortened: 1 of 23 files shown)


ThatOneCalculator (approved):
Locally tested as well, works as expected

mei23:
@syuilo え？いるこれ？

mei23:
入れるかとか仕様とか先に議論してくれないとキツすぎるわ

packages/backend/src/remote/activitypub/kernel/delete/clip.ts
```diff
@@ -0,0 +1,35 @@
...
+	try {
+		const dbResolver = new DbResolver();
+		const clip = await dbResolver.getClipFromApId(uri);
+
+		if (clip == null) {
+			return 'clip not found';
+		}
+
+		if (clip.userId !== actor.id) {
+			return 'Clipを削除しようとしているユーザーは投稿の作成者ではありません';
```
Johann150, line 23: If the other error messages are in English wouldn't it make sense to have this one in English as well?

mei23:
Thank you for your contribution.

First, decide whether to implement it in Issue.
If to implement it, please decide what to implement in Issue. 

In my opinion, I don't think this should be an important federation,
Even when federation, Collection and Add / Remote seem to be enough.

mei23:
いい加減こういうのはレビュー投げてくる前にちゃんと突っ返して 💢

syuilo:
> いい加減こういうのはレビュー投げてくる前にちゃんと突っ返して 💢

これから気を付けます🙇‍♂️

syuilo:
必要性などがまだ決まってない段階なのでCloseします🙏

PR #8233, opened 2022-01-30
fix: ensure that specified users does not get duplicates

nullobsi:
# What
Creating DMs no longer makes duplicat tags.

# Why
Causes an API error and gets longer the more replies you do

# Additional info (optional)


packages/client/src/components/post-form.vue
```diff
@@ -397,9 +400,15 @@ function setVisibility() {
 	}, 'closed');
 }
 
+function pushVisibleUser(user) {
+	if (!visibleUsers.some(u => u.username == user.username && u.host == user.host)) {
```
syuilo:
```suggestion
	if (!visibleUsers.some(u => u.username === user.username && u.host === user.host)) {
```

syuilo:
👍

PR #8254, opened 2022-02-04
fix: regular expressions in word mutes

### LINKED ISSUE #8245: Timeline loads old posts but not new posts, notifications work
## 💡 Summary
When logged into my account, older posts will load, but newer posts from people I follow will not load.
This happens regardless of browser or device, but other accounts load fine in the same browser and device.

Version 12.102.1 would load neither old or new posts; version 12.103.1 loads old posts, but not new.

## 🙂 Expected Behavior
As people I follow make posts, they load into my timeline within a reasonable time frame.

## ☹️ Actual Behavior
Notification sound for new posts goes off, but no post loads.
[Console log shows the same error over and over](https://github.com/misskey-dev/misskey/files/7988521/msky_log.txt).

## 📝 Steps to Reproduce
1. Log in
2. Wait for new posts from people you follow

## 📌 Environment
Any device and web broswer, so long as you sign in with this specific account.

Johann150:
# What
- Handle `SyntaxException`s from faulty `RegExp`s in word mutes.
- Prevent such exceptions from occurring in the first place by
    - checking that regular expressions in word mutes have valid syntax in the frontend and
    - checking that regular expressions in hard word mutes have valid syntax in the backend (endpoint `api/i/update`).
- Clean up empty lines etc. when setting word mutes instead of when checking each note. This should improve perfomance.
- Correctly handle regular expressions containing spaces: Before they would be split at spaces like other lines, causing them not to be recognized.

# Why
Handling exceptions fixes #8245.
Preventing such exceptions improves the user experience.

# Additional info
The storage format of muted words will be changed to `Array<string | string[]>`, where an array element of type `string` represents a regular expression, and `string[]` represents a normal word mute.

Hard word mutes can be migrated, but soft word mutes may stop working correctly if they contain `RegExp`s or empty lines. This can be fixed by going to soft word mute settings and saving once. This will update the structure to the new format.

## To do

- [x] Fix behaviour of frontend.
- [x] Fix behaviour of backend
    - [x] migration for hard word mutes.
- [x] New i18n strings need to be added. English version below, Japanese version welcome. 🙏
```yaml
regexpError: "Regular Expression error"
regexpErrorDescription: "Error in the regular expression on line {line} in your {tab} word mutes:"
```


robflop:
Suggestions for japanese strings:

```yaml
regexpError: "正規表現エラー"
regexErrorDescription: "{tab}ワードミュートの{line}列の正規表現にエラーが発生しました:"
```

Not completely sure what the ``{tab}`` bit refers to, so this might be wrong.

Johann150:
> Not completely sure what the `{tab}` bit refers to, so this might be wrong.

It will be either `_wordMute.soft` or `_wordMute.hard`, so they know which tab the line number is from.

robflop:
Ah, thanks. In this case, the string should be suitable.

locales/ja-JP.yml
```diff
@@ -593,6 +593,8 @@ smtpSecure: "SMTP 接続に暗黙的なSSL/TLSを使用する"
 smtpSecureInfo: "STARTTLS使用時はオフにします。"
 testEmail: "配信テスト"
 wordMute: "ワードミュート"
+regexpError: "正規表現エラー"
+regexErrorDescription: "{tab}ワードミュートの{line}列の正規表現にエラーが発生しました:"
```
rinsuki:
```suggestion
regexpErrorDescription: "{tab}ワードミュートの{line}行目の正規表現にエラーが発生しました:"
```

packages/client/src/pages/settings/word-mute.vue
```diff
@@ -81,18 +81,57 @@ export default defineComponent({
...
+			};
+
+			let softMutes = parseMutes(this.softMutedWords, this.$ts._wordMute.soft);
+			let hardMutes = parseMutes(this.hardMutedWords, this.$ts._wordMute.hard);
+
 			await os.api('i/update', {
-				mutedWords: this.hardMutedWords.trim().split('\n').map(x => x.trim().split(' ')),
+				mutedWords: hardMutes,
 			});
+
```
rinsuki: It seems you forgot to save soft mute words :P

packages/client/src/scripts/check-word-mute.ts
```diff
@@ -1,23 +1,27 @@
...
+				return filter.every(keyword => note.text!.includes(keyword))
+			} else {
+				// represents RegExp
+				const regexp = line.match(/^\/(.+)\/(.*)$/);
+				// This should never happen due to input sanitisation.
+				if(!regexp)
+					return false;
+				try {
 					return new RegExp(regexp[1], regexp[2]).test(note.text!);
+				} catch (err: SyntaxError) {
```
rinsuki: You can't write type annotation in catch clause (TS1196).

```suggestion
				} catch (err) {
```

rinsuki (changes requested):

packages/client/src/scripts/check-word-mute.ts
Johann150: Ah I see. I thought I read I could do that, but apparently I did not read thoroughly enough that i can only use `any` or `unknown`. I assume the same applies in `packages/backend/src/misc/check-word-mute.ts`?

rinsuki: yes.

packages/backend/src/server/api/endpoints/i/update.ts
```diff
@@ -113,7 +114,7 @@ export const meta = {
 		},
 
 		mutedWords: {
-			validator: $.optional.arr($.arr($.str)),
+			validator: $.optional.arr($.arr($.str.min(1)).min(1).or($.str)),
```
Johann150, line 117: This does not do what I thought. But is there a way to represent `Array<string | string[]>` in cafy?

packages/backend/migration/1644010796173-convert-hard-mutes.js
```diff
@@ -0,0 +1,48 @@
...
+
+    async up(queryRunner) {
+        (await queryRunner.query(`SELECT "userId", "mutedWords" FROM "user_profile"`)).forEach(entry => {
+                let words = entry.mutedWords
+                        .map(line => {
+                                const regexp = line.join(" ").match(/^\/(.+)\/(.*)$/);
+                                if (regexp) {
+                                        // convert regexp's
+                                        try {
+                                                new RegExp(regexp[1], regexp[2]);
```
acid-chicken: it should use `RE2` here too, instead of built-in `RegExp`?

packages/backend/migration/1644010796173-convert-hard-mutes.js
```diff
@@ -0,0 +1,49 @@
...
+                                        }
+                                } else {
+                                        // remove empty segments
+                                        return line.filter(x => x !== '');
+                                }
+                        })
+                        // remove empty lines
+                        .filter(x => !(Array.isArray(x) && x.length === 0));
+
+                queryRunner.query(`UPDATE "user_profile" SET "mutedWords" = '${JSON.stringify(words)}'::jsonb WHERE "userId" = '${entry.userId}'`);
```
syuilo: ここでSQLインジェクションが可能にならないか不安があります👀

Johann150: do you think this would be enough?
```suggestion
                let json = let json = JSON.stringify(words).replaceAll("'", "''");
                queryRunner.query(`UPDATE "user_profile" SET "mutedWords" = '${json}'::jsonb WHERE "userId" = '${entry.userId}'`);
```

robflop: I gave TypeORM's docs a quick check, and from what I gather this is how it should be done:

```suggestion
                queryRunner.query(`UPDATE "user_profile" SET "mutedWords" = ':json'::jsonb WHERE "userId" = ':userId'`, { json: json, userId: entry.userId });
```

From here: https://typeorm.io/#/select-query-builder/using-parameters-to-escape-data

Johann150: I tried that before, but that does not work because `queryRunner` is a `QueryRunner` and not a `SelectQueryBuilder`.

Johann150: I found a way to get a query builder from a query runner.

syuilo (approved):
LGTM

mei23:
Approvedなら別にいいけど、マイグレーションって必須？
この辺の対応項目ってIssueで最初に議論するんじゃなかったのだわ？

syuilo:
理想的には実装前にIssueで実装方針の認識が合わされてた方がベターだね

mei23:
最近みんな CONTRIBUTING ちゃんと読んでる？ってのがあるわ (contributor, collaborator 問わず)

syuilo:
🙏

tamaina:
ソフトミュートを設定しているとバグる（が私はまだコードを呑み込めていない）

tamaina:
check-word-muteのmatched内のreturn falseが問題やな

syuilo:
ミュート設定しなおすと直ると思う

tamaina:
> ミュート設定しなおすと直ると思う

それはいただけない  
ブラウザコンソールとか開かない限り設定前の値が出てこないし

tamaina:
とりあえずreturn falseせずに後方互換性を確保したい

tamaina:
いやなんか勘違いしてた

tamaina:
(return falseのせいではなかった

PR #8260, opened 2022-02-06
fix: instance ticker

### LINKED ISSUE #8205: Setting instance information of notes to always show breaks the timeline
## 💡 Summary

I tried setting the option "Instance information of notes" to always show, but this seems to break the timeline. Upon loading the timeline, no posts load on Home, and I'm not able to switch to any other timeline.

## 🙂 Expected Behavior

Posts show normally in the timeline and local notes display instance information.

## ☹️ Actual Behavior

Timeline posts don't load at all. Timeline selection buttons are unresponsive.

## 📝 Steps to Reproduce

1. set "Instance information of notes" to Always show
2. press OK in the pop-up "this requires a reload"
3. go to home timeline

## 📌 Environment

Misskey 12.103.1
browser: Brave

Johann150:
# What
1. Restore behaviour as before 21c9705a0f947bf6e3900052987695c67f2d7510 by adding default values that are taken from the local instance.
2. Added a respective type definition.
3. Removed an unnecessary string concatenation (in ec32a62b90b49a044f63b93337e3b6c205411fca).

# Why
Fix #8205 and `TODO` comment.

### DIFF
--- a/packages/client/src/components/instance-ticker.vue
+++ b/packages/client/src/components/instance-ticker.vue
@@ -27,7 +27,7 @@ const instance = props.instance ?? {
 const themeColor = instance.themeColor ?? '#777777';
 
 const bg = {
-	background: `linear-gradient(90deg, ${themeColor}, ${themeColor + '00'})`
+	background: `linear-gradient(90deg, ${themeColor}, ${themeColor}00)`
 };
 </script>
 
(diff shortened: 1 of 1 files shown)


ThatOneCalculator (approved):
Tried on my instance, works great!

syuilo:
👍

PR #8276, opened 2022-02-08
fix: save followers/following visibility

### LINKED ISSUE #8130: Follows/Followers Visibility changes won't be saved unless clicking on an other checkbox
## 💡 Summary
Follows/Followers Visibility changes won't be saved unless I clicking on an other checkbox.

## 🙂 Expected Behavior
Changes should be saved when I modifying Follows/Followers Visibility in Privacy Settings.

## ☹️ Actual Behavior
It seems that the saving operation will only be executed when I clicking on a checkbox. So in the case I just change the visibility, that change won't be saved.

## 📝 Steps to Reproduce

1. Navigate to Settings -> Privacy, find Follows/Followers Visibility
2. Change the visibility only
3. Navigate to anywhere else and then back to step 1.

## 📌 Environment
- Misskey 12.101.1 (docker)
- Firefox 95.0.2

Johann150:
# What
Add update handler for followers/following setting field that properly saves field contents via the API.

# Why
fix #8130

### DIFF
--- a/packages/client/src/pages/settings/privacy.vue
+++ b/packages/client/src/pages/settings/privacy.vue
@@ -8,7 +8,7 @@
 		<template #caption>{{ $ts.makeReactionsPublicDescription }}</template>
 	</FormSwitch>
 		
-	<FormSelect v-model="ffVisibility" class="_formBlock">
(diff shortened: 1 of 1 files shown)


ThatOneCalculator (approved):
Tested and works!

syuilo:
👍

PR #8283, opened 2022-02-10
enhance: search engine setting

### LINKED ISSUE #8266: Let instance admins configure search engine for `google.vue`
## Summary

Many people don't like the Google search engine, letting instance admins change the default search engine here

https://github.com/misskey-dev/misskey/blob/5416f38eb498b4f646747001eb33185128699332/packages/client/src/components/google.vue#L18

in the control panel would be great. I manually edited mine to 
```js
window.open(`https://gowogle.voring.me/search?q=${query.value}`, '_blank');
```

This would be ***especially*** for instance admins who run Whoogle/SearX/SearXNG instances.

Johann150:
# What
Add a setting to change the search engine used for MFM search. The setting is stored only in the client. Any search engine can be used as long as there is a URL that can be opened with HTTP GET. Some search engines are added as suggestions (see list below).

# Why
fix #8266

# Additional info
Please check the Japanese localisation as I only used a machine translator. 🙏

The default search engine was changed to DuckDuckGo as it is considered more user privacy friendly than Google. According to [their website](https://spreadprivacy.com/is-duckduckgo-a-good-search-engine/) the search results are also very good in quality.
It also supports special features, for example `!wja Misskey` will navigate to the Japanese Wikipedia page for Misskey, or `!map Tokio` will search for `Tokio` on Google Maps. (@tamaina is this a fix for #2479 too?)

Search engine suggestions are (alphabetically)
- Bing
- DuckDuckGo
- Ecosia
- Google
- Metager
- Mojeek
- Qwant
- Startpage
- Yahoo JP

If you think I missed an important one, please comment.


4ioskd:
If special features such like "!wja Misskey" and "!map Tokyo" are implemented once, I guess users(also including me) wants more special features and improvements of these. These maybe turn into a burden for Syuilo and contributors.

By the way, Swisscows is also good and privacified search engine.

packages/client/src/pages/settings/general.vue
```diff
@@ -131,6 +138,17 @@ export default defineComponent({
 			lang: localStorage.getItem('lang'),
 			fontSize: localStorage.getItem('fontSize'),
 			useSystemFont: localStorage.getItem('useSystemFont') != null,
+			searchEngines: [
```
ThatOneCalculator, line 141: Might I also suggest a Whoogle instance?
```
'https://gowogle.voring.me/search?q={{query}}',
```

Johann150:
see https://github.com/misskey-dev/misskey/issues/8266#issuecomment-1050540457

PR #8349, opened 2022-02-23
search uses misskey note search

### LINKED ISSUE #2479: 検索構文でGoogle以外の検索エンジンを利用できるように
# Summary
`hoge 検索:misskey`とするとMisskey、`fuga 検索:map`とするとマップサービス、というようにGoogle以外のいくつかの場所でも検索できるようになれば便利

# Environment

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/62595404-google?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github).
</bountysource-plugin>

### LINKED ISSUE #8266: Let instance admins configure search engine for `google.vue`

Johann150:
# What
The note search button leads to misskey's own note search feature instead of Google.

It is federated in HTML like this:
```html
<span>honi [🔍]</span>
```

# Why
With this the privacy problem can be resolved, and no discussion about which search engine should be used will occur again.

fix #8266 (see https://github.com/misskey-dev/misskey/issues/8266#issuecomment-1036406095)
close #8283
close #2479

### DIFF
--- a/packages/backend/src/mfm/to-html.ts
+++ b/packages/backend/src/mfm/to-html.ts
@@ -140,10 +140,9 @@ export function toHtml(nodes: mfm.MfmNode[] | null, mentionedRemoteUsers: IMenti
 		},
 
 		search(node) {
-			const a = doc.createElement('a');
-			a.href = `https://www.google.com/search?q=${node.props.query}`;
-			a.textContent = node.props.content;
-			return a;
+			const el = doc.createElement('span');
+			el.innerText = `${node.props.content} [\ud83d\udd0d]';
+			return el;
 		},
 	};
 
--- a/packages/client/src/components/mfm.ts
+++ b/packages/client/src/components/mfm.ts
@@ -7,7 +7,7 @@ import MkEmoji from '@/components/global/emoji.vue';
 import { concat } from '@/scripts/array';
 import MkFormula from '@/components/formula.vue';
 import MkCode from '@/components/code.vue';
-import MkGoogle from '@/components/google.vue';
+import MkSearch from '@/components/search.vue';
 import MkSparkle from '@/components/sparkle.vue';
 import MkA from '@/components/global/a.vue';
 import { host } from '@/config';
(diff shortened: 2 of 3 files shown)


ThatOneCalculator:
Imo this doesn't make much sense in the context most people *use* the search syntax. 99% of the time it's like `How do I be a femboy? Search` and the funny haha is that it goes to an internet search.

ThatOneCalculator:
It's by no means anything official per say, but I'm [running a poll](https://stop.voring.me/notes/8x3umym9cp) to see what people would prefer. Maybe that could tell us what the best option is :shrug:

Johann150:
I dont see how the poll is relevant to this pull request, since it does not even mention this option.
You can see the linked issue comment for justification, but tl;dr as long as there is a search engine thing inside of Misskey there will most likely be debate about it (most prominent example being the current one) which is generally undesirable, annoying and draining for collaborators. @syuilo had already mentioned that removing the search syntax alltogether might even be a good idea.

ThatOneCalculator:
I'm sorry but searching within Misskey is even worse than Google at this point. It doesn't fulfill the *purpose* of the Search syntax, see my previous comment.

Johann150:
Thank you for illustrating my point about this being a discussion that is dragging on.

syuilo:
see https://github.com/misskey-dev/misskey/issues/8266#issuecomment-1050540457
Thanks anyway

PR #8350, opened 2022-02-23
keepCw on by default

### LINKED ISSUE #8336: Improve "Keep Content Warnings", add `re:` option
## Summary

- Keep Content Warnings should be *on* by default for all users
- There should be an option to enable `re:` on Keep Content warnings.

ex. 
Post is: 
```
Dogs
▼
here is a picture of my dog
```

With re option on, reply is:
```
re: Dogs
▼
⮩ @user what a cute dog!
```

ThatOneCalculator:
Closes #8336

# What

Makes keepCw on by default. Pretty much every other fediverse service does this.

### DIFF
--- a/packages/client/src/store.ts
+++ b/packages/client/src/store.ts
@@ -17,7 +17,7 @@ export const defaultStore = markRaw(new Storage('base', {
 	},
 	keepCw: {
 		where: 'account',
-		default: false
+		default: true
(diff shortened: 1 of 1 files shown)


Johann150 (approved):

syuilo:
👍🏻

PR #8361, opened 2022-02-26
Fix: Only allow admins to access admin views

### LINKED ISSUE #3526: /admin へアクセスすると管理者・モデレーターでなくても管理画面が開けてしまう
# Summary
Misskeyのインスタンスで/admin（例：```https://misskey.xyz/admin```）へ行くと管理者・モデレーター権限がなくても管理画面の表示ができてしまう。

# Expected Behavior
管理者・モデレーター権限がないアカウントで/adminへアクセスする。

ただし、一部内容は表示されない。（ブラウザのコンソールでエラーが出ている）

# Environment
- misskey server - yuzulia.xyz : v10.61.0
- node.js(yuzulia.xyz) : v10.14.1
- misskey server - misskey.xyz : v10.62.0
- node.js(misskey.xyz) : v11.3.0

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/67120771-admin?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github).
</bountysource-plugin>

pixeldesu:
# What

This PR introduces a change that prevents logged out or non-admin users from accessing `/admin` and any subpages, now only showing a 404 page instead of the admin dashboard.

Fixes #3526

# Why

Generally, Misskey's admin dashboard doesn't show mission-critical information, and if so the endpoints require authentication.

In any case, if at some point a bug or an oversight makes more data visible (or actions possible) the simple access to the admin panel could have really bad consequences for an instance.

# Additional info (optional)

I wasn't particularly sure how to implement this, as Vue Router offers multiple ways of approach in itself (global `beforeEach` or component-specific `beforeEnter`) and I saw that other page components had conditions for accounts, rendering another component (`$i ? ... : ...`), which I then opted to use instead.


Johann150:
I think it would be good to address this more generally directly in the `page` function for any page where authentication or special privileges are required.

Johann150 (approved):
Maybe the `props` should also only be passed when the respective page is viewed but I dont think passing properties that are not read is a problem.

syuilo:
Moderators as well as administrators must be allowed.

pixeldesu:
@syuilo Done! I hope using the `iAmModerator` constant is the right way here!

syuilo:
🙏

PR #8367, opened 2022-02-28
fix(meta): Adjust path to contribution guidelines

pixeldesu:
# What

Adjusts the link to the contribution guidelines in the Pull Request template to work again.

# Why

If the link to contribution guidelines doesn't work, this might confuse some developers. Even if the GitHub UI links to the proper one, the one mentioned in the Pull Request template doesn't resolve.

# Additional info (optional)

I did a bit more than changing just the link, I decided to concatenate both comments into a single group (like it is done with the other headings as well) since there is no distinction required as there are no language-specific contribution guidelines anymore.


Johann150 (approved):

syuilo:
Thanks!

PR #8374, opened 2022-03-04
Refactor: Switch Settings index page to setup sugar

pixeldesu:
# What

Simply refactors `pages/settings/index` to use setup sugar instead of the Options API.

# Why

I was actually planning to tackle another issue with Settings and routing, but I was running into some issues and decided on a whim to quickly refactor the component to the new syntax.

# Additional info (optional)

I didn't change any code of the `setup()` method while converting it over (except for the obvious `defineExpose` at the end).

Generally, it seems to work fine, but I'm getting a few typehints that could be solved pretty easily:

The main point in question is the `el` reference, which is defaulted to `null` and so gets an implicit `never` type.

```ts
const el = ref(null);
```

This reference is later used as an `Element`, which makes TypeScript complain. It should be a `ref<Element | null>` instead. I'm just a bit hesitant because no other ref is typed (because they are partically infering their types anyway) and so I'll probably await feedback in the review for what I should do!

### DIFF
--- a/packages/client/src/pages/settings/index.vue
+++ b/packages/client/src/pages/settings/index.vue
@@ -48,7 +48,7 @@ const INFO = ref(indexInfo);
 const page = ref(props.initialPage);
 const narrow = ref(false);
 const view = ref(null);
-const el = ref(null);
+const el = ref<HTMLElement | null>(null);
 const childInfo = ref(null);
 const menuDef = computed(() => [{
 	title: i18n.ts.basicSettings,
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/settings/index.vue
```diff
@@ -34,241 +34,229 @@ import * as symbols from '@/symbols';
...
+	page,
+	menuDef,
+	narrow,
+	view,
+	el,
+	pageProps,
+	component,
+	emailNotConfigured,
+	pageChanged,
+	childInfo,
```
syuilo: When using setup suger, defineExpose is used to define the values you want to expose "outside the component".
So I think the only thing you actually need to expose here is `[symbols.PAGE_INFO]`.

syuilo:
> The main point in question is the el reference, which is defaulted to null and so gets an implicit never type.

In other components, it is written like `ref<HTMLElement | null>()`.

ThatOneCalculator (approved):

syuilo:
Thanks!

PR #8378, opened 2022-03-06
Fix: show login dialog on settings page if not logged in

### LINKED ISSUE #8377: User settings shouldn't be accessible when logged out
## 💡 Summary

## 🥰 Expected Behavior

## 🤬 Actual Behavior

## 📝 Steps to Reproduce

3. Be presented an empty page

## Additional Info

pixeldesu:
# What

This PR adds a login dialog to the settings page if a visitor is not logged in, and if they try to dismiss it, they will be redirected to the root page instead.

Fixes #8377

# Why

Generally, there is no kind of data leak on settings, because a user is required to even see anything, it still isn't a nice practise to possibly encounter an empty page if you don't need to, showing a dialog to log (back) in is way better.

# Additional info (optional)

As mentioned in my issue, I adjusted `XSigninDialog` to also emit the `close` event, which is required to handle what happens on manual closing of the dialog.

Also, I extracted the logic for this into a method `promptLoginOrRedirect(path: string = '/')` to easily reuse this in any other view that requires authentication!

### DIFF
--- a/packages/client/src/components/signin-dialog.vue
+++ b/packages/client/src/components/signin-dialog.vue
@@ -24,7 +24,7 @@ const props = withDefaults(defineProps<{
 
 const emit = defineEmits<{
 	(e: 'done'): void;
-	(e: 'close'): void;
+	(e: 'cancelled'): void;
 	(e: 'closed'): void;
 }>();
 
@@ -36,7 +36,7 @@ function onLogin(res) {
 }
 
 function onClose() {
-	emit('close');
+	emit('cancelled');
 	dialog.close();
 }
 </script>
--- a/packages/client/src/components/signin-dialog.vue
+++ b/packages/client/src/components/signin-dialog.vue
@@ -2,7 +2,7 @@
 <XModalWindow ref="dialog"
 	:width="370"
 	:height="400"
-	@close="dialog.close()"
+	@close="onClose"
 	@closed="emit('closed')"
 >
 	<template #header>{{ $ts.login }}</template>
@@ -24,6 +24,7 @@ const props = withDefaults(defineProps<{
 
 const emit = defineEmits<{
 	(e: 'done'): void;
+	(e: 'close'): void;
 	(e: 'closed'): void;
 }>();
 
@@ -31,6 +32,11 @@ const dialog = $ref<InstanceType<typeof XModalWindow>>();
 
 function onLogin(res) {
 	emit('done', res);
+	dialog.close();
+}
+
+function onClose() {
+	emit('close');
 	dialog.close();
 }
 </script>
--- a/packages/client/src/pages/settings/index.vue
+++ b/packages/client/src/pages/settings/index.vue
@@ -33,6 +33,7 @@ import { unisonReload } from '@/scripts/unison-reload';
 import * as symbols from '@/symbols';
 import { instance } from '@/instance';
 import { $i } from '@/account';
+import { popup } from '@/os';
 
 const props = defineProps<{
   initialPage?: string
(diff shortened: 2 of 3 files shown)


ThatOneCalculator (approved):

packages/client/src/components/signin-dialog.vue
```diff
@@ -2,7 +2,7 @@
 <XModalWindow ref="dialog"
 	:width="370"
 	:height="400"
-	@close="dialog.close()"
+	@close="onClose"
 	@closed="emit('closed')"
```
Johann150, line 6: Couldn't you have use the existing `closed` event?

packages/client/src/pages/settings/index.vue
```diff
@@ -212,6 +213,8 @@ const component = computed(() => {
 	return null;
 });
 
+promptLoginOrRedirect();
```
Johann150, line 216: I think this would usually be a case for `pleaseLogin();`. Maybe it should be used here too. Although in my opinion, the user experience of `pleaseLogin` is not good since there is no way to log in when you see the dialog, so maybe it should be refactored.

But having two scripts that do the same thing would be against DRY.

Johann150 (changes requested):
I like the idea from the user experience side, but I'm not sure about the implementation side.

packages/client/src/components/signin-dialog.vue
pixeldesu: From the PR description:

> As mentioned in my issue, I adjusted `XSigninDialog` to also emit the `close` event, which is required to handle what happens on manual closing of the dialog.

`closed` always gets emitted when the dialog is closed, which also happens after a successful login, so this would trigger a redirect after successfully logging in, and from my tests that caused race conditions with actually setting the account data (or: the login dialog would always show up)

`close` is a different event triggered when closing the modal manually, which fits the use case here.

packages/client/src/pages/settings/index.vue
pixeldesu: I agree with that. I saw the method too, but seeing it only sent off a small warning instead of actually prompting a login made me not use it.

I'm generally a bit cautious with going too far out of scope with PRs, because `pleaseLogin` is used in at least 3 other places.

Once the "new" method is agreed on, we can switch it out in the other places. This keeps what needs to be reviewed smaller and more thematically grouped, I'd say.

----

Thinking about it, the sign in dialog should probably be extended to include a notice why the dialog was opened, like Mastodon does it if you access a route that requires authentication:

Screenshot 2022-03-07 at 16 58 00

packages/client/src/components/signin-dialog.vue
Johann150: Okay, but then I still think the event is named confusingly. Maybe `cancelled` is a better name?

pixeldesu: I agree. Should I rename this just local to the sign-in modal for now? I wouldn't want to go further than that because as soon as we get to `XModalWindow` there's quite a few Vue handlers on that event (I was able to easily search for `@close=`) and I'd rather refactor that in it's own PR.

Johann150: Yes, I was only referring to the new event you created.

pixeldesu: Renamed the event that is emitted to `cancelled` now.

Leaving the handler as `onClose` however, as that is the parent event name still.

pixeldesu:
Succeeded by #8630

PR #8388, opened 2022-03-09
Fix: warn about outdated NodeJS fixed

### LINKED ISSUE #8387: Warn user if `node` is outdated
## Summary

Since a lot of people are coming from 12.107.0 and many don't have Node 16.14.0 installed yet, maybe have Misskey exit and display a message telling them to install the latest ver if `SyntaxError: Unexpected identifier` occurs. (As seen in errors such as #8386)

Proposed output:

```
$ NODE_ENV=production npm run start

> misskey@12.108.0 start
> cd packages/backend && node --experimental-json-modules ./built/index.js
  _____ _         _
 |     |_|___ ___| |_ ___ _ _
 | | | | |_ -|_ -| '_| -_| | |
 |_|_|_|_|___|___|_,_|___|_  |
 v12.108.0               |___|
 Misskey is an open-source decentralized microblogging platform.
 If you like Misskey, please donate to support development. https://www.patreon.com/syuilo
--- destiny.zotan.network (PID: 1362295) ---
INFO *        [core boot]        Welcome to Misskey!
INFO *        [core boot]        Misskey v12.108.0
INFO *        [core boot env]        NODE_ENV: production
INFO *        [core boot env]        You do not have root privileges
INFO *        [core boot nodejs]        Version 16.13.2
DONE *        [core boot config]        Loaded
INFO *        [core boot db]        Connecting...
DONE *        [core boot db]        Connected: v13.4
DONE *        [core boot]        Misskey initialized
INFO *        [core boot]        Starting 1 worker...
SyntaxError: Unexpected identifier
    at ESMLoader.moduleStrategy (node:internal/modules/esm/translators:139:18)
    at ESMLoader.moduleProvider (node:internal/modules/esm/loader:236:14)

===============================
Please install NodeJS 16.14.0!
===============================

$ 
```

ThatOneCalculator:
# What
Fixes #8387

# Why
Administrators may be confused about new node version requirement


Johann150:
I think this could be generalized a bit more, syntax errors are probably not the right and/or only thing to look for. For example, it would make sense to check the Node.js version at startup, e.g. in `packages/backend/src/boot/index.ts`.

I think the currently required Node version is in `.node-version`, but I could not find any place where that file is or was used (using `grep -r` and `git log -S`). I assume that is what the file is for? Maybe one of the other collaborators can comment on that.

ThatOneCalculator:
> For example, it would make sense to check the Node.js version at startup, e.g. in packages/backend/src/boot/index.ts.

That's the file I changed already, but I'll just make it check the node ver instead.

ThatOneCalculator:
@Johann150 how's that? (I'm not quite sure how to read in `.node-version` so I just made it a string)

ThatOneCalculator:
Changed as discussed privately :)

pixeldesu:
@Johann150 Not sure if you found out already, but `.node-version` is apparently some sort of standard used by version managers to figure out the version to be used.

Searched for it and found following repository pretty much explaining it:
https://github.com/shadowspawn/node-version-usage

ThatOneCalculator:
Ah so it might be for something like `nvm`. In that case, this PR should go forward as it is, since not everyone uses `nvm`

nullobsi:
If you're going to do a check, at least make it error on older versions and a warning on newer versions. I shouldn't have to edit a JS file when v16.14.1 comes out

ThatOneCalculator:
> If you're going to do a check, at least make it error on older versions and a warning on newer versions. I shouldn't have to edit a JS file when v16.14.1 comes out

Ah yea good idea lol

ThatOneCalculator:
@Johann150 @nullobsi how about now?

nullobsi:
It looks better, but as Johann mentioned there is already a .node-version file that could probably be used

ThatOneCalculator:
> It looks better, but as Johann mentioned there is already a .node-version file that could probably be used

I have no clue how to open it lol, any suggestions?

Johann150:
I think this is what would actually need to be fixed https://github.com/misskey-dev/misskey/blob/6cd3ff5848426c766958f52238548c4c897a0a23/packages/backend/src/boot/master.ts#L91-L93

ThatOneCalculator:
@Johann150 Apologies for the late follow-up, is this better?

packages/backend/src/boot/master.ts
```diff
@@ -7,6 +7,7 @@ import chalk from 'chalk';
 import chalkTemplate from 'chalk-template';
 import * as portscanner from 'portscanner';
 import { getConnection } from 'typeorm';
+import semver  from 'semver';
```
acid-chicken, line 10:
```suggestion
import semver from 'semver';
```

acid-chicken (commented):
looks mostly ok to me

packages/backend/src/boot/master.ts
acid-chicken:
```suggestion
import semver from 'semver';
import { getConnection } from 'typeorm';

File: packages/backend/package.json [hunk 2]
```diff
@@ -176,5 +177,8 @@
 		"@types/fluent-ffmpeg": "2.1.20",
 		"cross-env": "7.0.3",
 		"execa": "6.1.0"
+	},
+	"engines": {
+		"node": ">=16.14.0"
```
Johann150: This is still left over from an earlier attempt, should probably remove it.

packages/backend/src/boot/master.ts
```diff
@@ -110,8 +107,9 @@ function showNodejsVersion(): void {
 
 	nodejsLogger.info(`Version ${runningNodejsVersion.join('.')}`);
```
Johann150: `runningNodejsVersion` was removed above.

```suggestion
	nodejsLogger.info(`Version ${process.version} detected.`);
```

packages/backend/src/boot/master.ts
```diff
@@ -110,8 +107,9 @@ function showNodejsVersion(): void {
...
 
-	if (!satisfyNodejsVersion) {
-		nodejsLogger.error(`Node.js version is less than ${requiredNodejsVersion.join('.')}. Please upgrade it.`, null, true);
+	const minVersion = 'v16.14.0';
```
Johann150: I think it would be a good idea to read the version from the `.node-version` file, otherwise we could end up in the same situation again where only one of the two locations was maintained.

```suggestion
	const minVersion = fs.readFileSync(`${_dirname}/../../../../.node-version`, 'utf-8').trim();
```

packages/backend/src/boot/master.ts
```diff
@@ -110,8 +107,9 @@ function showNodejsVersion(): void {
...
+	if (semver.lt(process.version, minVersion)) {
+		nodejsLogger.error(`Please make sure you've installed and are using NodeJS >=${minVersion}! You're currently on NodeJS ${process.version}.`);
```
Johann150: No need to print the running version again. I'd also rephrase it to be more straight to the point since this is a logfile.
```suggestion
		nodejsLogger.error(`At least Node.js ${minVersion} required!`);
```

Johann150 (changes requested):

packages/backend/package.json
ThatOneCalculator: I've found that having this in actually stops `yarn`	 whenever you try to `yarn add <package>`, which can be useful.

ThatOneCalculator:
@Johann150 thoughts?

packages/backend/package.json
Johann150: Please remove this. I did not notice any difference in behaviour from `yarn add` and I also do not think we should break functionality on purpose.

ThatOneCalculator:
@Johann150 changed the package.json

Johann150 (approved):

acid-chicken (approved):

syuilo:
Thank you guys

PR #8412, opened 2022-03-15
enhance: Service Workerを常に登録するように

### LINKED ISSUE #8403: Progressive Web App,Android & iOS app generation
Although misskey serves both a manifest and service worker, it is not a progressive web app(PWA) (at least as of version 12.108) . This is because the service worker is registered only after a user sign-in is complete and does not cache any page although [almost entire site](https://misskey-hub.net/en/docs/admin/cdn.html#caching) is static . The service worker is used only for push notifications.
 Making it a PWA will enable easy android, iOS and windows app generation using https://www.pwabuilder.com which is not possible now as the pwabuilder crawler is not able to detect the service worker(because it acesses the site as a non signed-in user). 
sw
The admin panel can be changed to convert the given option(in pic) into two different ones-
a)Make a PWA
b)Enable push notifictions
Enabling (a) will be a prerequisite for enabling (b). At the same time. this will allow a user to generate a PWA without notifications by enabling (a) but disabling (b).

_Originally posted by @bil-ash in https://github.com/misskey-dev/misskey/discussions/8402_

tamaina:
Resolve #8403

# What
Always register Service Worker so that PWABuilder will be able to detect Misskey as PWA.

# Why

# Additional info (optional)
p1.a9z.devで動作中


tamaina:
image

よさそう

bil-ash:
This pull request only changes the time of service worker registration but does not cache any extra file. So if an android user has installed the PWA(either generated by chrome or manully by pwabuilder and submitted to store), and some days later opens the app while being offline, the user will  see an error. There are two ways of resolving this-
1) Cache an offline.html page whose content can be something like `You are offline. Close misskey,connect to internet and open again`.
2) Cache all requests other than the /api/* ones. Thus, a user will be able to see the skeleton of misskey while even being offline, and as soon as he/she goes online will be able to see the feed/interact.

The required service workers for both the cases can be found at https://github.com/pwa-builder/pwabuilder-serviceworkers/tree/b442633de19760dea482b21e8725789aaa270fe1 . Newer commits use workbox which bring an extra dependency which I think should be avoided

bil-ash:
swo
As of now, this is what happens on desktop(in the installed PWA) when the user is offline and on mobile also similar thing will happen. So we need to provide meaningfull feedback to user

tamaina:
It should display the message properly when offline, I thought, but then I realized that was a feature to be added in #7667.

I have no idea what causes the situation as shown in the screenshot. With a browser data reset and private browsing, I can see normal.

image

bil-ash:
How did you open an installed PWA in private browsing? I am taking about the case of installed PWA in offline mode

ThatOneCalculator:
> How did you open an installed PWA in private browsing? I am taking about the case of installed PWA in offline mode

With Firefox PWAs you can specify the profile, including private browsing.

tamaina:
> How did you open an installed PWA in private browsing? I am taking about the case of installed PWA in offline mode

Chromium's offline mode does not seem to be able to reproduce the complete offline experience.  
If it is completely offline, the following screen will appear.

image

bil-ash:
> > How did you open an installed PWA in private browsing? I am taking about the case of installed PWA in offline mode
> 
> Chromium's offline mode does not seem to be able to reproduce the complete offline experience. If it is completely offline, the following screen will appear.
> 
> image

Yes, that's why we will also need to cache a offline page. (Chrome's behaviour seams proper to me- we are not caching any page, nor responding to any page routes via the service worker, so we get an almost blank experience when offline. Firefox's (improper) behaviour when offline is most probably from browser cache and not service worker cache. )
 For a proper offline experience we will have to https://github.com/misskey-dev/misskey/pull/8412#issuecomment-1068674462

bil-ash:
I have created a PR (and then deleted it) so that I can provide the additional code which I think also needs to be added for proper offline behaviour

ThatOneCalculator:
Why not just put that in a review?

tamaina:
#7667 がマージされたのでclose

PR #8429, opened 2022-03-19
fix(federation): avoid duplicate activity delivery

### LINKED ISSUE #8428: ActivityPub delivery can cause race conditions in 3rd party software
## 💡 Summary
If both a `Followers` and `Direct` recipe are added to the Deliver Manager, two activities may be sent to the same instance.
This can for example happen when sending a reaction, because it is sent directly to the author of the note and to all followers.

I came across this while testing something and it caused a race condition in Pleroma, leading to the activities not being performed correctly. It also leads to unecessary network congestion and load on remote servers.

## 🥰 Expected Behavior

Each activity is only sent to each instance once.

## 🤬 Actual Behavior

Activities may be sent two or more times.

## 📝 Steps to Reproduce

There are multiple ways in which this issue can occur and because it is a race condition in 3rd party software, it may not always work. I found that it often happened when doing this:

1. React to note from a Pleroma instance.
2. Remove reaction.
3. Reaction is not removed on Pleroma.

## 📌 Environment

Misskey version: at least all versions since 44ab428803c086255ad552502764238589b4860b

## 🔧 Possible Solutions

The issue occurs because of how `Direct` recipies are handled:
https://github.com/misskey-dev/misskey/blob/c928941d297c5db383afcfd7df8df8a0be3d7caa/packages/backend/src/remote/activitypub/deliver-manager.ts#L95-L99

If multiple recipients from `Direct` recipies happen to have the same shared inbox, Misskey ignores this. Misskey also ignores if the shared inbox of a direct recipient is already in the inboxes list.

This could be solved by:
1. Always use the shared inbox if it is available for a user. This ensures deduplication across direct and follower recipies.
2. When building inboxes, get both the inbox and shared inbox. Then deduplicate by shared inbox. If only one user from a specific instance is in the list, they will get the message to their individual inbox.

Alternative 1 is easier to implement. I think alternative 2 is just unnecessary work because individual and shared inboxes are often handled exactly the same anyway (Misskey also handles them the same). Therefore I think alternative 1 is the best solution.

Johann150:
# What
Always prefer shared inboxes over individual inboxes to avoid sending an activity twice to the same instance.

# Why
fix #8428

### DIFF
--- a/packages/backend/src/remote/activitypub/deliver-manager.ts
+++ b/packages/backend/src/remote/activitypub/deliver-manager.ts
@@ -109,7 +109,7 @@ export default class DeliverManager {
 			}
 		}
 
-		this.recipes.filter(recipe => {
+		this.recipes.filter((recipe): recipe is IDirectRecipe => {
 			// followers recipes have already been processed
 			isDirect(recipe)
 			// check that shared inbox has not been added yet
--- a/packages/backend/src/remote/activitypub/deliver-manager.ts
+++ b/packages/backend/src/remote/activitypub/deliver-manager.ts
@@ -79,7 +79,13 @@ export default class DeliverManager {
 
 		const inboxes = new Set<string>();
 
-		// build inbox list
+		/*
+		build inbox list
+
+		Always prefer shared inbox over individual inbox to avoid duplication.
+		For example if a direct recipe is combined with a followers recipe,
+		or multiple directs from the same instance are present.
+		*/
 		for (const recipe of this.recipes) {
 			if (isFollowers(recipe)) {
 				// followers deliver
@@ -105,7 +111,7 @@ export default class DeliverManager {
 				}
 			} else if (isDirect(recipe)) {
 				// direct deliver
-				const inbox = recipe.to.inbox;
+				const inbox = recipe.to.sharedInbox || recipe.to.inbox;
 				if (inbox) inboxes.add(inbox);
 			}
 		}
--- a/packages/backend/src/remote/activitypub/deliver-manager.ts
+++ b/packages/backend/src/remote/activitypub/deliver-manager.ts
@@ -82,39 +82,42 @@ export default class DeliverManager {
 		/*
 		build inbox list
 
-		Always prefer shared inbox over individual inbox to avoid duplication.
-		For example if a direct recipe is combined with a followers recipe,
-		or multiple directs from the same instance are present.
+		Process follower recipes first to avoid duplication when processing
+		direct recipes later.
 		*/
-		for (const recipe of this.recipes) {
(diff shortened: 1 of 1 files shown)


mei23:
For direct recipients, you need to send it to inbox.
It is not good if it is applied to other than Like, so it seems better to limit the change to Like (Reaction) only. 
https://github.com/misskey-dev/misskey/commit/c24ba086581b8546396c295d4c7fcf39af77cf68

Edit: I want to change this further.

mei23:
Isn't it better to do the following instead of **Always**? 

- If the same sharedInbox has already been added to the deliver list, use that sharedInbox.
- If not, use inbox. (not a sharedInbox, Prevent sharedInbox from being used when only Direct is specified.) 

I think shared Inbox is for follower delivery.
https://www.w3.org/TR/activitypub/#shared-inbox-delivery
> When an object is being delivered to the originating actor's followers

mei23:
https://github.com/misskey-dev/misskey/pull/8429#issuecomment-1073225127
As below
```ts
// direct deliver
if (recipe.to.sharedInbox && inboxes.has(recipe.to.sharedInbox)) {
	// do nothing
} else if (recipe.to.inbox) {
	inboxes.add(recipe.to.inbox);
}
```

Oh, I have to change the hierarchy of the loop.

Johann150:
I think that would be the same as alternative 2 from the original issue and I also considered that. If you say it makes a significant difference then I'll change it to that.

mei23:
I don't know if it's the same as alternative 2, but my image is below. I think that is better if possible.
```ts
const inboxes = new Set<string>();

for (const recipe of this.recipes.filter((x): x is IFollowersRecipe => isFollowers(x))) {
	// followers deliver
	const followers = await Followings.find({
		followeeId: this.actor.id,
	});

	for (const following of followers) {
		if (Followings.isRemoteFollower(following)) {
			const inbox = following.followerSharedInbox || following.followerInbox;
			inboxes.add(inbox);
		}
	}
}

for (const recipe of this.recipes.filter((x): x is IDirectRecipe => isDirect(x))) {
	// direct deliver
	if (recipe.to.sharedInbox && inboxes.has(recipe.to.sharedInbox)) {
		// do nothing
	} else if (recipe.to.inbox) {
		inboxes.add(recipe.to.inbox);
	}
}
```

Johann150:
I did it a bit differently, because follow recipes always have the same effect so there is no need to loop over all follow recipes.

mei23:
Thank you, I think this is fine.

Johann150:
Thank you for the review 🙇

syuilo:
Thanks!

PR #8434, opened 2022-03-21
chore: AP Undo activity id にユニークidを使用するように

mei23:
# What
ActivityPub Undo activity id にユニークidを使用するように

```yaml
  type: Undo
  id: 'https://mk11.m213.xyz/2fdb93e8-85a5-417c-b82c-9fd46141aa05' # UUID
  object:
    type: Like
    id: 'https://mk11.m213.xyz/likes/:id'
```
↓
```yaml
  type: Undo
  id: 'https://mk11.m213.xyz/likes/:id/undo', # :id/undo
  object:
    type: Like
    id: 'https://mk11.m213.xyz/likes/:id'
```

# Why
直接関係ないけどいちおう
https://github.com/misskey-dev/misskey/issues/8428#issuecomment-1073218516

# Additional info (optional)

### DIFF
--- a/packages/backend/src/remote/activitypub/renderer/undo.ts
+++ b/packages/backend/src/remote/activitypub/renderer/undo.ts
@@ -3,9 +3,11 @@ import { ILocalUser, User } from '@/models/entities/user.js';
 
 export default (object: any, user: { id: User['id'] }) => {
 	if (object == null) return null;
+	const id = typeof object.id === 'string' && object.id.startsWith(config.url) ? `${object.id}/undo` : undefined;
 
 	return {
 		type: 'Undo',
+		...(id ? { id } : {}),
 		actor: `${config.url}/users/${user.id}`,
 		object,
 		published: new Date().toISOString(),


packages/backend/src/remote/activitypub/renderer/undo.ts
```diff
@@ -3,9 +3,11 @@ import { ILocalUser, User } from '@/models/entities/user.js';
 
 export default (object: any, user: { id: User['id'] }) => {
```
Johann150, line 4: I think we could add a type here:

```suggestion
export default (object: IObject, user: { id: User['id'] }) => {
```

The type check in line 5 makes me unsure wether we should allow `null`, from a cursory review I could not find a place where `null` may be passed in, nor does it make sense to do so.

packages/backend/src/remote/activitypub/renderer/undo.ts
```diff
@@ -3,9 +3,11 @@ import { ILocalUser, User } from '@/models/entities/user.js';
...
 	if (object == null) return null;
+	const id = typeof object.id === 'string' && object.id.startsWith(config.url) ? `${object.id}/undo` : undefined;
 
 	return {
 		type: 'Undo',
+		...(id ? { id } : {}),
```
Johann150, line 10: If we add the type above we can use `getApId`:
```suggestion

	return {
		type: 'Undo',
		id: getApId(object) + '/undo',
```

It has to be imported:
```ts
import { IObject, getApId } from '@/remote/activitypub/type.js';
```

I'm not sure if its necessary to check for `config.url`? It does not make sense to render an `Undo` for remote content and I don't think we are doing that anywhere. Even `renderFollowRelay` uses an ID created from the instance's URL.

packages/backend/src/remote/activitypub/renderer/undo.ts
mei23: - リファクタではないので今回そこを修正する予定はない
- IObjectはinboxに入ってくる方の型なので厳密には違う

packages/backend/src/remote/activitypub/renderer/undo.ts
mei23: 今はなくても、定義上 id は optional なため。
ローカルチェックもねんのため。

syuilo:
🙏

PR #8440, opened 2022-03-22
fix(federation): recognize null in _misskey_content for notes

### LINKED ISSUE #8439: quote renotes are not federated correctly
## 💡 Summary

When a quote renote contains no text and only an image it is not federated correctly: The `RN: url...` text that is meant for non-Misskey activitypub software is shown instead.

This probably occurs in this line, due to _misskey_content being `null`:
https://github.com/misskey-dev/misskey/blob/08bace6c7db7b34f271c270c0f81d70f0b412438/packages/backend/src/remote/activitypub/models/note.ts#L200

## 🥰 Expected Behavior

The note is displayed the same as on the original instance.

## 🤬 Actual Behavior

A quote renote that only contains an image on the remote instance now contains text like `RN: url...`.

## 📝 Steps to Reproduce

1. Quote another note without text and only an image.
2. View the note on another instance.
3. Text is displayed.

## 📌 Environment

Misskey version: 12.108.1

Johann150:
# What
Correctly recognize `null` (and also empty string) in `_misskey_content` on incoming notes.

# Why
fix #8439

### DIFF
--- a/packages/backend/src/remote/activitypub/models/note.ts
+++ b/packages/backend/src/remote/activitypub/models/note.ts
@@ -197,7 +197,7 @@ export async function createNote(value: string | IObject, resolver?: Resolver, s
 	const cw = note.summary === '' ? null : note.summary;
 
 	// テキストのパース
-	const text = note._misskey_content || (note.content ? htmlToMfm(note.content, note.tag) : null);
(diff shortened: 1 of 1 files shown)


ThatOneCalculator (approved):

syuilo:
👍

Issue #8317, opened 2022-02-15
Messages that aren't supposed to be accesible are still (kind of) visible


Solsticely:
## 💡 Summary

Private messages' content are substituted by parenthesis 
![image](https://user-images.githubusercontent.com/96320174/153989340-c9a988d6-7872-488f-8f33-dc0713baf022.png)

These messages are accessible through the user interface. I can spot a few just going to the "notes and responses" section of a profile and finding a publicly readable response to a private message.

## 🥰 Expected Behavior

Messages not meant to be seen by the user should not be visible to the user, or should atleast have a description along the lines of "This message is private/hidden/is part of a private discussion between other users/etc"

## 🤬 Actual Behavior

Parenthesies ( )w( )

## 📝 Steps to Reproduce

Here's a conversation: https://stop.voring.me/notes/8wprsu5kyb

A general guide to achieving this:

1. Find a profile
2. Go to their "Notes and Responses" section
3. Find a response to a private note (look for `()`s)
4. :tada: 

## 📌 Environment

Misskey version: 12.105.0
Your OS: Solus Budgie
Your browser: Tested on Firefox and Chromium

Johann150:
I'm a bit confused as to what you think is the problem here, as I can see multiple issues here.

I would consider `()` a form of description that this content is unavailable. I don't think users should normally see this so it would be unnecessary effort to add a localisation phrase for this. If someone were to find the exact URL of a direct note they should not see, I think it would be more appropriate for the server to return a `404` error code (as to not give away that a note with this ID exists) and the client to display the respective error message. There should not be a special error message for this.
I tested this special case on Mastodon and it does not display the replied to messages at all.

The other issue is in my opinion that replies to direct notes should be required to also be direct notes. This is also the case on Pleroma (but not Mastodon).

Johann150:
I discovered that this is also an i18n bug because the `private` localisation string was removed.

https://github.com/misskey-dev/misskey/blob/45462e4a5e8b94be110968afb2985bb145d7b6b3/packages/client/src/components/sub-note-content.vue#L4-L5 (note that also `deleted` does not exist as a localisation string)

https://github.com/misskey-dev/misskey/blob/b01a0325ba8b682da7fc44764d6abb9c320aa90e/packages/client/src/components/note.vue#L51

https://github.com/misskey-dev/misskey/blob/b01a0325ba8b682da7fc44764d6abb9c320aa90e/packages/client/src/components/note-detailed.vue#L65

[...]

PR #8448, opened 2022-03-26
enhance(federation): limit federation of reactions on direct notes

Johann150:
# What
Reactions to notes that have `specified` visibility are only federated to the people that the note is visible to, instead of delivering it to all followers.

# Why
Related to #8317, but does not yet fix it.

Without this, remote instances may be able to gather information about a private note.

### DIFF
--- a/packages/backend/src/services/note/reaction/create.ts
+++ b/packages/backend/src/services/note/reaction/create.ts
@@ -124,7 +124,16 @@ export default async (user: { id: User['id']; host: User['host']; }, note: Note,
 			const reactee = await Users.findOneBy({ id: note.userId });
 			dm.addDirectRecipe(reactee as IRemoteUser);
 		}
-		dm.addFollowersRecipe();
+
+		if (['public', 'home', 'followers'].includes(note.visibility)) {
+			dm.addFollowersRecipe();
+		} else if (note.visibility === 'specified') {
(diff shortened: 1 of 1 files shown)


ThatOneCalculator (approved):

packages/backend/src/services/note/reaction/create.ts
```diff
@@ -124,7 +124,16 @@ export default async (user: { id: User['id']; host: User['host']; }, note: Note,
 			const reactee = await Users.findOne(note.userId);
 			dm.addDirectRecipe(reactee as IRemoteUser);
 		}
-		dm.addFollowersRecipe();
+
+		if (['public', 'home', 'followers'].includes(note.visibility)) {
+			dm.addFollowersRecipe();
+		} else if (note.visibility === 'specified') {
+			const visibleUsers = await Promise.all(note.visibleUserIds.map(id => Users.findOne(id)));
```
syuilo: Misskey has been migrated to TypeORM 3.0. In 3.0, you need to write as follows: `Users.findOneBy({ id })`

ThatOneCalculator: Oh yea, forgot about that, still haven't pulled that to my instance yet

Johann150: I think I forgot to pull the up to date version so I got confused by code still using the old way. Has been fixed now.

syuilo:
👍🏻

PR #8456, opened 2022-03-31
enhance(doc): required input fields

### LINKED ISSUE #2006: APIに対して虚無JSONを送るとレスポンスが帰ってこない
https://misskey.xyz/notes/5b5ab8e040c98369543c4422

<bountysource-plugin>

---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/61400889-api-json?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F51027576&utm_medium=issues&utm_source=github).
</bountysource-plugin>

Johann150:
## What
- Added which fields are required by the API if there are multiple variants. This improves validation and documentation.
    - Some specific errors can no longer occur because of the stricter validation and have therefore been removed.
- A few other small enhancements to the documentation such as documenting default values and adding descriptions for unclear attributes.

## Why
Improve the documentation (ROADMAP). Because fixing the documentation also fixes the input validation, this also fixes #2006.

## Additional Info
I tried to break it up into multiple commits to make it easier to review. Some additional information is in the commit messages.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,9 @@ You should also include the user name that made the change.
 ### Bugfixes
 - API: fix endpoint endpoint @Johann150
 - API: fix admin/meta endpoint @syuilo
+- API: improved validation and documentation for endpoints that accept different variants of input @Johann150
+- API: `notes/create`: The `mediaIds` property is now deprecated. @Johann150
+  - Use `fileIds` instead, it has the same behaviour.
 
 ## 12.108.1 (2022/03/12)
 
--- a/packages/backend/src/models/schema/emoji.ts
+++ b/packages/backend/src/models/schema/emoji.ts
@@ -27,6 +27,7 @@ export const packedEmojiSchema = {
 		host: {
 			type: 'string',
 			optional: false, nullable: true,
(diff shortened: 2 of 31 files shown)


syuilo:
Thanks!
It has not yet been decided whether the documentation for the individual APIs will be written in TypeScript or managed separately on Misskey Hub (i think the latter is better), but it makes sense to be able to specify at the schema level that you can choose among multiple parameters to specify. I will merge after a short review.

Johann150:
> It has not yet been decided whether the documentation for the individual APIs will be written in TypeScript or managed separately on Misskey Hub.

I think because they are automatically generated and the schemas have to be written for validation anyway, this kind of documentation is easier to create. But the one kind does not rule out the other.

syuilo:
Generating documentation from TypeScript would make i18n difficult (which may not be necessary) and would require building a live instance to view the documentation.
Anyway, this PR is useful and will be merged.

syuilo:
Thanks!

PR #8461, opened 2022-04-02
fix: validation (better #8456)

Johann150:
- fix mistake from #8456: pure renotes are also allowed via `notes/create`
- fix pure renote checks in `notes/create`


syuilo:
🙏🏻

PR #8462, opened 2022-04-02
Fix: Adjust ESLint calls to properly interpret globs

### LINKED ISSUE #8002: PRでのlintが機能していない
明らかにLint違反のコード(キーワード間にスペースが無いとか)でもなぜかエラーになっていない
同じコードをローカルでlintするとちゃんと怒られる

pixeldesu:
# What

This PR does some adjustments to how ESLint is called:
* Renames the backend `.eslintrc.js` to `.eslintrc.cjs`, otherwise calling `yarn run lint` would error out
* Adds quotation marks `'` around the path globs for the `lint` tasks, so those are properly interpreted (fixes #8002)
* Makes the lint workflow run again with updating it to Node 16, matching the minimum required version for Misskey

# Why

> Renames the backend `.eslintrc.js` to `.eslintrc.cjs`, otherwise calling `yarn run lint` would error out

This is pretty simple and self-explanatory

> Adds quotation marks `'` around the path globs for the `lint` tasks, so those are properly interpreted

By default npm tasks are run with `sh` and not wrapping globs with quotation marks (passing them onto the tools itself) might cause the globs to be interpreted by the shell itself, and not by node. 

`sh`'s interpreted `src/**/*.ts` is different to node-glob `src/**/*.ts`

**Example:**

backend `eslint --quiet src/**/*.ts`

```
...

✖ 7 problems (7 errors, 0 warnings)
```

backend `eslint --quiet 'src/**/*.ts'`

```
...

✖ 46 problems (46 errors, 0 warnings)
  7 errors and 0 warnings potentially fixable with the `--fix` option.
```

I also made this change for the client package for consistency, but here the advanced glob usage probably caused it to be interpreted correctly already, the error count stayed the same before/after (935 errors)


syuilo:
Thanks!
Maybe related to https://github.com/misskey-dev/misskey/issues/8002 ?

pixeldesu:
> Thanks! Maybe related to #8002 ?

Oh yeah, that looks like it might be related to this.

tamaina:
Looks like we need to upgrade the node.js version to 16.x (the checks shows successful...)

pixeldesu:
@tamaina not entirely sure if that will fix the issue (as the workflow complains about not finding `eslint` at all) but I've done the adjustment to Node 16 in the workflow, so maybe that fixes it!

syuilo:
Good catch 🙏🙏🙏

syuilo:
私の環境だと
```
> eslint --quiet 'src/**/*.ts'


Oops! Something went wrong! :(

ESLint: 8.12.0

No files matching the pattern "'src/**/*.ts'" were found.
Please check for typing mistakes in the pattern.
```
になってしまうわね

pixeldesu:
@syuilo does `eslint --quiet "src/**/*.ts"` work in your environment? Maybe we need to switch to that then, I think those will be supported everywhere :/

I opted for `'` first because it (generally) works, and we need to escape usages of `"` in the package.json, but if that is the option that works, then we should switch to that.

syuilo:
@pixeldesu 
When `"` was used instead of `'`, it was executed successfully.

tamaina:
On Node.js on Windows, npm scripts are executed on cmd.exe, which don't allows to use `'` instead of `"`.

...Please use WSL2.

PR #8472, opened 2022-04-05
Add `.yarn` to `.gitignore`

ThatOneCalculator:
# What

Adds `.yarn` to `.gitignore`

# Why

When building with `yarn` > 2.x, it creates the `.yarn` folder which should be ignored.

# Additional info (optional)

Without `.yarn` in `.gitignore`, git shows over 2k changes.
image

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 # Node.js
 node_modules
 report.*.json
+.yarn
 
 # Cypress
 cypress/screenshots


acid-chicken:
No, we are still using Yarn Classic, not Yarn berry.
They are not supported concurrently and should be migrated.

If you are interested in a discussion on migration, please see here: #5858

ThatOneCalculator:
I use yarn berry on my development machine, even if it isn't officially supported, it doesn't hurt to have it in the gitignore.

acid-chicken:
Even if you exclude the .yarn directory, there are other incompatible files, such as yarn.lock, that you should always be careful not to include in your commits anyway.
That is, even if they work with Yarn berry, you will want to write a .gitignore that ignores them in your environment, so they should be written there.

btw I welcome you to the discussion about Yarn berry if you'd like.

ThatOneCalculator:
I'm happy to join the discussion. I think for now specifically, having the `.yarn` dir excluded would be a good first step.

PR #8485, opened 2022-04-11
Add English translations for CHANGELOG

ThatOneCalculator:
# What

Adds English translations for CHANGELOG.md for versions v12.108.2 --> v12.110.0


Johann150:
I think it is better if we keep the English changelog in misskey-hub. See e.g. misskey-dev/misskey-hub#63.

PR #8521, opened 2022-04-20
make emoji stand out more on reaction button

### LINKED ISSUE #8520: some emoji are hard to see in own reactions
## 💡 Summary

Some emoji are hard to see when they are shown in a users own reaction to a note. This is especially problematic if the emoji is in a colour that is the same as the background.

## 🥰 Expected Behavior

Emojis are easy to see.

## 🤬 Actual Behavior

⛳ on a green background

🦕 on a green background

## 📌 Environment

Misskey version: 12.110.0

## 🔧 Possible Solutions
- Change theme colours
    ⚠️ Other emoji colours may accidentally be used.
- Add a bit of shadow to the emoji.
```
.reacted > .mk-emoji {
	filter: drop-shadow(0 0 3px var(--bg));
}
```
⛳ on a green background, with a slight shadow

🦕 on a green background, with a slight shadow

Johann150:
a slight shadow makes them easier to see

# What
Added a slight shadow to a persons own reaction emoji as described in the issue.

⛳ with a slight shadow

# Why
fix #8520

### DIFF
--- a/packages/client/src/components/reactions-viewer.reaction.vue
+++ b/packages/client/src/components/reactions-viewer.reaction.vue
@@ -144,6 +144,10 @@ export default defineComponent({
 		> span {
 			color: var(--fgOnAccent);
 		}
+
+		> .mk-emoji {
+			filter: drop-shadow(0 0 3px var(--bg));
+		}
 	}
 
 	> span {


ThatOneCalculator:
Works great on my instance, even with odd-size emojis
image
image

ThatOneCalculator:
Even works great on animated emojis too, btw. I think this is ready to merge


ezgif com-gif-maker(13)

ThatOneCalculator (approved):

packages/client/src/components/reactions-viewer.reaction.vue
```diff
@@ -144,6 +144,10 @@ export default defineComponent({
 		> span {
 			color: var(--fgOnAccent);
 		}
+
+		> .mk-emoji {
```
syuilo, line 148: It is better for parents not to know what classes their children have.

Johann150: Without that, the only possible solution would be to add shadow to all emojis. I'm not sure if that is a good idea.

syuilo:
Fixed by https://github.com/misskey-dev/misskey/commit/b811de53b6c6841ec0816f58ea3c759be6f94a1e

Johann150:
I think it might have been better to use `var(--bg)` for the drop shadow because black may be hard to see in some themes.

syuilo:
In some cases, the use of bg can adversely affect the appearance. In the first place, it is not that much of a problem since it is the case of reactions you have made that are hard to see (you can hover to see the details if necessary).

PR #8527, opened 2022-04-22
enhance: only render public notes in HTML template

Johann150:
# What
Only public notes are rendered in the HTML template. These notes were only rendered without content before, but still showed the author and time when they were created. This information is no longer shown in just the HTML template.

The rendering of notes in Vue will be unaffected, so this does not change the rendering of notes for normal users.

# Why
related to #8317


syuilo:
🙏🏻

PR #8528, opened 2022-04-22
chore(lint): fix type definitions for jsrsasign

Johann150:
# What
Fix the type definitions for `jsrsasign`.

The `@types/jsrsasign` package is not available in exactly the same version as the jsrsasign package misskey uses, so i used an earlier patch version of the same package.

# Why
Fix the backend lint errors. Note that the new lint errors are all for the frontend.


packages/backend/package.json
```diff
@@ -173,6 +173,7 @@
 		"@types/web-push": "3.3.2",
 		"@types/websocket": "1.0.5",
 		"@types/ws": "8.5.3",
+		"@types/jsrsasign": "8.0.12",
```
acid-chicken: they should be sorted

Johann150: has been fixed

syuilo:
👍🏻

PR #8530, opened 2022-04-23
Refactor: Split package lints into separate jobs

pixeldesu:
# What

This PR splits the GitHub Action workflow "Lint" into the two separate jobs for backend and frontend linting.

# Why

This is here to avoid further issues with Lint issues eventually going unnoticed.

The case for this is the frontend lints that will never be run if the backend lints fail (as per comment by @Johann150 here: https://github.com/misskey-dev/misskey/issues/5543#issuecomment-1106665920).

# Additional info (optional)

Implementation detail, or why I'm using `yarn --cwd` instead of the `working-directory` property in GitHub Actions:

Some of the packages require shared dependencies, so I can't only run `yarn install` in `./packages/{client,backend}`. I tested around with using `cd` in between, but the working directory changes per run, so using `yarn --cwd` is the cleanest option.

_Also, I'm not linking the above issue because translating it didn't yield me with a clear answer if this PR would fix what they are talking about_

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -8,7 +8,7 @@ on:
   pull_request:
 
 jobs:
-  lint:
+  backend:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


syuilo:
Thanks!

PR #8534, opened 2022-04-23
fix(activitypub): add authorization checks

### LINKED ISSUE #6127: 自分宛てではないダイレクト投稿にリアクションを付けられる
## 💡 Summary

自分がリアクションをつけた他人の投稿に対し、通知欄に自分宛てではないダイレクト投稿が表示されてしまう。そこからリアクションを付与することが可能。

## 🙂 Expected Behavior

- 自分宛てではないダイレクト投稿が見えてしまうのを修正？
- リアクションを付けてしまえることの是非は検討？

## 📝 Steps to Reproduce

1.
2.
3.

## 📌 Environment

Firefox

### LINKED ISSUE #8529: some ActivityPub activities may be performed by unauthorized actors
## 💡 Summary

Some activities may be performed by actors which are not authorized to carry them out. For example, Misskey handles federated reactions to notes which the actor should not be able to see.

A review of similar cases would be necessary.

ℹ️ This should be hard to achieve in a normal scenario because the note IDs are partly randomly generated and should thus not be easy to guess. But sometimes Misskey leaks these IDs, a fixed example is #8448.

## 🥰 Expected Behavior

When actor C sends an activity to react to a direct note sent from actor A to actor B, Misskey does not perform the activity.

## 🤬 Actual Behavior

Actor C can send a reaction and Misskey will perform the activity.

## 📝 Steps to Reproduce

1. Send direct note with no recipient.
2. Send reaction to the ID of that note from another (custom/manipulated) server.
3. Reaction shows up on secret note.

Johann150:
# What
- Fixed spelling of `formerType`.
- Add missing authorization checks when performing ActivityPub actions, so the following are **rejected**:
    - pure renote deletion if actor !== renoter
        - ℹ️ This was already impossible before because the user ID is checked [later in `deleteNote`](https://github.com/misskey-dev/misskey/blob/f23d5a75f4cb64328a99e25f8a17150fe5b8a4d3/packages/backend/src/services/note/delete.ts#L81-L84) and is not a security issue. But since this is checked last, it may still mess up renote counts etc. so it is better to check it earlier too.
    - renotes where the renoter should not be able to see the original note
    - reactions where the reacter should not be able to see the original note

# Why
fix #8529
fix #6127

### DIFF
--- a/packages/backend/src/remote/activitypub/kernel/announce/note.ts
+++ b/packages/backend/src/remote/activitypub/kernel/announce/note.ts
@@ -9,6 +9,7 @@ import { fetchMeta } from '@/misc/fetch-meta.js';
 import { getApLock } from '@/misc/app-lock.js';
 import { parseAudience } from '../../audience.js';
 import { StatusError } from '@/misc/fetch.js';
+import { Notes } from '@/models/index.js';
 
 const logger = apLogger;
 
@@ -51,6 +52,8 @@ export default async function(resolver: Resolver, actor: CacheableRemoteUser, ac
 			}
 			throw e;
 		}
+
+		if (!await Notes.isVisibleForMe(renote, actor)) return 'skip: invalid actor for this activity';
 
 		logger.info(`Creating the (Re)Note: ${uri}`);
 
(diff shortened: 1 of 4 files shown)


syuilo:
👍🏻

PR #8541, opened 2022-04-24
refactor: use composition API for reports page

### LINKED ISSUE #8479: Can't look at All or Resolved Reports
## 💡 Summary

I can't see All or Resolved reports, only Unresolved. Because selecting another option won't do anything.

## 🥰 Expected Behavior

If I select another option from the "State" context menu of the Reports page the page should update and show the other kinds of reports.

## 🤬 Actual Behavior

The "State" context menu doesn't do anything if I select other options so I am stuck at only seeing Unresolved reports.

## 📝 Steps to Reproduce

1. Go to Control Panel > Reports
2. Click on the "State" context menu and select something else
3. Expect for the page to show you the kinds of reports you selected

## 📌 Environment

Misskey version: v12.109.2
Your OS: Arch Linux
Your browser: Firefox (but same issue seems to apply to any browser)

Johann150:
# What
Use composition API for `admin/abuses` page.

# Why
- refactor to composition API (per contribution guide)
- fix #8479

### DIFF
--- a/packages/client/src/pages/admin/abuses.vue
+++ b/packages/client/src/pages/admin/abuses.vue
@@ -24,10 +24,10 @@
 			</div>
 			<!-- TODO
 			<div class="inputs" style="display: flex; padding-top: 1.2em;">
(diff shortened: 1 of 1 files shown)


syuilo:
👍🏻

PR #8542, opened 2022-04-24
use assets submodule instead of discord twemoji

Johann150:
# What
Adjust paths for using Twemoji from assets repository instead of `@discord/twemoji`.

# Why
- part of #8505
- counterpart of https://github.com/misskey-dev/assets/pull/2
- `@discord/twemoji` also contains the PNG versions of twemoji leading to unnecessary files. When including twemoji as separate assets, these unnecessary files can be left out.

# Additional info
Waiting for https://github.com/misskey-dev/assets/pull/2 to be merged first, because the git submodule has to be updated then.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -15,7 +15,6 @@
 	},
 	"dependencies": {
 		"@bull-board/koa": "3.10.4",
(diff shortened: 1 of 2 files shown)


Johann150:
obsoleted by #8699

PR #8548, opened 2022-04-26
Fix lint issues in autocomplete component

pixeldesu:
# What

This PR doesn't change any logic, but it gets rid of about 14 lint errors in client.

# Why

# Additional info (optional)

While looking at the file I noticed a lot of variable names that are even worse than `e`. These are `x` and `y` used for filtering/mapping. These should be considered getting added to the denylist as well so we can factor them out later.

### DIFF
--- a/packages/client/src/components/autocomplete.vue
+++ b/packages/client/src/components/autocomplete.vue
@@ -131,8 +131,8 @@ const props = defineProps<{
 }>();
 
 const emit = defineEmits<{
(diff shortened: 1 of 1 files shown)


Johann150 (approved):

ThatOneCalculator (approved):

syuilo:
Personally, I don't see any problem with short names within functions such as map and filter, since the contents of the variable are clear.

syuilo:
🙏

PR #8549, opened 2022-04-26
enhance(MFM): add parameters to rotate

### LINKED ISSUE #8523: (MFM) rotate関数のspin関数相当への機能拡張
## Summary

spin関数で指定できるx, y等のプロパティをrotate関数でも使いたい

Johann150:
# What
The `x` and `y` parameters are added to the `$[rotate ]` MFM function rendering, analogous to the `$[spin ]` function.

# Why
fix #8523

### DIFF
--- a/packages/client/src/components/mfm.ts
+++ b/packages/client/src/components/mfm.ts
@@ -178,8 +178,12 @@ export default defineComponent({
 							return h(MkSparkle, {}, genEl(token.children));
 						}
 						case 'rotate': {
+							const rotateDir =
+								token.props.args.x ? 'X' :
+								token.props.args.y ? 'Y' :
+								'';
 							const degrees = parseInt(token.props.args.deg) || '90';
-							style = `transform: rotate(${degrees}deg); transform-origin: center center;`;
+							style = `transform: rotate${rotateDir}(${degrees}deg); transform-origin: center center;`;
 							break;
 						}
 					}
--- a/packages/client/src/components/mfm.ts
+++ b/packages/client/src/components/mfm.ts
@@ -178,12 +178,12 @@ export default defineComponent({
 							return h(MkSparkle, {}, genEl(token.children));
 						}
 						case 'rotate': {
-							const rotateDir =
-								token.props.args.x ? 'X' :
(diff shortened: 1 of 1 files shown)


ThatOneCalculator:
Doing `$[rotate.x` or `$[rotate.y` just causes the thing being rotated to dissapear

Regular:
image

X:
image

Y:
image

Johann150:
Yes thats because the default angle is 90°. Since the letters do not have any depth and you are looking at them from the side, that is not suprising. Try changing the angle.

ThatOneCalculator:
> Yes thats because the default angle is 90°. Since the letters do not have any depth and you are looking at them from the side, that is not suprising. Try changing the angle.

I can't seem to figure out the syntax, as `$[rotate.x=70` doesn't work

Johann150:
The angle must be changed as before by the `deg` parameter, e.g. `$[rotate.x,deg=45 a]`.

ThatOneCalculator:
[sick](https://stop.voring.me/notes/8zk82nn9lc)

ThatOneCalculator:
My conclusion on this: 

Does it look great? Not really. 
Is it *bad*? No. 
Could it possibly be used in some cool #mfmart? Maybe. 
Should it be added? Yea, more MFM options imo is never really a bad thing

futchitwo:
Note on compatibility with forking: In Groundpolis, rotateX and rotateY are preceded by `perspective(128px)`. (But probably syuilo does not care about compatibility with forking.)
https://github.com/Groundpolis/Groundpolis/blob/4f9c4532f51604056b2999212f8ff8323f59c357/src/client/components/mfm.functions.ts#L118

PS: This comes from $[spin]

Johann150:
I think it would be a good idea to add here too:

without perspective it can cause strange empty space
image

with `perspective(128px)` it looks better in my opinion
image

ThatOneCalculator:
Very tiny bug(?): the start of text gets cut off in chat bubbles
image

ThatOneCalculator (approved):

ThatOneCalculator:
:(

PR #8554, opened 2022-04-27
ci(docker): push arm64 images

### LINKED ISSUE #8555: ARM64 Support for Docker Hub Image
## Summary
High demand but building on GitHub Action is too slow.

jaredallard:
Resolve  #8555

# What

This PR adds multi-arch image building to the Github actions workflow docker release. This also publishes to GHCR, while keeping the current docker hub. I recognize that GHCR is technically an unnecessary change, so I'm happy to remove it if it's contentious. Also included is caching via Github actions cache API. This should marginally speed up docker image builds.

# Why

I'd like to be able to run misskey in Docker on ARM64 devices and provide the ability to easily add more platforms in the future.

# Additional info (optional)

I've tested this on my own fork, the results can be viewed here:

* Docker Image: https://hub.docker.com/r/jaredallard/misskey/tags
* GHCR Image: https://github.com/jaredallard/misskey/pkgs/container/misskey
* Github Actions Run: https://github.com/jaredallard/misskey/runs/6186949480?check_suite_focus=true

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -5,19 +5,33 @@ on:
     types: [published]
   workflow_dispatch:
 
+env:
+  DOCKER_IMAGE: misskey/misskey
+  GHCR_IMAGE: ghcr.io/${{ github.repository }}
+
 jobs:
   push_to_registry:
-    name: Push Docker image to Docker Hub
+    name: Push Docker Image
     runs-on: ubuntu-latest
 
     steps:
       - name: Check out the repo
         uses: actions/checkout@v2
+
+      # Setup QEMU to support arm64 image building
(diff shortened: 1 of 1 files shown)


tamaina:
Duplicated with #8279

ビルド時間が長いのは変わらないかぁ

jaredallard:
Oops, my bad :D

jaredallard:
And, yeah, the Github Action builds are going to be slow because of ARM64 emulation. It's also possible to use `docker manifest` and stitch together a cross-platform manifest. We'd need a Github Actions ARM64 runner to do that, though.

Or, the alternative is to use a self-hosted runner which will also be faster, but of course costs 💵

PR #8561, opened 2022-04-28
feat: option to collapse long notes

### LINKED ISSUE #8559: option to collapse long notes
## Summary

I think it would be nice if there was an option to collapse a long note again after you expanded it by clicking "show more".  A "show less" option.
This would make scrolling past long posts (and large mfm animations) a bit more manageable.

yitsushi:
# What
Add an option to re-collapse long messages.

# Why
Opening long messages on the timeline can make it hard to scroll back and forth, that mostly results in a "meh, I'll just refresh the page"

# Additional info (optional)
Please give me a string value for "Show less" in Japanese, Got a tip to use "閉じる", if it's not the right one, let me know. As I understand, the new keyword will get into the flow if it has a Japanese value, the rest come from Crowdin (that's why I did not commit any other languages/translations).

Closes #8559


yitsushi:
added en-US manually while testing, did not commit that file as it will be overwritten by Crowdin.

image

It has no effect on other variations (short message, has cw)

image

packages/client/src/components/note.vue
```diff
@@ -452,6 +456,21 @@ function readPromo() {
 				}
 
 				> .content {
+					&.isLong {
+						> .showLess {
+							width: 100%;
+							margin-top: 1em;
```
Johann150, line 462: I think it would be even better if you can make the button for collapsing sticky. I.e. it does not go out of view as long as you can see the note. This would be helpful because you can immediately collapse something again without having to scroll all the way to the bottom.

I think something like this should work:
```suggestion
							margin-top: 1em;
							position: sticky;
							bottom: 1em;
```
But then the button gets hard to see when its over the content, so maybe also increase the shadow.

yitsushi:
It was a bit hard to find something that makes it well visible on both light and dark themes.
image
image

packages/client/src/components/note.vue
yitsushi: With that sticky addition, I love it even more 😆

packages/client/src/components/note.vue
```diff
@@ -452,6 +456,24 @@ function readPromo() {
...
+							bottom: 1em;
+
+							> span {
+								display: inline-block;
+								background: var(--panel);
+								padding: 6px 10px;
+								font-size: 0.8em;
+								border-radius: 999px;
+								box-shadow: 0 2px 6px rgb(0 0 0 / 40%);
+								border: 1px solid var(--renote);
```
Johann150: I think you copied this from the "show more" button but that has a different background so the design needs to be a bit different here. I think you can use the CSS variable for background color here instead of adding a border. In my opinion the border does not look that great and can be replaced by making use of the [`spread-radius`](https://developer.mozilla.org/en-US/docs/Web/CSS/box-shadow#spread-radius).

I tried around a bit and I think this looks better:
```suggestion
								box-shadow: 0 0 7px 7px var(--bg);
```
image

What do you think?

yitsushi: Yes with that background color it's much more readable without border. (I have close to zero talent to UI design)

on light themes it's still a bit hard to read, but not painfully hard. At least for me. Tbf, the whole UI is harder to follow in light theme for me.
image

But on dark, it's awesome.
image

Johann150 (approved):

syuilo:
👍

Johann150:
Why was it reverted?

syuilo:
There was a problem with URL previews and images being cropped.
I am considering how to handle this issue.

Johann150:
I tried with images, with link previews, and with both but I did not notice any problems. 😕

PR #8563, opened 2022-04-28
chore(deps): Update github actions to use the same version as defined in .node-version

yitsushi:
# What
Update GitHub Actions to use the same version as defined in the .node-version file.

# Why

The [.node_version file](https://github.com/misskey-dev/misskey/blob/develop/.node-version) has `v18.x`, and with [this PR](https://github.com/misskey-dev/misskey/pull/8388) it will complain if the version is outdated. The CI can use `v17.x`, but it makes more sense to just use the same.

# Additional info (optional)
Right now all CI jobs are failing. I don't know if it fixes it, but potentially it can (will not fix lint that's for sure). I remember I had to update once from v16.x while updated my server, but I don't remember which version (and I couldn't find mention in the release notes)

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -16,7 +16,7 @@ jobs:
         submodules: true
     - uses: actions/setup-node@v3
       with:
-        node-version: 16.x
+        node-version: 18.x
         cache: 'yarn'
         cache-dependency-path: |
           packages/backend/yarn.lock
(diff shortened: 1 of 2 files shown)


yitsushi:
Nope, tests have the same error 🤷‍♀️

ThatOneCalculator:
Afaik tests have been broken for a while

Johann150 (approved):

syuilo:
👍

PR #8565, opened 2022-04-28
Refactor scratchpad to use Composition API

pixeldesu:
# What

Use Composition API for `scratchpad` page

# Why


syuilo:
🙏

PR #8566, opened 2022-04-28
Refactor api-console to use Composition API

pixeldesu:
# What

Use Composition API for `api-console` page

# Why


syuilo:
🙏

PR #8568, opened 2022-04-29
Refactor account-info to use Composition API

pixeldesu:
# What

Use Composition API for `settings/account-info` page

# Why

### DIFF
--- a/packages/client/src/pages/settings/account-info.vue
+++ b/packages/client/src/pages/settings/account-info.vue
@@ -127,7 +127,7 @@
 </template>
 
 <script lang="ts" setup>
-import { defineExpose, ref } from 'vue';
+import { defineExpose, onMounted, ref } from 'vue';
 import FormSection from '@/components/form/section.vue';
 import MkKeyValue from '@/components/key-value.vue';
 import * as os from '@/os';
@@ -139,10 +139,12 @@ import { i18n } from '@/i18n';
 
 const stats = ref<any>({});
 
-os.api('users/stats', {
-	userId: $i?.id
-}).then(response => {
-	stats.value = response;
+onMounted(() => {
+	os.api('users/stats', {
+		userId: $i?.id
+	}).then(response => {
+		stats.value = response;
+	});
 });
 
 defineExpose({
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/settings/account-info.vue
```diff
@@ -7,163 +7,150 @@
...
-	mounted() {
-		os.api('users/stats', {
-			userId: this.$i.id
-		}).then(stats => {
-			this.stats = stats;
-		});
-	},
+onMounted(() => {
+	os.api('users/stats', {
+		userId: $i?.id
```
Johann150: I think using `?.` here is incorrect as this would make a wrong API request with `{userId: null}`*. Instead you can use `pleaseLogin` to ensure there is user data. Note that the call is outside of `onMounted` because the user data is also used in the template.

```typescript
pleaseLogin();

onMounted(() => {
	os.api('users/stats', {
		userId: $i!.id
```

You would need to import the function of course:
```typescript
import { pleaseLogin } from '@/scripts/please-login';
```

I'm not sure if this was built into `defineComponent` before so this might strictly speaking not be your fault, but it would be nice if you can fix it anyway.
- - -
\* in actuality it will not make any API request because it crashes with `$i is null` when rendering the template

Johann150 (commented):
Looks good in general, just one small nitpick:

packages/client/src/pages/settings/account-info.vue
pixeldesu: It wouldn't change the behaviour, as the page doesn't work if you are not logged in. Not before and not now either. There was no login check present before as well. 

And in order to prevent massive duplication of login checks, there should only be one in `settings/index` which loads all the other pages. (like I've done in #8378)

ex. https://misskey.io/settings/account-info also throws a `u.$i is null`.

Johann150: I understand, but even then i think `!.` would fit with that idea better.

pixeldesu: Yeah, that makes sense, I can adjust it to that!

Johann150 (approved):

syuilo:
👍

PR #8572, opened 2022-04-29
Refactor delete-account to use Composition API

pixeldesu:
# What

Use Composition API for `settings/delete-account` page

# Why


packages/client/src/pages/settings/delete-account.vue
```diff
@@ -1,64 +1,52 @@
...
-				password: password
-			});
+	await os.alert({
+		title: i18n.ts._accountDelete.started,
+	});
 
-			await os.alert({
-				title: this.$ts._accountDelete.started,
-			});
+	signout();
```
Johann150: previous version was wrong here too, but still
```suggestion
	await signout();
```

Johann150 (approved):

syuilo:
🙏

PR #8576, opened 2022-04-30
Refactor settings/email to use Composition API

pixeldesu:
# What

Use Composition API for `settings/email` page

# Why


Johann150 (approved):
Non-null assertions are fine because that is also the previous behaviour. See also #8378.

syuilo:
🙏

PR #8577, opened 2022-04-30
chore: synchronize code and database schema

### LINKED ISSUE #8556: foreign key on abuse_user_report.targetUserId missing
## 💡 Summary
The foreign key `FK_d049123c413e68ca52abe734203` was removed in migration `1603094348345-refine-abuse-user-report.js`. This was a foreign key constraint on the column that is now `targetUserId`. This causes an internal error when a user that was reported has already been deleted.

This is not an unlikely scenario as a moderator/admin of the remote instance may delete the user because of other reports of the same offensive content.

## 🥰 Expected Behavior

When a user gets deleted, all reports that report them get deleted as well (`ON DELETE CASCADE`) and other reports can be viewed.

## 🤬 Actual Behavior

`ERR 1 [api] Internal error occurred in admin/abuse-user-reports: Could not find any entity of type "User" matching:`...

## 📝 Steps to Reproduce

1. Local user reports remote account.
2. Remote account gets deleted.
3. Try to view abuse reports

## 📌 Environment

Misskey version: all since 12.49.0

Johann150:
# What
Synchronize the database schema and the existing code so that running the command to generate new migrations mentioned in the contribution guide (`npx typeorm migration:generate -d ormconfig.js -o <name>`) reports that there are no differences in the database schema:

- `default: null` has been removed in several places. I found that when `nullable: true` existed in the codebase, `default: null` was often not used so I removed it where it existed. This should lead to a more uniform code style.
- `DEFAULT NULL` has been removed from the database schema because it is  unnecessary, as some comments in the migration say. When generating migrations with typeorm, it will therefore remove these default null values. To satisfy typeorm they have been removed.
- for auth sessions, I fixed a type definition according to the respective `JoinColumn`.
- instead of adding a Postgresql index for `storedInternal`, I removed the `@Index()` annotation because I could not find a case where this column is used in a query so an index seems to be unnecessary.
- some other minor alterations to correctly represent the database schema in the entities code

# Why
Make the migration generation command usable again. (also fixes #8556)

### DIFF
--- a/packages/backend/migration/1651224615271-foreign-key.js
+++ b/packages/backend/migration/1651224615271-foreign-key.js
@@ -1,0 +1,89 @@
+export class foreignKeyReports1651224615271 {
+    name = 'foreignKeyReports1651224615271'
+
+    async up(queryRunner) {
+        await Promise.all([
+            queryRunner.query(`ALTER INDEX "public"."IDX_seoignmeoprigmkpodgrjmkpormg" RENAME TO "IDX_c8cc87bd0f2f4487d17c651fbf"`),
+            queryRunner.query(`DROP INDEX "public"."IDX_note_on_channelId_and_id_desc"`),
+
+            // remove unnecessary default null, see also down
+            queryRunner.query(`ALTER TABLE "user" ALTER COLUMN "followersUri" DROP DEFAULT`),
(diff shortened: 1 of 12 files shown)


syuilo:
👍👍👍

PR #8578, opened 2022-04-30
Refactor settings/general to use Composition API

pixeldesu:
# What

Use Composition API for `settings/general` page

# Why

### DIFF
--- a/packages/client/src/pages/settings/general.vue
+++ b/packages/client/src/pages/settings/general.vue
@@ -1,10 +1,10 @@
 <template>
 <div class="_formRoot">
 	<FormSelect v-model="lang" class="_formBlock">
-		<template #label>{{ $ts.uiLanguage }}</template>
+		<template #label>{{ i18n.ts.uiLanguage }}</template>
 		<option v-for="x in langs" :key="x[0]" :value="x[0]">{{ x[1] }}</option>
 		<template #caption>
-			<I18n :src="$ts.i18nInfo" tag="span">
+			<I18n :src="i18n.ts.i18nInfo" tag="span">
 				<template #link>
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/settings/general.vue
```diff
@@ -102,122 +102,87 @@ import FormLink from '@/components/form/link.vue';
...
+	fontSize,
+	useSystemFont,
+	enableInfiniteScroll,
+	squareAvatars,
+	aiChanMode,
+	showGapBetweenNotesInTimeline,
+	instanceTicker,
+	overridedDeviceKind
+], () => {
+	reloadAsk();
```
Johann150: Since you refactored this a bit, the function is only called here, so maybe you want to merge these two as well? Otherwise, this is `async reloadAsk` so you would have to `await` here.

Johann150 (changes requested):

packages/client/src/pages/settings/general.vue
pixeldesu: Now, you would think it is only called here, but in fact, we have a case of code duplication (I found while checking out the next files to refactor)

https://github.com/misskey-dev/misskey/blob/develop/packages/client/src/pages/settings/menu.vue#L95-L104 has exactly the same code.

My course of action for this PR would be to just `async/await` here to fix this up and after refactoring menu, I would create a `reloadAsk` script instead, that can be referenced in both.

Johann150: 🤔 It might make sense to add that to the same module that `unisonReload` is from. But yeah, sure.

Johann150 (approved):

syuilo:
👍

PR #8579, opened 2022-04-30
Refactor import-export to use Composition API

pixeldesu:
# What

Use Composition API for `settings/import-export` page

# Why


Johann150 (approved):
I think there is a lot of code duplication here that could be removed. Not sure if that should be done in this PR. For example: `exportNotes`, `exportBlocking`, `exportUserLists` and `exportMuting` could be done with one function that takes an argument.

Apart from that the changes made here look good to me.

syuilo:
🙏

PR #8580, opened 2022-04-30
Refactor instance-mute to use Composition API

pixeldesu:
# What

Use Composition API for `settings/instance-mute` page

# Why

### DIFF
--- a/packages/client/src/pages/settings/instance-mute.vue
+++ b/packages/client/src/pages/settings/instance-mute.vue
@@ -1,67 +1,53 @@
 <template>
 <div class="_formRoot">
-	<MkInfo>{{ $ts._instanceMute.title }}</MkInfo>
+	<MkInfo>{{ i18n.ts._instanceMute.title }}</MkInfo>
 	<FormTextarea v-model="instanceMutes" class="_formBlock">
-		<template #label>{{ $ts._instanceMute.heading }}</template>
-		<template #caption>{{ $ts._instanceMute.instanceMuteDescription }}<br>{{ $ts._instanceMute.instanceMuteDescription2 }}</template>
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/settings/instance-mute.vue
```diff
@@ -1,67 +1,53 @@
...
-			// Refresh filtered list to signal to the user how they've been saved
-			this.instanceMutes = mutes.join('\n');
-		},
+import { $i } from '@/account';
+import { i18n } from '@/i18n';
+
+const instanceMutes = ref('');
+const changed = ref(false);
+
+instanceMutes.value = $i!.mutedInstances.join('\n');
```
Johann150: I think setting up `instanceMutes` could be merged into one.
```suggestion
const instanceMutes = ref($i!.mutedInstances.join('\n'));
const changed = ref(false);
```

pixeldesu: Yeah, that makes a lot of sense actually!

Johann150 (approved):

packages/client/src/pages/settings/instance-mute.vue
```diff
@@ -1,67 +1,51 @@
...
+async function save() {
+	let mutes = instanceMutes.value
+		.trim().split('\n')
+		.map(el => el.trim())
+		.filter(el => el);
+	
+	await os.api('i/update', {
+		mutedInstances: mutes,
+	});
+
```
Johann150: git complains about whitespace
```suggestion

	await os.api('i/update', {
		mutedInstances: mutes,
	});

```

syuilo:
🙏

PR #8583, opened 2022-05-01
fix: Add rel attribute to host-meta

### LINKED ISSUE #8582: No rel attribute on <Link /> in /.well-known/host-meta
I don't know much about the fediverse protocols, and I can't find a spec for whatever is supposed to be happening here, so forgive me if this isn't your bug.  I've been trying to figure out why GNU Social won't talk to Misskey, and I think it's because it looks for a link with `rel="lrdd"` in `/.well-known/host-meta`.  When talking to Mastodon, it finds a link to webfinger there.  Misskey has a link to webfinger in that file, but it doesn't have a `rel` attribute at all, and GNU Social gives up.

mei23:
# What
Add rel attribute to host-meta

# Why
Fix #8582

# Additional info (optional)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ You should also include the user name that made the change.
 - Client: fix profile tabs @futchitwo
 - Server: await promises when following or unfollowing users @Johann150
 - Client: fix abuse reports page to be able to show all reports @Johann150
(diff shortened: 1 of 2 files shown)


acid-chicken (approved):

syuilo:
🙏🏻

PR #8586, opened 2022-05-01
Refactor settings/menu to use Composition API

pixeldesu:
# What

Use Composition API for `settings/menu` page

# Why

# Additional info (optional)

The `reloadAsk()` method present here is exactly the same as in `settings/general`, once both this PR and #8578 have been merged, I will split out the method into it's own script that can be referenced in both places.

### DIFF
--- a/packages/client/src/pages/settings/menu.vue
+++ b/packages/client/src/pages/settings/menu.vue
@@ -50,9 +50,9 @@ async function addItem() {
 		title: i18n.ts.addItem,
 		items: [...menu.map(k => ({
 			value: k, text: i18n.ts[menuDef[k].title]
-		})), ...[{
+		})), {
 			value: '-', text: i18n.ts.divider
-		}]]
+		}]
 	});
 	if (canceled) return;
 	items.value = [...split.value, item].join('\n');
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/settings/menu.vue
```diff
@@ -27,81 +27,60 @@ import { menuDef } from '@/menu';
...
+}
 
-		items() {
-			this.save();
-		},
-	},
+function save() {
+	defaultStore.set('menu', split.value);
+	reloadAsk();
+}
```
Johann150: you already `await` save later, but it is not defined as async

```suggestion
async function save() {
	defaultStore.set('menu', split.value);
	await reloadAsk();
}
```

packages/client/src/pages/settings/menu.vue
```diff
@@ -27,81 +27,60 @@ import { menuDef } from '@/menu';
...
-		},
+async function addItem() {
+	const menu = Object.keys(menuDef).filter(k => !defaultStore.state.menu.includes(k));
+	const { canceled, result: item } = await os.select({
+		title: i18n.ts.addItem,
+		items: [...menu.map(k => ({
+			value: k, text: i18n.ts[menuDef[k].title]
+		})), ...[{
+			value: '-', text: i18n.ts.divider
+		}]]
```
Johann150:
```suggestion
		})), {
			value: '-', text: i18n.ts.divider
		}]
```

Johann150 (changes requested):

Johann150 (approved):

syuilo:
🙏

PR #8587, opened 2022-05-01
Refactor settings/notifications to use Composition API

pixeldesu:
# What

Use Composition API for `settings/notifications` page

# Why


syuilo:
🙏

PR #8590, opened 2022-05-01
Refactor settings/plugin to use Composition API

pixeldesu:
# What

Use Composition API for `settings/plugin` page

# Why

### DIFF
--- a/packages/client/src/pages/settings/plugin.vue
+++ b/packages/client/src/pages/settings/plugin.vue
@@ -1,38 +1,38 @@
 <template>
 <div class="_formRoot">
-	<FormLink to="/settings/plugin/install"><template #icon><i class="fas fa-download"></i></template>{{ $ts._plugin.install }}</FormLink>
+	<FormLink to="/settings/plugin/install"><template #icon><i class="fas fa-download"></i></template>{{ i18n.ts._plugin.install }}</FormLink>
 
 	<FormSection>
-		<template #label>{{ $ts.manage }}</template>
+		<template #label>{{ i18n.ts.manage }}</template>
 		<div v-for="plugin in plugins" :key="plugin.id" class="_formBlock _panel" style="padding: 20px;">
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/settings/plugin.vue
```diff
@@ -41,67 +41,54 @@ import MkKeyValue from '@/components/key-value.vue';
...
+	const { canceled, result } = await os.form(plugin.name, config);
+	if (canceled) return;
+
+	const coldPlugins = ColdDeviceStorage.get('plugins');
+	coldPlugins.find(p => p.id === plugin.id)!.configData = result;
+	ColdDeviceStorage.set('plugins', coldPlugins);
+
+	nextTick(() => {
+		location.reload();
+	});
```
Johann150, line 73: Shouldn't this use `unisonReload` (or something like `askReload`)?

packages/client/src/pages/settings/plugin.vue
```diff
@@ -41,67 +41,54 @@ import MkKeyValue from '@/components/key-value.vue';
...
+}
+
+function changeActive(plugin, active) {
+	const coldPlugins = ColdDeviceStorage.get('plugins');
+	coldPlugins.find(p => p.id === plugin.id)!.active = active;
+	ColdDeviceStorage.set('plugins', coldPlugins);
+
+	nextTick(() => {
+		location.reload();
+	});
```
Johann150, line 83: Also here: use `unisonReload`?

Johann150 (approved):

packages/client/src/pages/settings/plugin.vue
pixeldesu: yeah, both `location.reload()` should probably be replaced with `unisonReload()`

Especially since it is actually used in the `uninstall` method on the same page already

syuilo:
🙏

PR #8593, opened 2022-05-01
Fix missing import of defineAsyncComponent in os.ts

pixeldesu:
# What

Adds a missing import for `defineAsyncComponent` in `os.ts`

# Why

This missing import breaks a lot of the functionality from `os.*` methods.

# Additional info (optional)


tamaina:
🙏🙏🙏

PR #8604, opened 2022-05-02
Refactor settings/accounts to use Composition API

pixeldesu:
# What

Use Composition API for `settings/accounts` page

# Why

# Additional info (optional)

Switching accounts and listing them works fine, deleting an account from the list does not however. To my knowledge, checking the git history and asking Misskey users, this never worked before. Just as a consideration for the review.


syuilo:
🙏

PR #8613, opened 2022-05-05
Fix lint issues in Drive components

pixeldesu:
# What

This PR fixes lint issues in the Drive component and all the sub components that make it up, reducing the total lint error count by 93 errors (as of opening this PR from 870 -> 777)

# Why

# Additional info (optional)

This PR includes a lot of files, but I decided to group it all together because the `drive*` components are used all together most of the time, so review of the functionality should stay mostly in one place.

### DIFF
--- a/packages/client/src/components/drive.folder.vue
+++ b/packages/client/src/components/drive.folder.vue
@@ -118,7 +118,7 @@ function onDrop(ev: DragEvent) {
 
 	//#region ドライブのファイル
 	const driveFile = ev.dataTransfer.getData(_DATA_TRANSFER_DRIVE_FILE_);
-	if (driveFile !== null && driveFile !== '') {
+	if (driveFile != null && driveFile !== '') {
 		const file = JSON.parse(driveFile);
 		emit('removeFile', file.id);
 		os.api('drive/files/update', {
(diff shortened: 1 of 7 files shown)


packages/client/src/components/drive.folder.vue
```diff
@@ -118,7 +118,7 @@ function onDrop(ev: DragEvent) {
 
 	//#region ドライブのファイル
 	const driveFile = ev.dataTransfer.getData(_DATA_TRANSFER_DRIVE_FILE_);
-	if (driveFile != null && driveFile != '') {
+	if (driveFile !== null && driveFile !== '') {
```
syuilo: As for comparisons with null, `!=` and `!==` treat `undefined` differently, so Misskey uses `!=` and `==` for null comparisons.
https://github.com/misskey-dev/misskey/blob/9230334a319c93d1604576778ed39c9de9d510ce/packages/client/.eslintrc.js#L21

pixeldesu: 👍 I'll readjust it for the null checks then, I probably accidentally did this in multiple places. (because of the other reported issues in the same lines usually)

pixeldesu: This should be fixed now! Readjusted all null checks I was able to find and lint error count didn't increase back up again 😃

syuilo:
🙏

PR #8619, opened 2022-05-06
Fix lint issues in post form component

pixeldesu:
# What

This PR fixes lint issues in the post form component (as of opening this PR from 770 -> 710)

# Why

# Additional info (optional)

There are still two more lint issues in post-form. One is a complaint because of `data` being used inside data structures (for drafts) and the other one is an actually properly defined default value being requested to be a function, which seems wrong?

### DIFF
--- a/packages/client/src/components/post-form.vue
+++ b/packages/client/src/components/post-form.vue
@@ -244,11 +244,10 @@ if (props.reply && props.reply.text != null) {
 												`@${x.username}@${toASCII(otherHost)}`;
 
 		// 自分は除外
-		if ($i.username === x.username && x.host == null) continue;
-		if ($i.username === x.username && x.host === host) continue;
+		if ($i.username === x.username && (x.host == null || x.host === host)) continue;
 
 		// 重複は除外
-		if (text.indexOf(`${mention} `) !== -1) continue;
+		if (text.includes(`${mention} `)) continue;
 
 		text += `${mention} `;
 	}
--- a/packages/client/src/components/post-form.vue
+++ b/packages/client/src/components/post-form.vue
@@ -228,7 +228,7 @@ if (props.mention) {
 	text += ' ';
 }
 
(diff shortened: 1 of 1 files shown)


packages/client/src/components/post-form.vue
```diff
@@ -239,16 +239,16 @@ if (props.reply && props.reply.text != null) {
...
 
 		// 自分は除外
-		if ($i.username == x.username && x.host == null) continue;
-		if ($i.username == x.username && x.host == host) continue;
+		if ($i.username === x.username && x.host == null) continue;
+		if ($i.username === x.username && x.host === host) continue;
 
 		// 重複は除外
-		if (text.indexOf(`${mention} `) != -1) continue;
+		if (text.indexOf(`${mention} `) !== -1) continue;
```
Johann150: I think this can be simplified?
```suggestion
		if (text.includes(`${mention} `)) continue;
```

packages/client/src/components/post-form.vue
```diff
@@ -239,16 +239,16 @@ if (props.reply && props.reply.text != null) {
...
-											(otherHost == null || otherHost == host) ?
+											(otherHost == null || otherHost === host) ?
 												`@${x.username}` :
 												`@${x.username}@${toASCII(otherHost)}`;
 
 		// 自分は除外
-		if ($i.username == x.username && x.host == null) continue;
-		if ($i.username == x.username && x.host == host) continue;
+		if ($i.username === x.username && x.host == null) continue;
+		if ($i.username === x.username && x.host === host) continue;
```
Johann150: a duplicate check of username can be avoided here
```suggestion
		if ($i.username === x.username && (x.host == null || x.host === host)) continue;
```

Johann150 (approved):
The lint fixing is good. I think while we are at it, there are some small adjustments that can be done as well:

syuilo:
🙏

PR #8620, opened 2022-05-06
Fix lint issues in emoji picker components

pixeldesu:
# What

This PR fixes lint issues in the emoji picker and its subcomponents (as of opening this PR from 770 -> 755)

# Why

### DIFF
--- a/packages/client/src/components/emoji-picker.section.vue
+++ b/packages/client/src/components/emoji-picker.section.vue
@@ -24,7 +24,7 @@ const props = defineProps<{
 }>();
 
 const emit = defineEmits<{
-	(event: 'chosen', v: string, ev: MouseEvent): void;
+	(ev: 'chosen', v: string, event: MouseEvent): void;
 }>();
 
 const shown = ref(!!props.initialShown);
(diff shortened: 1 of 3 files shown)


packages/client/src/components/emoji-picker.section.vue
```diff
@@ -24,7 +24,7 @@ const props = defineProps<{
 }>();
 
 const emit = defineEmits<{
-	(e: 'chosen', v: string, ev: MouseEvent): void;
+	(event: 'chosen', v: string, ev: MouseEvent): void;
```
Johann150: I think consistent naming would be nice.
```suggestion
	(ev: 'chosen', v: string, ev: MouseEvent): void;
```

pixeldesu: True, I was debating with myself how to approach this the best way since `ev` is taken as an argument name here already 😵‍💫

pixeldesu: Decided to fix this by switching it around, using `ev` as the event name (as everywhere) and `event` for the event class type

Johann150 (approved):

syuilo:
🙏

PR #8621, opened 2022-05-06
Fix lint issues in scripts

pixeldesu:
# What

This PR fixes lint issues in the scripts directory (as of opening this PR from 770 -> 696)

# Why


packages/client/src/scripts/url.ts
```diff
@@ -4,7 +4,7 @@ export function query(obj: {}): string {
 		.reduce((a, [k, v]) => (a[k] = v, a), {} as Record<string, any>);
 
 	return Object.entries(params)
-		.map((e) => `${e[0]}=${encodeURIComponent(e[1])}`)
+		.map((p) => `${p[0]}=${encodeURIComponent(p[1])}`)
```
Johann150, line 7: I'm not sure if this is much of an improvement.

Johann150 (approved):
This also seems good to me.

packages/client/src/scripts/url.ts
syuilo: `e` is confusing because it is often used as an abbreviation for `error`.

syuilo:
🙏

PR #8622, opened 2022-05-06
Refactor Chart component

pixeldesu:
# What

This PR refactors the `Chart` component moving it over to the Composition API and also fixing a lot of the lint issues present.

# Why

Move to Composition API, as per contribution guide, and also lowering of lint issues (91 in this file alone).

# Additional info (optional)

I fixed all lint issues that I was able to. The main problem with the `Chart` component is the usage of `data` as an attribute by Chart.js itself, as this is required for its use. That's why I opted to disable `id-denylist` for the script tag in `Chart.vue`, including a comment in the file itself explaning why.

### DIFF
--- a/packages/client/src/components/chart.vue
+++ b/packages/client/src/components/chart.vue
@@ -7,8 +7,14 @@
 </div>
 </template>
 
-<script lang="ts">
-import { defineComponent, onMounted, ref, watch, PropType, onUnmounted, shallowRef } from 'vue';
+<script lang="ts" setup>
+/* eslint-disable id-denylist --
+  Chart.js has a `data` attribute in most chart definitions, which triggers the
+  id-denylist violation when setting it. This is causing about 60+ lint issues.
+  
+  As this is part of Chart.js's API it makes sense to disable the check here.
+*/
+import { defineExpose, defineProps, onMounted, ref, watch, PropType, onUnmounted } from 'vue';
 import {
 	Chart,
 	ArcElement,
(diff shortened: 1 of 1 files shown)


packages/client/src/components/chart.vue
```diff
@@ -7,8 +7,14 @@
 </div>
 </template>
 
-<script lang="ts">
-import { defineComponent, onMounted, ref, watch, PropType, onUnmounted, shallowRef } from 'vue';
+<script lang="ts" setup>
+/* eslint-disable id-denylist --
+  Chart.js has a `data` attribute in most chart definitions, which triggers the
+  id-denylist violation when setting it. This is causing about 60+ lint issues.
+
```
Johann150: trailing whitespace
```suggestion

```

packages/client/src/components/chart.vue
```diff
@@ -80,826 +126,783 @@ const getColor = (i) => {
...
+			case 'users': return fetchUsersChart(false);
+			case 'users-total': return fetchUsersChart(true);
+			case 'active-users': return fetchActiveUsersChart();
+			case 'notes': return fetchNotesChart('combined');
+			case 'local-notes': return fetchNotesChart('local');
+			case 'remote-notes': return fetchNotesChart('remote');
+			case 'notes-total': return fetchNotesTotalChart();
+			case 'drive': return fetchDriveChart();
+			case 'drive-files': return fetchDriveFilesChart();
+
```
Johann150: trailing whitespace
```suggestion

```

packages/client/src/components/chart.vue
```diff
@@ -7,8 +7,14 @@
 </div>
 </template>
 
-<script lang="ts">
-import { defineComponent, onMounted, ref, watch, PropType, onUnmounted, shallowRef } from 'vue';
+<script lang="ts" setup>
+/* eslint-disable id-denylist --
+  Chart.js has a `data` attribute in most chart definitions, which triggers the
+  id-denylist violation when setting it. This is causing about 60+ lint issues.
```
Johann150, line 13: I understand what you are doing here, but would it not be better to limit the area covered by this a bit more? The issue with `data` does not occur until line 80 or so, and that is only in the type definition. Not sure if eslint picks that up.

pixeldesu: Well, the initial type information thing is not picked up by `id-denylist`, but everytime an assignment to an object property `data` is happening this is triggered, causing the many errors.

The first actual lint coming up is on line 215 probably. So I get your point, but a lot of the early code in the script tag is just component setup and imports in general, which is why I opted to put the comment right at the top so I don't have a huge comment + explainer somewhere in the middle of it.

I actually wanted to only remove `data` from `id-denylist` for this file, but the ESLint comment to adjust the rule didn't really work and only fully disabling with with the `eslint-disable` comment had an effect. :/

packages/client/src/components/chart.vue
```diff
@@ -80,826 +125,782 @@ const getColor = (i) => {
...
+});
+
+onUnmounted(() => {
+	if (disposeTooltipComponent) disposeTooltipComponent();
+});
+
+defineExpose({
+	chartEl,
+	fetching,
 });
```
syuilo: Do we need to export these values?

pixeldesu: I checked and didn't find any references to the exposed values, so I removed the exposing alltogether!

syuilo:
🙏🏻

PR #8630, opened 2022-05-08
Refactor pleaseLogin to show a sign-in dialog

pixeldesu:
# What

Refactored `pleaseLogin()` to show a sign-in dialog instead of just an alert message prompting to login.

This PR is a successor to #8378 

<details>
<summary>Screenshot/Recording</summary>

</details>

# Why

This improves the user experience greatly, and makes the script more reusable for cases like blocking navigation to pages.

# Additional info (optional)

This Pull Request also refactors the `MkSignin` component to the Composition API and adds a `message` property to show a (informational) message to the user.


packages/client/src/scripts/please-login.ts
```diff
@@ -1,14 +1,21 @@
...
-	});
+	popup(defineAsyncComponent(() => import('@/components/signin-dialog.vue')), {
+		autoSet: true,
+		message: i18n.ts.signinRequired
+	}, {
+		cancelled: () => {
+			if (path) {
+				window.location.href = path;
+			}
+		},
```
Johann150, line 17: Maybe this should be more general but I think redirecting should usually make sense.

packages/client/src/components/signin.vue
```diff
@@ -45,190 +48,185 @@
 </form>
 </template>
 
-<script lang="ts">
-import { defineAsyncComponent, defineComponent } from 'vue';
+<script lang="ts" setup>
+import { defineAsyncComponent, defineEmits, defineProps } from 'vue';
```
Johann150: > `defineProps` and `defineEmits` are compiler macros only usable inside `<script setup>`. They do not need to be imported, and are compiled away when `<script setup>` is processed.

https://vuejs.org/api/sfc-script-setup.html#defineprops-defineemits
```suggestion
import { defineAsyncComponent } from 'vue';
```

packages/client/src/components/signin.vue
```diff
@@ -45,190 +48,185 @@
...
-					this.onLogin(res);
-				}).catch(this.loginFailed);
-			}
-		},
+function onLogin(res) {
+	if (props.autoSet) {
+		return login(res.i);
+	} else {
+		return;
+	}
```
Johann150: explicit return is unnecessary
```suggestion
	}
```

packages/client/src/components/signin.vue
```diff
@@ -45,190 +48,185 @@
...
+		}
+	}).catch(() => {
+		queryingKey = false;
+		return Promise.reject(null);
+	}).then(credential => {
+		queryingKey = false;
+		signing = true;
+		return os.api('signin', {
+			username: username,
+			password: password,
```
Johann150:
```suggestion
			username,
			password,
```

packages/client/src/components/signin.vue
```diff
@@ -45,190 +48,185 @@
...
-			this.signing = false;
-		},
+function onSubmit() {
+	signing = true;
+	console.log('submit')
+	if (!totpLogin && user && user.twoFactorEnabled) {
+		if (window.PublicKeyCredential && user.securityKeys) {
+			os.api('signin', {
+				username: username,
+				password: password
```
Johann150:
```suggestion
				username,
				password,
```

packages/client/src/components/signin.vue
```diff
@@ -45,190 +48,185 @@
...
+				return queryKey();
+			}).catch(loginFailed);
+		} else {
+			totpLogin = true;
+			signing = false;
+		}
+	} else {
+		os.api('signin', {
+			username: username,
+			password: password,
```
Johann150:
```suggestion
			username,
			password,
```

packages/client/src/components/signin.vue
```diff
@@ -45,190 +48,185 @@
...
+			token: user && user.twoFactorEnabled ? token : undefined
+		}).then(res => {
+			console.log('successfully signed in')
```
Johann150:
```suggestion
```
I think you forgot to remove this from debugging.

Johann150 (approved):

packages/client/src/scripts/please-login.ts
pixeldesu: well, we could turn it into a callback-based approach, but for all that's worth I think there only ever are the two follow-up cases of either not doing anything (like for the note-related actions) or redirecting somewhere else (as a route-guard).

I think it's fine if we leave it as-is for now, extending it when we find other use-cases shouldn't be too much of an issue.

locales/ja-JP.yml
```diff
@@ -425,7 +425,7 @@ quoteQuestion: "引用として添付しますか？"
 noMessagesYet: "まだチャットはありません"
 newMessageExists: "新しいメッセージがあります"
 onlyOneFileCanBeAttached: "メッセージに添付できるファイルはひとつです"
-signinRequired: "ログインしてください"
+signinRequired: "続行する前に、サインアップまたはサインインが必要です"
```
robflop, line 428: Using the proper japanese word for registering and login instead of sign-in here seems better to me.
Rephrased to "Please register or login to continue", as that is closer to the previous as well.

```suggestion
signinRequired: "進むために、登録またはログインしてください"
```

Would nonetheless wait for input from a maintainer on that phrasing.

syuilo:
👍🏻

packages/client/src/components/signin.vue
```diff
@@ -47,199 +50,192 @@
...
-					text: this.$ts.signinFailed
-				});
-				this.signing = false;
-			});
-		},
+function onUsernameChange() {
+	os.api('users/show', {
+		username: username
+	}).then(user => {
+		user = user;
```
Johann150, line 110: Overlooked this mistake in reviewing, fixed in f90c947036dc6d9c0b7e4ea7cfa1dc53a0b98c15.

PR #8634, opened 2022-05-09
enhance: page image component with alt text

Johann150:
# What
- refactor page image component to composition API
- image uses the existing image component to display the image, so blurhash and alt text work

# Why
succeeds #6601


syuilo:
👍🏻

PR #8641, opened 2022-05-10
fix: hide metadata of private notes

### LINKED ISSUE #8317: Messages that aren't supposed to be accesible are still (kind of) visible

Johann150:
# What
- [x] the `getNote` getter is expanded to also check visibility as this is a common task across all API endpoints
- [x] some uses of `Notes.find` or similar are replaced by uses of `getNote` so the visibility is checked properly
- [x] `Notes.pack` will return an error if the note is not visible to the specified user instead of only hiding metadata
- [x] `Notes.packMany` is modified to not throw an error, but only filters out notes that would have raised an error
- [x] all uses of `Notes.pack` properly handle the new error
- [x] adjust tests

# Why
fix #8317

# Additional Info
The `hideNote` function has been removed. This has the additional advantage that it removes duplicated code for checking the visibility of a note already in `isVisibleForMe`.

Users will no longer be able to see notes that are not visible to them. The previous behaviour was to show the note just with the text replaced by a message that the note is not visible to the user. The new behaviour is that the server claims that the note does not exist at all if it is not visible to the user.

Users will no longer be able to reply to or renote notes that are not visible to them.

### DIFF
--- a/packages/backend/src/misc/get-note-summary.ts
+++ b/packages/backend/src/misc/get-note-summary.ts
@@ -9,10 +9,6 @@ export const getNoteSummary = (note: Packed<'Note'>): string => {
 		return `(❌⛔)`;
 	}
 
-	if (note.isHidden) {
-		return `(⛔)`;
-	}
-
 	let summary = '';
 
 	// 本文
--- a/packages/backend/src/models/repositories/note-favorite.ts
+++ b/packages/backend/src/models/repositories/note-favorite.ts
@@ -24,6 +24,6 @@ export const NoteFavoriteRepository = db.getRepository(NoteFavorite).extend({
 		me: { id: User['id'] }
 	) {
 		return Promise.allSettled(favorites.map(x => this.pack(x, me)))
-		.then(promises => promises.flatMap(result => result.status === 'fulfilled' ? [result.value] : []);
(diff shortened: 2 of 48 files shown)


Johann150:
I discovered some issues with this, it needs more testing.

Johann150:
Rebased to remove the changes that were already merged in #8838.

Johann150:

PR #8657, opened 2022-05-12
Refactor emoji-edit-dialog to use Composition API

pixeldesu:
# What

Use Composition API for emoji edit dialog component in the admin settings

# Why

# Additional info (optional)

After submitting edit changes with the dialog I noticed that nothing happened. `MkPagination` lost the `replaceItem` and `removeItem` methods while it was refactored to the Composition API. Was this intentional?

To "quick-fix" this problem I mitigated the issue with just opting to `reload()` the entire pagination instead. If we should bring back the missing methods tell me so and I'll add them back to the component!

### DIFF
--- a/packages/client/src/pages/admin/emoji-edit-dialog.vue
+++ b/packages/client/src/pages/admin/emoji-edit-dialog.vue
@@ -35,6 +35,7 @@ import MkInput from '@/components/form/input.vue';
 import * as os from '@/os';
 import { unique } from '@/scripts/array';
 import { i18n } from '@/i18n';
+import { emojiCategories } from '@/instance';
 
 const props = defineProps<{
 	emoji: any,
@@ -44,11 +45,7 @@ let dialog = $ref(null);
 let name: string = $ref(props.emoji.name);
 let category: string = $ref(props.emoji.category);
 let aliases: string = $ref(props.emoji.aliases.join(' '));
-let categories: string[] = $ref([]);
-
-os.api('meta', { detail: false }).then(({ emojis }) => {
-	categories = unique(emojis.map((x: any) => x.category || '').filter((x: string) => x !== ''));
-});
+let categories: unknown[] = $ref(emojiCategories);
 
 const emit = defineEmits<{
 	(ev: 'done', v: { deleted?: boolean, updated?: any }): void,
--- a/packages/client/src/pages/admin/emoji-edit-dialog.vue
+++ b/packages/client/src/pages/admin/emoji-edit-dialog.vue
@@ -27,85 +27,75 @@
 </XModalWindow>
 </template>
 
-<script lang="ts">
-import { defineComponent } from 'vue';
+<script lang="ts" setup>
+import { } from 'vue';
 import XModalWindow from '@/components/ui/modal-window.vue';
 import MkButton from '@/components/ui/button.vue';
 import MkInput from '@/components/form/input.vue';
 import * as os from '@/os';
 import { unique } from '@/scripts/array';
+import { i18n } from '@/i18n';
 
-export default defineComponent({
-	components: {
-		XModalWindow,
-		MkButton,
-		MkInput,
-	},
-
-	props: {
-		emoji: {
-			required: true,
-		}
-	},
-
-	emits: ['done', 'closed'],
-
-	data() {
-		return {
-			name: this.emoji.name,
-			category: this.emoji.category,
-			aliases: this.emoji.aliases?.join(' '),
-			categories: [],
-		}
-	},
-
-	created() {
-		os.api('meta', { detail: false }).then(({ emojis }) => {
-			this.categories = unique(emojis.map((x: any) => x.category || '').filter((x: string) => x !== ''));
-		});
-	},
-
-	methods: {
-		ok() {
-			this.update();
-		},
-
-		async update() {
-			await os.apiWithDialog('admin/emoji/update', {
-				id: this.emoji.id,
-				name: this.name,
-				category: this.category,
-				aliases: this.aliases.split(' '),
-			});
-
-			this.$emit('done', {
-				updated: {
-					name: this.name,
-					category: this.category,
-					aliases: this.aliases.split(' '),
-				}
-			});
-			this.$refs.dialog.close();
-		},
-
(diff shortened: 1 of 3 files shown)


packages/client/src/pages/admin/emoji-edit-dialog.vue
```diff
@@ -27,85 +27,75 @@
...
-	components: {
-		XModalWindow,
-		MkButton,
-		MkInput,
-	},
+const props = defineProps({
+	emoji: {
+		required: true,
+	}
+});
```
Johann150: The codebase usually uses the TypeScript way of `defineProps`:
```suggestion
const props = defineProps<{
	emoji: any,
}>();
```
I'm not quite sure what type `emoji` is so I put `any`. If you know the more specific type that would be appreciated.

packages/client/src/pages/admin/emoji-edit-dialog.vue
```diff
@@ -27,85 +27,75 @@
...
-	props: {
-		emoji: {
-			required: true,
-		}
-	},
+let dialog = $ref(null);
+let name: string = $ref(props.emoji.name);
+let category: string = $ref(props.emoji.category);
+let aliases: string = $ref(props.emoji.aliases?.join(' '));
+let categories: string[] = $ref([]);
```
Johann150:
```suggestion
let categories: string[] = $ref($instance.emojiCategories);
```

packages/client/src/pages/admin/emoji-edit-dialog.vue
```diff
@@ -27,85 +27,75 @@
...
 
-	emits: ['done', 'closed'],
+os.api('meta', { detail: false }).then(({ emojis }) => {
+	categories = unique(emojis.map((x: any) => x.category || '').filter((x: string) => x !== ''));
+});
```
Johann150: Using the above suggestion this API call can be removed.

packages/client/src/pages/admin/emoji-edit-dialog.vue
```diff
@@ -27,85 +27,75 @@
...
 
-	methods: {
-		ok() {
-			this.update();
-		},
+async function update() {
+	await os.apiWithDialog('admin/emoji/update', {
+		id: props.emoji.id,
+		name: name,
+		category: category,
```
Johann150:
```suggestion
		name,
		category,
```

packages/client/src/pages/admin/emoji-edit-dialog.vue
```diff
@@ -27,85 +27,75 @@
...
-			await os.apiWithDialog('admin/emoji/update', {
-				id: this.emoji.id,
-				name: this.name,
-				category: this.category,
-				aliases: this.aliases.split(' '),
-			});
+	emit('done', {
+		updated: {
+			name: name,
+			category: category,
```
Johann150:
```suggestion
			name,
			category,
```

packages/client/src/pages/admin/emoji-edit-dialog.vue
```diff
@@ -27,85 +27,75 @@
...
 
-	props: {
-		emoji: {
-			required: true,
-		}
-	},
+let dialog = $ref(null);
+let name: string = $ref(props.emoji.name);
+let category: string = $ref(props.emoji.category);
+let aliases: string = $ref(props.emoji.aliases?.join(' '));
```
Johann150: According to the [type definition in the Misskey SDK](https://github.com/misskey-dev/misskey.js/blob/c17c42fa05bcbcc38c7a4395d1bb3d2336d3d402/src/entities.ts#L252), `aliases` can not be null.
```suggestion
let aliases: string = $ref(props.emoji.aliases.join(' '));
```

packages/client/src/pages/admin/emojis.vue
```diff
@@ -134,13 +134,8 @@ const edit = (emoji) => {
...
 		done: result => {
-			if (result.updated) {
-				emojisPaginationComponent.value.replaceItem(item => item.id === emoji.id, {
-					...emoji,
-					...result.updated
-				});
-			} else if (result.deleted) {
-				emojisPaginationComponent.value.removeItem(item => item.id === emoji.id);
+			if (result.updated || result.deleted) {
+				emojisPaginationComponent.value.reload();
```
Johann150: The pagination component still has `updateItem` which is exported so I think at least that could be used here.

@syuilo I think it would be more performant to not reload the pagination component when a custom emoji gets deleted, but I am not sure why you removed the `removeItem` functionality here: https://github.com/misskey-dev/misskey/commit/586c11251a8c0e7ca9f8f3bbaad9bf745e6ef948#diff-412470ddca46a77339849fd70b56ce462cef9d5682a9186b505d5c6499f74b78

Johann150 (changes requested):

packages/client/src/pages/admin/emojis.vue
syuilo: I don't see any reason to remove it, so it would be better to re-implement `removeItem`.

pixeldesu: I reimplemented `removeItem` and adjusted the code for the emoji page to use that now! 👍

packages/client/src/pages/admin/emoji-edit-dialog.vue
```diff
@@ -27,85 +27,71 @@
...
-			required: true,
-		}
-	},
-
-	emits: ['done', 'closed'],
+let dialog = $ref(null);
+let name: string = $ref(props.emoji.name);
+let category: string = $ref(props.emoji.category);
+let aliases: string = $ref(props.emoji.aliases.join(' '));
+let categories: unknown[] = $ref(emojiCategories);
```
Johann150: [The `category` property of emoji is type `string`.](https://github.com/misskey-dev/misskey.js/blob/c17c42fa05bcbcc38c7a4395d1bb3d2336d3d402/src/entities.ts#L251)

```suggestion
let categories: string[] = $ref(emojiCategories);
```

Johann150 (approved):

syuilo:
🙏🏻

PR #8659, opened 2022-05-13
fix: keep file order

### LINKED ISSUE #8647: 添付した画像の順番がドライブに上げた順になる
## 💡 Summary

画像を下記のように添付しても https://misskey.io/notes/905gpqcna5 になってしまう

image

## 🥰 Expected Behavior

ドライブで選択した順番になる（以前の挙動）

## 🤬 Actual Behavior

ドライブに上げた順になる（fileIdの降順で表示される？）

## 📝 Steps to Reproduce

1.画像を上げる
2.ドライブから選択する
3.順番が異なる

## 📌 Environment

Misskey version: 12.110.1
Your OS: Windows 10
Your browser:
- Vivaldi 5.2.2623.41 (Stable channel) (64-bit)
- Chrome 101.0.4951.54 (Official Build) （64 ビット） (cohort: Stable)
- FireFox 100.0 (64 ビット)

Johann150:
# What
The file IDs are used in the order they are sent.

When using the `IN` operator, the database will return the files in the order they are stored on disk, which is usually the order they were inserted in. That explains why attachments were always ordered by file ID as noted in the issue.

To keep the advantage of having a single database query, I modified the query to add an order by clause instead of partly reverting 9f7cdb4bc734e02be0acc03bfae5abe94ac7466c which would have meant multiple queries. This should keep any improved performance.

# Why
fix #8647

### DIFF
--- a/packages/backend/src/server/api/endpoints/notes/create.ts
+++ b/packages/backend/src/server/api/endpoints/notes/create.ts
@@ -172,10 +172,14 @@ export default define(meta, paramDef, async (ps, user) => {
 	let files: DriveFile[] = [];
 	const fileIds = ps.fileIds != null ? ps.fileIds : ps.mediaIds != null ? ps.mediaIds : null;
 	if (fileIds != null) {
-		files = await DriveFiles.findBy({
-			userId: user.id,
-			id: In(fileIds),
-		});
+		files = await DriveFiles.createQueryBuilder('file')
+			.where('file.userId = :userId AND file.id IN (:...fileIds)', {
+				userId: user.id,
+				fileIds,
+			})
+			.orderBy('array_position(ARRAY[:...fileIds], "id")')
+			.setParameters({ fileIds })
+			.getMany();
 	}
 
 	let renote: Note | null = null;


syuilo:
👍🏻👍🏻👍🏻

tamaina:
`function array_position(text[], character varying) does not exist` が出た

Johann150:
It worked for me 😬

Which version of postgres are you using?
Can you try if it works if you cast the id to the same type with `"id"::text`?

absturztau-be:
same happened to me. resulted into not being able to post notes with attachments and it slowed down drive too

psql version 13.5 for me

according to the psql docs, it should be part of it too

https://www.postgresql.org/docs/13/functions-array.html

> Can you try if it works if you cast the id to the same type with `"id"::text`?

haven't tested this yet

absturztau-be:
changing the query to ` array_position(ARRAY[<whatever>], "id"::text)` seems to work on psql 13.5

tamaina:
I use 13.7.

@Johann150 Which version are you using?


> Can you try if it works if you cast the id to the same type with "id"::text?

haven't tested this yet

Johann150:
Ah, my testing environment used Postgres 14.2, so I was using the wrong version of Postgres. My mistake 🙇

Johann150:
Thanks for catching this, has been fixed on develop.

PR #8660, opened 2022-05-13
perf: fix caching

Johann150:
# What
Fixes the cache implementation to actually store the result when fetching a value for a cache miss.

# Why
The cache implementation did previously not store the results of the computation and was thus not a cache at all. This can cause a significant number of database queries each time someone with a large number of followers does something that causes an activity to be federated.


syuilo:
Good catch

PR #8661, opened 2022-05-13
Refactor file-dialog to use Composition API

pixeldesu:
# What

Use Composition API for file dialog component in the admin settings

# Why

### DIFF
--- a/packages/client/src/pages/admin/file-dialog.vue
+++ b/packages/client/src/pages/admin/file-dialog.vue
@@ -48,11 +48,9 @@ let file: any = $ref(null);
 let info: any = $ref(null);
 let isSensitive: boolean = $ref(false);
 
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/admin/file-dialog.vue
```diff
@@ -34,74 +34,54 @@
...
-	},
-
-	created() {
-		this.fetch();
-	},
+const props = defineProps({
+	fileId: {
+		required: true,
+	}
+});
```
Johann150:
```suggestion
const props = defineProps<{
	fileId: string,
}>();
```

Johann150 (changes requested):

Johann150 (approved):

syuilo:
🙏

PR #8662, opened 2022-05-13
Refactor admin/index to use Composition API

pixeldesu:
# What

Use Composition API for `admin/index` page

# Why

# Additional info (optional)

This PR also fixes navigation to different initial pages, before always `overview` appeared, it now uses the same logic as in `settings/index` written by tamaina in https://github.com/misskey-dev/misskey/commit/eac71ae1d7f7d1ee4c06c4060979b7b292c0e57e

### DIFF
--- a/packages/client/src/pages/admin/index.vue
+++ b/packages/client/src/pages/admin/index.vue
@@ -1,6 +1,6 @@
 <template>
 <div ref="el" class="hiyeyicy" :class="{ wide: !narrow }">
-	<div v-if="!narrow || page == null" class="nav">
+	<div v-if="!narrow || initialPage == null" class="nav">
 		<MkHeader :info="header"></MkHeader>
 	
 		<MkSpacer :content-max="700" :margin-min="16">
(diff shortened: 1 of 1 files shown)


packages/client/src/pages/admin/index.vue
```diff
@@ -35,292 +35,279 @@ import { instance } from '@/instance';
...
 import * as os from '@/os';
 import { lookupUser } from '@/scripts/lookup-user';
+import { MisskeyNavigator } from '@/scripts/navigate';
 
-export default defineComponent({
-	components: {
-		MkSuperMenu,
-		MkInfo,
-	},
+const isEmpty = (x: any) => x == null || x === '';
```
Johann150: I think this is a type error, what about this:
```suggestion
const isEmpty = (x: string | null) => x == null || x === '';
```
Since you use it for `maintainerName` and `maintainerEmail` only, that matches the [type definition in the Misskey SDK](https://github.com/misskey-dev/misskey.js/blob/c17c42fa05bcbcc38c7a4395d1bb3d2336d3d402/src/entities.ts#L256-L257)

packages/client/src/pages/admin/index.vue
```diff
@@ -35,292 +35,279 @@ import { instance } from '@/instance';
...
-				childInfo.value = viewInfo;
-			}
-		};
-		const pageProps = ref({});
+const props = defineProps({
+	initialPage: {
+		type: String,
+		required: false
+	}
+});
```
Johann150: Again, compile time `defineProps` is usually used:
```suggestion
const props = defineProps<{
	initialPage?: string,
}>();
```

Johann150 (changes requested):

Johann150:
@tamaina Maybe you can also take a look because you similarly refactored the settings index page. 🙏

packages/client/src/pages/admin/index.vue
```diff
@@ -35,292 +35,276 @@ import { instance } from '@/instance';
...
-				case 'overview': return defineAsyncComponent(() => import('./overview.vue'));
-				case 'users': return defineAsyncComponent(() => import('./users.vue'));
-				case 'emojis': return defineAsyncComponent(() => import('./emojis.vue'));
-				case 'federation': return defineAsyncComponent(() => import('../federation.vue'));
-				case 'queue': return defineAsyncComponent(() => import('./queue.vue'));
-				case 'files': return defineAsyncComponent(() => import('./files.vue'));
-				case 'announcements': return defineAsyncComponent(() => import('./announcements.vue'));
-				case 'ads': return defineAsyncComponent(() => import('./ads.vue'));
-				case 'database': return defineAsyncComponent(() => import('./database.vue'));
-				case 'abuses': return defineAsyncComponent(() => import('./abuses.vue'));
```
Johann150, line 223: The `abuses` one got removed?

Johann150 (approved):
Looks fine to me now.

syuilo:
🙏

PR #8665, opened 2022-05-13
Refactor metrics component to Composition API

pixeldesu:
# What

Use Composition API for the metrics component on the admin dashboard

# Why

# Additional info (optional)

This component is commented out in the only place it is used. Is it actually required at all anymore?


Johann150:
@syuilo is this component still needed? It has been commented out since commit 4e4c559db6964cbf17fcadf38d55fc79c995ca42.

PR #8667, opened 2022-05-13
Refactor admin/other-settings to use Composition API

pixeldesu:
# What

Use Composition API for `admin/other-settings` page

# Why

# Additional info (optional)

This page is empty, it does nothing aside of pulling `admin/meta` and updating meta with nothing on Save press. This should actually be removed until "other" settings are actually needed.


ThatOneCalculator:
+1

syuilo:
🙏

PR #8671, opened 2022-05-14
Add additional background for acrylic popups if backdrop-filter is unsupported

### LINKED ISSUE #7996: reaction & renote tooltip breaks
## 💡 Summary

The detail list viewer for reactions / renotes breaks when many people react or renote:

image

Also the `._acrylic` background does not work in Firefox, making it hard to look at.

## 🙂 Expected Behavior

- [x] All items stay in the box.
- [ ] The box is not transparent.

## ☹️ Actual Behavior

Some items can spill out of the box. The box is transparent and you see text behind which makes text in the tooltip hard to read.

## 📌 Environment

Misskey 12.96.1

pixeldesu:
# What

This PR adds an additional background layer if the `backdrop-filter` is not supported.

This fixes  #7996

**Screenshot:**

Firefox 100.0 on Windows 11, no `backdrop-filter` support:
image

Microsoft Edge 101.0.1210.39 on Windows 11, `backdrop-filter` support
image

# Why

Reaction or chart popups are hardly readable on Firefox.

# Additional info (optional)

This implementation is somewhat of a "hack". The clean solution would be introducing an additional theme color variable that generates a less transparent background color variant, but I'm not sure what the effects of adding/changing theme colors would be for...all the created themes, basically.

So, for clarity, this implementation just adds an `:after` element via CSS *only* if `backdrop-filter` is not supported. It takes up the height/width of the element, places itself behind it and uses `var(--bg)` at 75% transparency.

### DIFF
--- a/packages/client/src/style.scss
+++ b/packages/client/src/style.scss
@@ -385,6 +385,20 @@ hr {
 	background: var(--acrylicPanel);
 	-webkit-backdrop-filter: var(--blur, blur(15px));
 	backdrop-filter: var(--blur, blur(15px));
+
+	@supports not (backdrop-filter: var(--blur, blur(15px))) {
+		&:after {
+			content: "";
+			position: absolute;
+			top: 0;
+			right: 0;
+			left: 0;
+			bottom: 0;
+			z-index: -1;
+			background: var(--bg);
+			opacity: .75;
+		}
+	}
 }
 
 ._formBlock {


Johann150 (approved):
👍

ThatOneCalculator (approved):
Works great on https://stop.voring.me with Firefox 101.0.1 on Linux

tamaina:
Closing because of the conflict.

PR #8672, opened 2022-05-14
Fix unexpected token breaking builds

pixeldesu:
# What

A typo while refactoring broke builds, this PR fixes it.

# Why

From a recent `Build` action step output:
```
/home/runner/work/misskey/misskey/packages/client/src/pages/admin/bot-protection.vue
57 |  const MkCaptcha = defineAsyncComponent(() => import('@/components/captcha.vue'));
58 |  
59 |  let provider: = $ref(null);
   |                ^
```


syuilo:
👍

PR #8680, opened 2022-05-14
Refactor my-antennas/edit to use Composition API

pixeldesu:
# What

Use Composition API for `my-antennas/edit` page and its subcomponents

# Why

### DIFF
--- a/packages/client/src/pages/my-antennas/edit.vue
+++ b/packages/client/src/pages/my-antennas/edit.vue
@@ -24,8 +24,8 @@ function onAntennaUpdated() {
 	nav.push('/my/antennas');
 }
 
-watch(() => props.antennaId, async () => {
-	antenna = await os.api('antennas/show', { antennaId: props.antennaId });
+os.api('antennas/show', { antennaId: props.antennaId }).then((antennaResponse) => {
+	antenna = antennaResponse;
 });
 
 defineExpose({
--- a/packages/client/src/pages/my-antennas/edit.vue
+++ b/packages/client/src/pages/my-antennas/edit.vue
@@ -4,49 +4,34 @@
 </div>
 </template>
 
-<script lang="ts">
-import { defineComponent } from 'vue';
-import MkButton from '@/components/ui/button.vue';
+<script lang="ts" setup>
+import { watch } from 'vue';
 import XAntenna from './editor.vue';
 import * as symbols from '@/symbols';
 import * as os from '@/os';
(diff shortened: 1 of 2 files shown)


packages/client/src/pages/my-antennas/edit.vue
```diff
@@ -4,49 +4,34 @@
...
+	antennaId: string
+}>();
+
+function onAntennaUpdated() {
+	nav.push('/my/antennas');
+}
+
+watch(() => props.antennaId, async () => {
+	antenna = await os.api('antennas/show', { antennaId: props.antennaId });
+});
```
Johann150: Does this really need a `watch`?

packages/client/src/pages/my-antennas/editor.vue
```diff
@@ -44,135 +44,110 @@
...
+			userListId,
+			userGroupId,
+			withReplies,
+			withFile,
+			notify,
+			caseSensitive,
+			users: users.trim().split('\n').map(x => x.trim()),
+			keywords: keywords.trim().split('\n').map(x => x.trim().split(' ')),
+			excludeKeywords: excludeKeywords.trim().split('\n').map(x => x.trim().split(' ')),
+		});
```
Johann150: This is a lot of code duplication.
```suggestion
	const data = {
		name,
		src,
		userListId,
		userGroupId,
		withReplies,
		withFile,
		notify,
		caseSensitive,
		users: users.trim().split('\n').map(x => x.trim()),
		keywords: keywords.trim().split('\n').map(x => x.trim().split(' ')),
		excludeKeywords: excludeKeywords.trim().split('\n').map(x => x.trim().split(' ')),
	};
	if (props.antenna.id == null) {
		await os.apiWithDialog('antennas/create', data);
		emit('created');
	} else {
		await os.apiWithDialog('antennas/update', data);
```

packages/client/src/pages/my-antennas/edit.vue
pixeldesu: It doesn't! Removed it.

packages/client/src/pages/my-antennas/editor.vue
pixeldesu: I didn't apply your patch right here, because I needed to make some adjustments:
* I didn't know you made your "I'll have to add more lints" threat true already (`data` is not allowed 😛)
* On updating `antennaId` is required, this was removed in your deduplication

Johann150 (approved):

syuilo:
🙏

PR #8681, opened 2022-05-15
Fix lint issues in Deck UI components

pixeldesu:
# What

This PR fixes lint issues in the Deck UI components (as of opening this PR from 518 -> 483)

# Why

# Additional info (optional)

I didn't fix all lint issues, the one that's still there across some columns and the deck store is `data` because that is used in the widget columns, and I assume that adjusting that would require some local migration strategy to not break users widget columns.


Johann150 (approved):

syuilo:
🙏🏻

PR #8687, opened 2022-05-15
chore: synchronize visibility checks

### LINKED ISSUE #8686: different visibility checks between different API endpoints
## 💡 Summary
Visibility checks are different between different API endpoints. This is caused by e.g. `generateVisibilityQuery` and `Notes.isVisibleForMe` having different behaviour.

## 🥰 Expected Behavior
When Note can be seen from one API endpoint it should be visible from other endpoint.

## 🤬 Actual Behavior
Note can be seen from `/api/notes/show` but not from `/api/notes/children`. This is very confusing because a note shows up in notifications but when going to the parent note, it is not visible.

## 📝 Steps to Reproduce

1. User A posts a note, mentioning user B.
2. User C replies to the note, also mentioning user B.
3. User B can see both notes in his notifications.
4. User B can not see the note from step 2 when he looks at the note from step 1.

## 📌 Environment

Misskey version: 12.110.1

Johann150:
# What
- small code quality improvements
- synchronize the code for checking note visibility: followers only notes are visible to all mentioned users

# Why
fix #8686

### DIFF
--- a/packages/backend/src/models/repositories/note.ts
+++ b/packages/backend/src/models/repositories/note.ts
@@ -136,6 +136,7 @@ async function populateMyReaction(note: Note, meId: User['id'], _hint_?: {
 
 export const NoteRepository = db.getRepository(Note).extend({
 	async isVisibleForMe(note: Note, meId: User['id'] | null): Promise<boolean> {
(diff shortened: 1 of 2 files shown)


syuilo:
🙏🏻

PR #8635, opened 2022-05-09
fix(federation): accept HTTP signatures with hs2019

Johann150:
# What
Use a different http-signature module that supports the currently recommended `hs2019` scheme. This is a modified version of the original library that supports `hs2019`, but fixed to interpret `hs2019` as `rsa-sha256` because that is Mastodon's implementation.

# Why
fix #8165
fix #8629
close #3996

Issue #8165, opened 2022-01-21
Help debugging why Misskey cannot follow an Owncast account


gabek:
Hello.  In testing ActivityPub Federation support between Owncast and Misskey I found I've been unable to follow an Owncast account from Misskey.  The follow request is handled and accepted by Owncast, but nothing happens on the Misskey side when the `Accept` is sent to it.

Could you point me in the right direction of what might be missing to fix this incompatibility?

## 💡 Summary

Unable to follow an Owncast server from Misskey.  The Follow `Accept` activity from Owncast that is sent from Owncast to Misskey is ignored and the state stays in `Processing...`.

## 🙂 Expected Behavior

`Processing...` should conclude and the user will be seen as Followed.

## ☹️ Actual Behavior

Misskey stays in `Processing...` status and the Owncast user is not successfully followed.

## 📝 Steps to Reproduce

1. Visit an Owncast profile with Misskey.
2. Press "Follow".
3. Owncast receives the Follow request and sends back an `Accept`.
4. Nothing further happens.

An example Owncast profile is: https://misskey.io/@demo@watch.owncast.online

The following `Accept` payload is what is sent to Misskey:

```
{"@context":"https://www.w3.org/ns/activitystreams","actor":"https://OWNCAST_SERVER/federation/user/testing","id":"https://OWNCAST_SERVER/federation/FoC_r6Jng","object":"https://misskey.io/2395e9f0-4e85-42d3-bb27-63180fb5b258","type":"Accept"}
```
Sent to https://misskey.io/users/8vr4n4kqug/inbox

gabek:
Hi there. Just checking in to see if anybody has any suggestions on what might be an issue here.  Other ActivityPub powered services can follow Owncast servers, but Misskey can't. So if there's something specific that I'm missing that I need to add so this can be fixed, I'd love to resolve it. I want all the Misskey users out there to be able to take advantage of following Owncast, too.

Amolith:
I don't know much about TypeScript or Misskey's codebase but I'd be more than happy to help debug things and provide admin-side logs; I run instances of both Owncast and Misskey.

mei23:
Understood. Does not support hs2019. 
> signature parse error: InvalidParamsError: hs2019 is not supported

I will consider it.

gabek:
Thank you for looking into it!  PeerTube recently made the change to support `hs2019` with [this fork](https://github.com/Chocobozzz/node-http-signature), so maybe you could drop that in as well.

[...]

Issue #8688, opened 2022-05-16
Resolver tries to fetch local ActivityPub content via HTTP


Johann150:
## 💡 Summary
The current resolver implementation always tries to fetch content from the web. This also applies to content generated by the server itself.
While this works, it is very inefficient because it has to go through the complete HTTP stack and takes up resources in the reverse proxy server.

## 📌 Environment
Misskey version: 12.110.1

mei23:
Wow, I didn't expect to be able to Resovle and process local objects with `activitypub/kernel`.
(I'm not an implementer, so I don't know the original intention.)

In many places, local uri has a process that does not call Resolver in the first place.

Johann150:
> In many places, local uri has a process that does not call Resolver in the first place.

I should have looked into that more in detail. But I think it would be nice if there is a way where you do not have to think about resolving a URL too much. And I think the most obvious solution would for `Resolver.resolve` to check if a URL is local. It could then make use of `DbResolver` functionality.

But there is a difference between `Resolver` and `DbResolver`: The former returns the ActivityStreams representation and the other returns the Misskey object representation. So some mapping might still be necessary.

Or alternatively, uses of `Resolver.resolve` should be checked if they can use `DbResolver` instead.

mei23:
If Resolver can use local uri, I think there is a part that will be smart.
However, both functions are different, so I think you will need both.

Johann150:
Maybe `DbResolver` should have functionality like the `api/ap/show` endpoint.

side note: That endpoint should probably be refactored to use `DbResolver.parseUri` or the new `Resolver.resolve` functionality I described above.

Johann150:
I wonder if we should also prefer checking the database to making a HTTPS request in the normal `Resolver`, even for non-local URLs.

PR #8689, opened 2022-05-16
fix: add id for activitypub follows

### LINKED ISSUE #8218: Follow Activityに対してstringでAccept/Rejectを受け取った場合に詰んでしまう
## 💡 Summary

Misskey sends
```yml
  id: https://local/:uuid ★random id
  type: Follow
  actor: https://local/users/follower
  object: https://remote/users/followee
```

Misskey receives... This is OK 
```yml
  type: Accept
  actor: https://remote/users/followee
  object:
    id: https://local/:uuid ★random id
    type: Follow
    actor: https://local/users/follower
    object: https://remote/users/followee
```

Suppose you receive this, This is valid. But unable to perform processing.
```yml
  type: Accept
  actor: https://remote/users/followee
  object: https://local/:uuid ★random id
```

#8165

## 📝 Steps to Reproduce

1.
2.
3.

Johann150:
# What
`Follow` activities are rendered with a proper `id` so that when servers only respond with an accept with the id, it can be processed correctly.

The respective activity is provided at the new URL.
# Why
fix #8218

# Additional Info
together with #8635 this allows to follow users on Owncast 0.0.11 correctly (#8165)

### DIFF
--- a/packages/backend/src/server/activitypub.ts
+++ b/packages/backend/src/server/activitypub.ts
@@ -231,13 +231,13 @@ router.get('/follows/:follower/:followee', async ctx => {
 	// check if the following exists.
 
 	const [follower, followee] = await Promise.all([
-		 Users.findOneBy({ id: ctx.params.follower }),
-		 Users.findOneBy({ id: ctx.params.followee }),
+		Users.findOneBy({ id: ctx.params.follower }),
+		Users.findOneBy({ id: ctx.params.followee }),
 	]);
 
 	if (follower == null || followee == null) {
-		 ctx.status = 404;
-		 return;
+		ctx.status = 404;
+		return;
 	}
 
 	ctx.body = renderActivity(renderFollow(follower, followee));
--- a/packages/backend/src/server/activitypub.ts
+++ b/packages/backend/src/server/activitypub.ts
@@ -231,8 +231,14 @@ router.get('/follows/:follower/:followee', async ctx => {
 	// check if the following exists.
 
 	const [follower, followee] = await Promise.all([
-		Users.findOneBy({ id: ctx.params.follower }),
-		Users.findOneBy({ id: ctx.params.followee }),
+		Users.findOneBy({
+			id: ctx.params.follower,
+			host: IsNull(),
+		}),
+		Users.findOneBy({
+			id: ctx.params.followee,
+			host: Not(IsNull()),
+		}),
 	]);
 
 	if (follower == null || followee == null) {
--- a/packages/backend/src/server/activitypub.ts
+++ b/packages/backend/src/server/activitypub.ts
@@ -15,7 +15,7 @@ import { inbox as processInbox } from '@/queue/index.js';
 import { isSelfHost } from '@/misc/convert-host.js';
 import { Notes, Users, Emojis, NoteReactions } from '@/models/index.js';
 import { ILocalUser, User } from '@/models/entities/user.js';
-import { In, IsNull } from 'typeorm';
+import { In, IsNull, Not } from 'typeorm';
 import { renderLike } from '@/remote/activitypub/renderer/like.js';
 import { getUserKeypair } from '@/misc/keypair-store.js';
 import renderFollow from '@/remote/activitypub/renderer/follow.js';
--- a/packages/backend/src/server/activitypub.ts
+++ b/packages/backend/src/server/activitypub.ts
@@ -18,6 +18,7 @@ import { ILocalUser, User } from '@/models/entities/user.js';
 import { In, IsNull } from 'typeorm';
 import { renderLike } from '@/remote/activitypub/renderer/like.js';
 import { getUserKeypair } from '@/misc/keypair-store.js';
+import renderFollow from '@/remote/activitypub/renderer/follow.js';
 
 // Init router
 const router = new Router();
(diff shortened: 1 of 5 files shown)


mei23:
> together with https://github.com/misskey-dev/misskey/pull/8635 this allows to follow users on Owncast 0.0.11 correctly (https://github.com/misskey-dev/misskey/issues/8165)

Need a parser for `https://example.com/follows/<follower id>/<followee id>` ?

Johann150:
I'm not sure where you mean this parser should be added? Misskey resolves everything via HTTP at the moment as far as I know (#8688), so that is also what I did for now: https://github.com/misskey-dev/misskey/pull/8689/files#diff-f6e4f1a116f4168da1998803a9ddb39d8f8a395aeced91cf67c4e42519807996R229

mei23:
I think need to interpret the Accept payload returned remotely.
https://github.com/misskey-dev/misskey/blob/fee2878b98253f2510ae2bc5c746c058a043ef82/packages/backend/src/remote/activitypub/kernel/accept/follow.ts

#8688
`Resolver.resolve()` only gets the JSON of the AP, not the information required for local DB processing.

Johann150:
The `Accept` payload will be turned into the AP object by this `Resolver.resolve`:
https://github.com/misskey-dev/misskey/blob/fee2878b98253f2510ae2bc5c746c058a043ef82/packages/backend/src/remote/activitypub/kernel/accept/index.ts#L16-L19
This will then fetch the AP JSON from the new route I added. Just as before, the AP JSON representation should contain all the necessary information. I'm not sure which information is missing.

mei23:
Ah, it may work. But not expecting to get a remote object via http.
I haven't seen it well, but maybe validation is needed.

Johann150:
Yes, I thought about validation, see my (edited) post in the original issue: https://github.com/misskey-dev/misskey/issues/8218#issuecomment-1127710881
> I first thought this could be a security vulnerability because remote ActivityPub servers could send crafted messages with these IDs, but since Misskey checks that a follow request exists when accepting, it should be fine.
> https://github.com/misskey-dev/misskey/blob/fee2878b98253f2510ae2bc5c746c058a043ef82/packages/backend/src/services/following/requests/accept.ts#L12-L19

mei23:
Okay, I was worried about Undo and Reject processing, but maybe it's okay.

packages/backend/src/server/activitypub.ts
```diff
@@ -224,4 +225,24 @@ router.get('/likes/:like', async ctx => {
...
+
+	if (follower == null || followee == null) {
+		ctx.status = 404;
+		return;
+	}
+
+	ctx.body = renderActivity(renderFollow(follower, followee));
+	ctx.set('Cache-Control', 'public, max-age=180');
+	setResponseType(ctx);
+});
```
mei23: うーん、インスタンスに保存されているユーザーのパターンなら、どんなパターンでも作らされてしまうのが少し気になってる。

Johann150: Yes, it will be technically possible to create any pair of users stored in the database.

As I have said above already, being able to create this means little: Misskey uses follow requests to track whether a follow was requested by a local user. Before accepting a follow, Misskey checks that a matching follow request exists. If it does not exist, nothing happens.

Similarly, using this ID to create a fraudulent Activity should also not work because the actor that made the signature will not be the same as the `actor` of the `Follow` activity. This would only be a concern if the private key becomes public, which is in my understanding outside the threat model of Misskey.

To ensure this is not abused I think it does make sense to check that the follower is a local user and the followee is not a local user by adding those parameters to the queries.

If this is not what you meant, please be more specific.

mei23: まあ、そうは思うわ…

でもやっぱり、HTTP Resolver経由で取得するのが引っかかる。（endpointの公開が必要になってしまう）
Activity endpointは公開せずに /packages/backend/src/remote/activitypub/kernel/accept/follow.ts などでURLパースでDB解決するとかしたい。
Related #8688

Johann150: The resolve of the accepted object happens in `packages/backend/src/remote/activitypub/kernel/accept/index.ts`, and there a `Resolver` is used. Parsing the URL instead of making a HTTP request is the job of `DbResolver`. I am not sure if I can replace `Resolver` with a `DbResolver` there without breaking behaviour.

If it is possible, a new `getFollowFromApId` function could be added to `DbResolver` to be used instead, since misskey only seems to be able to accept `Follow` objects there anyway.

mei23: I was just thinking about it.

In the case of Undo, there can be Follow and Block, but since it is from local, it can be identified by URL Path.

mei23:
It may be another issue, but should the path that stores Activity of ActivityPub be isolated to /activities/ or another path?

mei23:
There is a possibility of conflict when implementing the Web or other protocols, and in other implementations we often see isolation in a dedicated path.

Johann150:
It might have been a good design idea if it had been made at an earlier point. I think even besides it being a separate issue, such a change has the potential to (badly) break federation for servers that keep track of the existing IDs (for example: Misskey keeps track of the remote IDs of notes).

So in Misskey's case I think it is too late to change this.

mei23:
Yes, most objects are too slow, but I'm wondering what about new Activities.

packages/backend/src/server/activitypub.ts
Johann150: As I said before, I think this should be handled in the issue you linked, because in my opinion, there needs to be a way to automatically fetching the right type of object locally, as I wrote in <https://github.com/misskey-dev/misskey/issues/8688#issuecomment-1128811932>.
An additional complexity is that the kernel always uses ActivityPub objects, so use of `DbResolver` might be inappropriate because it returns database objects.

Johann150:
Having different URL styles would require changes to `DbResolver.parseUri` which might be used in the future e.g. for #8688. So I do not want to create a mess of different URL formats.

mei23:
Okay, there is such an idea.

However, `DbResolver.parseUri` does not assume two arguments, so it may be necessary to change it after all ...

Johann150:
It would still create a mess of different URLs which is not something I would want to introduce.

mei23:
How about `blocks`? I think it's the same pattern as `follows`.

mei23:
> How about `blocks`? I think it's the same pattern as `follows`.

But unlike `follows`, ` blocks` can use an immutable DB ID.

Johann150:
Does it matter for blocks? I don't think there is any case where local blocks are handled via ActivityPub, since they do not have to be `Accept`ed.

But I guess it would be best practice anyway...

mei23:
I agree. There is no problem for the time b

syuilo:
👍

PR #8694, opened 2022-05-17
perf: resolver recognizes local URLs

### LINKED ISSUE #8688: Resolver tries to fetch local ActivityPub content via HTTP

Johann150:
# What
The ActivityPub resolver will recognize local URLs and not make HTTP requests for them.

Also some minor style and type fixes.
# Why
fix #8688

### DIFF
--- a/packages/backend/src/misc/fetch.ts
+++ b/packages/backend/src/misc/fetch.ts
@@ -6,7 +6,7 @@ import fetch from 'node-fetch';
 import { HttpProxyAgent, HttpsProxyAgent } from 'hpagent';
 import config from '@/config/index.js';
 
-export async function getJson(url: string, accept = 'application/json, */*', timeout = 10000, headers?: Record<string, string>) {
+export async function getJson(url: string, accept = 'application/json, */*', timeout = 10000, headers?: Record<string, string>) : Record<string, any> {
 	const res = await getResponse({
 		url,
 		method: 'GET',
@@ -20,7 +20,7 @@ export async function getJson(url: string, accept = 'application/json, */*', tim
 	return await res.json();
 }
 
-export async function getHtml(url: string, accept = 'text/html, */*', timeout = 10000, headers?: Record<string, string>) {
+export async function getHtml(url: string, accept = 'text/html, */*', timeout = 10000, headers?: Record<string, string>) : string {
 	const res = await getResponse({
 		url,
 		method: 'GET',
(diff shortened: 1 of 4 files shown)


packages/backend/src/misc/fetch.ts
```diff
@@ -6,7 +6,7 @@ import fetch from 'node-fetch';
 import { HttpProxyAgent, HttpsProxyAgent } from 'hpagent';
 import config from '@/config/index.js';
 
-export async function getJson(url: string, accept = 'application/json, */*', timeout = 10000, headers?: Record<string, string>) {
+export async function getJson(url: string, accept = 'application/json, */*', timeout = 10000, headers?: Record<string, string>) : Record<string, any> {
```
syuilo: JSON is not only an object, but string, number, array, bool, null can be placed at the top level, which may cause a mismatch with the type definition.

Johann150: I didn't check before but now I saw that all uses of `getJson` are cast to a more specific type anyway. Same applies to `signedGet`. Reverted the type fix.

packages/backend/src/remote/activitypub/resolver.ts
```diff
@@ -70,4 +89,52 @@ export default class Resolver {
...
+
+		if (!url.startsWith(config.url + '/') || !url.slice(localOffset).includes('/')) {
+			throw new Error(`resolveLocal: unrecognized URL ${url}`);
+		}
+
+		const i = url.slice(localOffset).indexOf('/');
+		const type = url.slice(localOffset, localOffset + i);
+		const id = url.slice(localOffset + i + 1);
+
+		const activitySuffix = '/activity';
```
mei23, line 104: Can't you share the URI separation routine with the following or use a regular expression in one shot?
https://github.com/misskey-dev/misskey/blob/fee2878b98253f2510ae2bc5c746c058a043ef82/packages/backend/src/remote/activitypub/db-resolver.ts#L124-L141

Johann150: Maybe it should be made `static` so I do not need to make an instance of `DbResolver` to use it.

Johann150: I wonder if `DbResolver` should have a general resolving function. If `Resolver` detects that a URL is local it could then delegate to `DbResolver`.

Johann150:
I think this is not ready yet and needs more discussion in the original issue.

PR #8702, opened 2022-05-18
enhance: uniform theme color

Johann150:
# What
All theme colors are formatted in a uniform way, e.g. `#ffffff`.
- Existing theme colors are reformatted. Invalid colors are set to `null`.
- When fetching instance metadata, the color is formatted in the correct way. Invalid colors are not used.

Also applies to the instance's own theme color:
- The instances own theme color (`meta.themeColor`) is reformatted. If invalid it is set to `null`.
- The API endpoint to set the theme color has stricter validation.

# Why
https://github.com/misskey-dev/misskey/pull/8697#issuecomment-1129268434

This saves repeated computation in the client when rendering the theme color.


syuilo (approved):

Johann150:
merging this because [✅ approved](https://github.com/misskey-dev/misskey/pull/8702#pullrequestreview-977897536) and #8697 was merged already and depends on this

PR #8703, opened 2022-05-18
fix: first user gets admin

### LINKED ISSUE #8685: First user did not receive admin rights
## 💡 Summary

I am moving my instance from one VPS to another on a different service. I did a fresh install on the new VPS and created the first account. It did not receive admin rights and it is the only user showing. Can someone advise how I can set the account as admin in the command line?

## 🥰 Expected Behavior

First account should become admin.

## 🤬 Actual Behavior

First account is just a regular account.

## 📝 Steps to Reproduce

1. Point domain at new IP
2. Install Misskey
3. Register first user

## 📌 Environment

Misskey version: 12.110.1
Your OS: Debian 10
Your browser: Safari and Firefox

Johann150:
# What
Some of the system users may be created before the admin signs up. Because the system users do not get admin rights, adding a check for `isAdmin` will ignore system users in this check.

# Why
fix #8685

### DIFF
--- a/packages/backend/src/server/api/common/signup.ts
+++ b/packages/backend/src/server/api/common/signup.ts
@@ -87,6 +87,7 @@ export async function signup(opts: {
 			token: secret,
 			isAdmin: (await Users.countBy({
 				host: IsNull(),
+				isAdmin: true,
 			})) === 0,
 		}));
 
--- a/packages/backend/src/server/api/endpoints/admin/accounts/create.ts
+++ b/packages/backend/src/server/api/endpoints/admin/accounts/create.ts
@@ -33,6 +33,7 @@ export default define(meta, paramDef, async (ps, _me) => {
 	const me = _me ? await Users.findOneByOrFail({ id: _me.id }) : null;
 	const noUsers = (await Users.countBy({
 		host: IsNull(),
+		isAdmin: true,
 	})) === 0;
 	if (!noUsers && !me?.isAdmin) throw new Error('access denied');
 
--- a/packages/backend/src/server/api/endpoints/meta.ts
+++ b/packages/backend/src/server/api/endpoints/meta.ts
@@ -385,6 +385,7 @@ export default define(meta, paramDef, async (ps, me) => {
 			cacheRemoteFiles: instance.cacheRemoteFiles,
 			requireSetup: (await Users.countBy({
 				host: IsNull(),
+				isAdmin: true,
 			})) === 0,
 		} : {}),
 	};


tamaina:
メモ: 最初のadminアカウント作成はadmin/accounts/createなので問題~~ない~~大アリじゃん

Johann150:
The change seems to be necessary there as well. But why is there this check in the services code then?

Edit: ah, I see it is used for different checks, so I think both changes are needed.

tamaina (approved):

mei23:
これAdminが消えてたら次に来た人がAdminになっちゃわない？

Johann150:
Yes, that is the intended behaviour. I think if all admins delete their accounts, something went completely wrong already so I'm not sure this should be considered as a likely scenario. How else would you solve the issue?

mei23:
Local system users always contain `.`, So I think you can exclude them.

Johann150:
I am concerned about the different processing needs for checking a string with `LIKE` and comparing a boolean column.

Maybe another approach is needed: Block creation of all system users until an administrator has been registered. (If a system user would be required to complete the operation, it will fail instead.) This should ensure that the first user is never a system user. What do you think?

mei23:
> I am concerned about the different processing needs for checking a string with LIKE and comparing a boolean column.

ローカルユーザーのみにユーザー名スキャンするだけ & signup時にしか発生しない、から大したことないかなと。

または、システムユーザーは2つしかなくて、~~そのうち自動的に作られるのは1つ `instance.actor` しかないから~~ いや、`relay.actor` も作られる？
それだけ除外するだけでもいいかも。

tamaina:
> Adminが消えてたら

そんなことある？

Johann150:
I think it is unlikely that all admins get deleted.

But if we want to be sure we could throw an exception somewhere here when there is no existing admin account, to block creation of all system accounts.
https://github.com/misskey-dev/misskey/blob/7ed0763ad2357ff7edeb2bfd7428dd50f6c512a8/packages/backend/src/services/create-system-user.ts#L29-L36

mei23:
>> Adminが消えてたら
> 
>そんなことある？

Adminは普通に自分で消せそう

v10まではユーザーは論理削除だったから最後のAdminを消しても問題はなかったんだけど…

Johann150:
The API technically allows deletion of all admins, `packages/backend/src/server/api/endpoints/i/delete-account.ts` does not check for `isAdmin`. But if that is the problem I think it should be handled in a separate issue / PR.

Johann150:

PR #8713, opened 2022-05-20
refactor: temporary files

Johann150:
# What
Refactors the use of temporary files and the `tmp` module. To reduce code duplication, the already existing `createTemp` function is used instead.

Some uses of temporary files are missing cleanup entirely or may miss cleanup under some error conditions. These uses are refactored to ensure that the cleanup procedure for temporary files is always run.

# Why
- follow the DRY principle to create more easily maintainable code
- large number of temporary files may fill up hard drive space unnecessarily because of missing cleanup

### DIFF
--- a/packages/backend/src/misc/create-temp.ts
+++ b/packages/backend/src/misc/create-temp.ts
@@ -1,7 +1,7 @@
 import * as tmp from 'tmp';
 
-export function createTemp(): Promise<[string, any]> {
-	return new Promise<[string, any]>((res, rej) => {
+export function createTemp(): Promise<[string, () => void]> {
+	return new Promise<[string, () => void]>((res, rej) => {
 		tmp.file((e, path, fd, cleanup) => {
 			if (e) return rej(e);
 			res([path, cleanup]);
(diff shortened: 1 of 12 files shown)


syuilo (approved):

packages/backend/src/misc/create-temp.ts
```diff
@@ -8,3 +8,12 @@ export function createTemp(): Promise<[string, any]> {
 		});
 	});
 }
+
+export function createTempDir(): Promise<[string, any]> {
+	return new Promise<[string, any]>((res, rej) => {
```
acid-chicken:
```suggestion
export function createTempDir(): Promise<[string, () => void]> {
	return new Promise<[string, () => void]>((res, rej) => {
```

Maybe we can avoid `any` here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/75f98a0b191aa1458658b90b5b1329748cc3c082/types/tmp/index.d.ts#L121

Johann150: I also fixed the use of `any` for the `createTemp` function, since the same applies there.

syuilo:
🙏

packages/backend/src/services/drive/generate-video-thumbnail.ts
```diff
@@ -1,38 +1,31 @@
...
-	// JPEGに変換 (Webpでもいいが、MastodonはWebpをサポートせず表示できなくなる)
-	const thumbnail = await convertToJpeg(outPath, 498, 280);
-
-	// cleanup
-	await fs.promises.unlink(outPath);
-	cleanup();
 
-	return thumbnail;
+		// JPEGに変換 (Webpでもいいが、MastodonはWebpをサポートせず表示できなくなる)
+		return await convertToJpeg(498, 280);
```
mei23, line 27: https://github.com/misskey-dev/misskey/issues/8814

PR #8714, opened 2022-05-20
Add pixeldesu to patron list

pixeldesu:
# What

Just adding myself to the list of patrons on the about page!

# Why

Because I pledge money to support Misskey!


syuilo:
Thanks!

PR #8716, opened 2022-05-20
Prevent access to user pages when not logged in

### LINKED ISSUE #8377: User settings shouldn't be accessible when logged out

pixeldesu:
# What

This PR uses the newly refactored `pleaseLogin` script to prevent non-logged in users from accessing parts of the website that require a user object or are in general not supposed to be visible to off-site users.

It also removes two routes that point to pages that don't exist as components (`advanced-theme-editor` and `test`)

# Why

Pages that require user data are broken, and thus shouldn't be accessible in that case. Trying to close the sign-in modal then redirects to the homepage, preventing the users from seeing (partially) broken pages and thus improving the user experience.

fixes #8377

# Additional info (optional)

I extended `pleaseLogin` with another optional argument `stopExecution`. This allows to trigger the throwing of the error on demand, because if logic is followed after the prompt that should be cancelled, we need the error, but in cases like the navigation or component rendering this breaks Misskey.

### DIFF
--- a/packages/client/src/scripts/please-login.ts
+++ b/packages/client/src/scripts/please-login.ts
@@ -3,7 +3,7 @@ import { $i } from '@/account';
 import { i18n } from '@/i18n';
 import { popup } from '@/os';
 
-export function pleaseLogin(path?: string) {
+export function pleaseLogin(path?: string, stopExecution: boolean = true) {
 	if ($i) return;
 
 	popup(defineAsyncComponent(() => import('@/components/signin-dialog.vue')), {
@@ -17,5 +17,5 @@ export function pleaseLogin(path?: string) {
 		},
 	}, 'closed');
 
-	throw new Error('signin required');
+	if (stopExecution) throw new Error('signin required');
 }
--- a/packages/client/src/scripts/please-login.ts
+++ b/packages/client/src/scripts/please-login.ts
@@ -3,7 +3,7 @@ import { $i } from '@/account';
 import { i18n } from '@/i18n';
 import { popup } from '@/os';
 
-export function pleaseLogin(path?: string, stopExecution: boolean = true) {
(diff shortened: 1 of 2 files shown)


ThatOneCalculator:
hi chat :star_struck:

ThatOneCalculator (approved):

packages/client/src/scripts/please-login.ts
```diff
@@ -3,7 +3,7 @@ import { $i } from '@/account';
 import { i18n } from '@/i18n';
 import { popup } from '@/os';
 
-export function pleaseLogin(path?: string) {
+export function pleaseLogin(path?: string, stopExecution: boolean = true) {
```
Johann150: I think these two are exclusive: If a path is provided, the execution should continue (with the redirection), and if no path is provided,  the execution should not continue. So I think instead of adding a separate parameter you could check if the existing parameter is set or not.

Johann150 (approved):

ThatOneCalculator:
@pixeldesu Seems like this PR doesn't catch trying to make a note.
image

pixeldesu:
@ThatOneCalculator I'd say that's a completely different problem.

The hotkey to open the post form should not even do anything if not logged in, there's already an issue open for that!

ThatOneCalculator:
<details><summary>Off topic discussion, I suppose</summary>

> The hotkey to open the post form should not even do anything if not logged in, there's already an issue open for that!

Suppose the <kbd>I :heart: #Misskey</kbd> button should be disabled too? @pixeldesu do you know the issue number?

image


</details>

Johann150:
superseded by #8904

PR #8733, opened 2022-05-25
fix: ensure resolver does not fetch local resources via HTTP(S)

### LINKED ISSUE #8688: Resolver tries to fetch local ActivityPub content via HTTP

Johann150:
# What
## refactor `parseUri`
`DbResolver.parseUri` is made into a function instead of being a method of `DbResolver`. The return type of `parseUri` is changed to a union type to better reflect the actual return type. To be able to easily distinguish between the two, a new field `local` is added to tell the two types apart.

The implementation is adjusted so it can recognize that there is additional data after the type and ID. This is useful for URIs like `https://example.com/notes/0000000000/activity` and the new kind of ID introduced in #8689 (e.g. `https://example.com/follows/0000000000/0000000001`)

## enhance `Resolver`

When `Resolver` is used for local URIs it will defer the fetching to `DbResolver` instead. Since `Resolver` is expected to return the ActivityStreams types but `DbResolver` returns the Misskey database types, some glue code is necessary to convert the objects.

The new URI for follows is also already implemented to work together with #8689.

# Why
fix #8688


syuilo:
👍

PR #8734, opened 2022-05-25
fix: assume remote users are following each other

Johann150:
# What
Misskey does not know if two remote users are following each other. Because ActivityPub actions would otherwise fail on followers only notes, we have to assume that two remote users are following each other when an interaction about a remote note occurs.

# Why
Incoming reactions of remote users to follower only notes of remote users fail to be delivered due to changes introduced in #8534.


syuilo:
It might be great to be able to add a test for this process.

PR #8735, opened 2022-05-26
feat(tests): add e2e tests for widgets

pixeldesu:
# What

This PR adds 22 end-to-end test cases for adding all widgets to the widget sidebar.

3 basic tests:
* checking if the `Edit widgets` button is present
* checking if the select is visible after `Edit widgets` has been pressed (Edit mode)
* checking if removing widgets works

19 tests for all current widgets

# Why

After the recent merge of the widget refactoring/lint issues a lot of regressions happened. Testing if widgets are present after being added at least makes sure that they are not breaking the application completely.

# Additional info (optional)

* I added `mk-widget-edit/select/add` classes to mark up the elements that are used to add widgets
* I added more `mkw-*` classes to the widgets that had them missing
* The test cases do not check for specific widget functionality, they only check that the wrapped `mkw-{widgetName}` is present after being added

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,6 @@ on:
     branches:
       - master
       - develop
-      - feature/widget-tests
   pull_request:
 
 jobs:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+      - feature/widget-tests
   pull_request:
 
 jobs:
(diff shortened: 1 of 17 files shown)


syuilo:
Cool!
Do you know why the e2e test is failing?

pixeldesu:
@syuilo No clue 😕 

Before the final push, that case worked...so I assume that it's something timeout-related mainly and we might need to adjust something here so that it consistently works.

Here's the last action run on my fork before I made the PR, and there it passed fine!
https://github.com/pixeldesu/misskey/runs/6603509947?check_suite_focus=true

This also might be an issue with the `basic` testsuite, as the beforeEach-hook is the same there.

syuilo:
👍🏻👍🏻👍🏻

PR #8737, opened 2022-05-26
fix lints

Johann150:
# What
Fixes multiple different kinds of lints, one category of lints per commit.

# Why
Improve code quality.


syuilo:
Thanks🙏

PR #8740, opened 2022-05-26
enhance: replace signin CAPTCHA with rate limit

### LINKED ISSUE #8739: Concerns about requiring Captcha for login
I feel ***requiring*** either reCaptcha or hCaptcha for login is a bad idea. While Captchas can be great (and I already used them on my instance for signing up), forcing them for login poses three big downsides:

- Breaks login for third party clients
- Forces users to rely on non-free network services
- Gives admins less choice

If Misskey had some sort of built-in Captcha solution I could *maybe* understand this decision, but as of right now it really doesn't seem like a good idea.

Johann150:
# What
Instead of solving a CAPTCHA, the signin endpoint uses a rate limit. For this to work the rate limiter code had to be slightly adjusted to also work with IP addresses if no user has yet signed in. With this adjustment it might be possible to rate limit other API endpoints in the future as well, without requiring signing in.

# Why
- fix #8739
- undo breaking change to signin process

# Additional info
Applications which are using this endpoint should be aware that it is not strictly a part of the API. The native login token that gets returned from this endpoint gives special privileges due to the `secure` flag on some endpoints. This includes for example: 2FA setup, data import & export, app authentication, miauth, account deletion.

But these capabilities may be desired for alternative front ends.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,8 @@ You should also include the user name that made the change.
   Your own theme color may be unset if it was in an invalid format.
   Admins should check their instance settings if in doubt.
 - Perform port diagnosis at startup only when Listen fails @mei23
+- Rate limiting is now also usable for non-authenticated users. @Johann150
+  Admins should make sure the reverse proxy sets the `X-Forwarded-For` header to the original address.
 
 ### Bugfixes
 - Client: fix settings page @tamaina
--- a/locales/ja-JP.yml
+++ b/locales/ja-JP.yml
@@ -842,6 +842,7 @@ oneDay: "1日"
 oneWeek: "1週間"
 reflectMayTakeTime: "反映されるまで時間がかかる場合があります。"
 failedToFetchAccountInformation: "アカウント情報の取得に失敗しました"
+rateLimitExceeded: "レート制限を超えました"
 
 _emailUnavailable:
   used: "既に使用されています"
--- a/packages/backend/src/server/api/call.ts
+++ b/packages/backend/src/server/api/call.ts
@@ -6,6 +6,7 @@ import endpoints, { IEndpointMeta } from './endpoints.js';
 import { ApiError } from './error.js';
 import { apiLogger } from './logger.js';
 import { AccessToken } from '@/models/entities/access-token.js';
+import IPCIDR from 'ip-cidr';
 
 const accessDenied = {
 	message: 'Access denied.',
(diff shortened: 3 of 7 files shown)


tamaina:
I believe that targeting only IP addresses for rate limiting is not an effective defense against brute force attacks.

I think it is better to implement limiting the number of login attempts per username.

tamaina:
I reconsidered that IP limiting would have a certain effect.

"Can we generate a lot of IP addresses nowadays?" I thought, but, cloud services may be not stupid and they will ban us if we issue too many IP addresses.

ThatOneCalculator:
In my opinion, this is a much better solution, and works great with Kaiteki.

tamaina:
(That said, it's not a good idea for third-party apps to implement their own login screens.)

ThatOneCalculator:
Agreed (told the Kaiteki developer about that), but at least it works. Kaiteki also plans to offer itself as a default alternative frontend for multiple fediverse services so that's part of why they want a custom login screen.

ThatOneCalculator (approved):
Approving as it works as expected on my instance. Did a couple spamming tests.

syuilo:
- The change to allow rate limits to work on a per IP address basis is great.
- But in terms of security at signin, as tamaina mentioned, it would be better (ideally) to keep track of the number of consecutive failed signin attempts per user and require a CAPTCHA after a certain number of failed attempts.
- I am not sure which is better, the current implementation that requires a CAPTCHA for all signins or the implementation that applies rate limits to signins as well, like this PR.

ThatOneCalculator:
> I am not sure which is better, the current implementation that requires a CAPTCHA for all signins or the implementation that applies rate limits to signins as well, like this PR.

I personally feel that merging this for now (before the release of 12.111.x) would be the smarter idea, and then add in Captcha after multiple failed sign-ins later.

ThatOneCalculator:
@Johann150 

> Admins should make sure the reverse proxy sets the X-Forwarded-For header to the original address.

Do you have an updated version of the default nginx config file with this change, if it needs to be changed at all?

tamaina (approved):

tamaina:
p1.a9z.devに適用してみた

ThatOneCalculator:
Also applied to the main build of stop.voring.me

Johann150:
> I think it is better to implement limiting the number of login attempts per username.

I think this would easily allow for a DoS attack which would stop a legitimate user from being able to log in. That is why I chose the IP address for limiting instead.

> Can we generate a lot of IP addresses nowadays?

That is a concern with IPv6 addresses, because a single device can have very many individual addresses assigned to the same interface. That is why I added IPv6 mitigation in commit ddf18dfbe3090eeb1c8abcf8325f9cf66a06145f to only use the first 64 bits. The first 64 bits are usually the network prefix for a home network (i.e. at least some of the other 64 bits may be able to be set by the device as it wishes).

Johann150:
> Do you have an updated version of the default nginx config file with this change, if it needs to be changed at all?

The version from misskey hub already includes the necessary configuration in both the [Japanese](https://github.com/misskey-dev/misskey-hub/blob/701dce9cfa5c476fc43e9014205bc7e9fffb6ee8/src/docs/admin/nginx.md?plain=1#L72) and [English](https://github.com/misskey-dev/misskey-hub/blob/701dce9cfa5c476fc43e9014205bc7e9fffb6ee8/src/en/docs/admin/nginx.md?plain=1#L72) version of the page.

ThatOneCalculator:
Perfect! Already using it, just making sure :)

pixeldesu (approved):

syuilo:
👍🏻👍🏻

PR #8746, opened 2022-05-27
fix: correctly render empty note text

### LINKED ISSUE #8745: empty note content is not correctly represented in some cases
## 💡 Summary
If a note has no content (e.g. only image attachments) on the original instance, it will be displayed with a single dot as its text content on remote instances.

Deleting the note and manually fetching it fixes this.

## 🥰 Expected Behavior
Initial federation and manually fetching both correctly result in empty content.

## 🤬 Actual Behavior
Initial federation results in content being `.`.

## 📝 Steps to Reproduce

1. Send note with no text and an image or poll attached.
2. View note on another instance to which it was automatically federated.
3. Note content is `.` (wrong)
4. Admin deletes the note.
5. Search for the note URI, so Misskey refetches it.
6. Note content is empty (correct)

## 📌 Environment

Misskey version: 12.110.1

Johann150:
# What
Ensure that the `_misskey_content` attribute will always be set. Because the API endpoint does not require the existence of the `text` field, that field may be `undefined` when rendering the ActivityPub representation. If it does not exist it will now be set to `null` instead.

The code was made a bit more succinct by using the null coercion operator for initializing `apText`.

# Why
fix #8745

# Additional info
While this is not strictly necessary to fix the issue at hand, the rendered HTML of a note with empty text will now also be rendered correctly. From git blame it seems that the previous behaviour was added because of a Mastodon bug that might have previously existed. However, this seems to be no longer the case, as I can find Mastodon posts that have empty content.

### DIFF
--- a/packages/backend/src/remote/activitypub/misc/get-note-html.ts
+++ b/packages/backend/src/remote/activitypub/misc/get-note-html.ts
@@ -3,8 +3,6 @@ import { Note } from '@/models/entities/note.js';
 import { toHtml } from '../../../mfm/to-html.js';
 
 export default function(note: Note) {
-	let html = note.text ? toHtml(mfm.parse(note.text), JSON.parse(note.mentionedRemoteUsers)) : null;
-	if (html == null) html = '<p>.</p>';
-
-	return html;
+	if (!note.text) return '';
+	return toHtml(mfm.parse(note.text), JSON.parse(note.mentionedRemoteUsers));
 }
--- a/packages/backend/src/remote/activitypub/renderer/note.ts
+++ b/packages/backend/src/remote/activitypub/renderer/note.ts
@@ -82,15 +82,15 @@ export default async function renderNote(note: Note, dive = true, isTalk = false
 
 	const files = await getPromisedFiles(note.fileIds);
 
-	const text = note.text;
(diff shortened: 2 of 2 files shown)


pixeldesu (approved):
Johann provided a test post using this PR: https://genau.qwertqwefsday.eu/notes/90yt7bhfsg

<details>
<summary>Post from Mastodon view</summary>

image

</details>

[Post on Misskey.io](https://misskey.io/notes/90yt7bhfii)

This looks fine! 🎉

acid-chicken:
idk why we were doing this 54dcc10250cce8bfadf424245252719fea8a3896
cc: @syuilo これ、なんでこうしてたんだっけ

syuilo:
> idk why we were doing this [54dcc10](https://github.com/misskey-dev/misskey/commit/54dcc10250cce8bfadf424245252719fea8a3896)
> cc: @syuilo これ、なんでこうしてたんだっけ

画像添付されててもテキストが空だとMastodonで表示されないみたいな理由だったと思う

Johann150:
I guess that bug must have been fixed since then as it worked fine like this.

syuilo:
🙏🏻

PR #8761, opened 2022-05-29
Import shared ESLint config in client package

pixeldesu:
# What

Because of errors with ESLint and TypeScript/Vue setups the shared configuration was never imported, causing us to miss out on a few project-specific code style guidelines (like semicolons, spacing, etc.)

# Why

This is done to ensure a consistent code style between all parts of Misskey.

# Additional info (optional)

The errors from before are fixed like so:
* typescript-eslint has an option `extraFileExtensions` to specify for non-TS files
* `parserOptions.project` is set again (also was commented out)
* in `tsconfig.json` we add `.eslintrc.js` to the project so IDEs/ESLint does not complain about non-project files

### DIFF
--- a/packages/client/.eslintrc.js
+++ b/packages/client/.eslintrc.js
@@ -1,68 +1,79 @@
 module.exports = {
 	root: true,
 	env: {
-		"node": false
+		'node': false,
 	},
-	parser: "vue-eslint-parser",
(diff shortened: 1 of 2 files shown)


syuilo:
👍🏻

PR #8762, opened 2022-05-29
User moderation details

### LINKED ISSUE #5502: すでに凍結されたローカルユーザーのSignin-Historyを照会できるAdmin API
## Summary

タイトル通り

PikaDude:
# What

This PR allows administrators to view more detailed information about users, such as their sign-in history, email, and more.

# Why

Currently, there is no interface to help determine alternative or spam accounts. In the interface, there is no IP information, no email information, etc.
While you can run queries to figure this information out, this is not a viable solution.

fix #5502

# Additional info (optional)

This information will only be available to administrators.


packages/backend/src/server/api/endpoints/admin/show-user.ts
```diff
@@ -24,8 +24,9 @@ export const paramDef = {
 // eslint-disable-next-line import/no-default-export
 export default define(meta, paramDef, async (ps, me) => {
 	const user = await Users.findOneBy({ id: ps.userId });
+	const profile = await UserProfiles.findOneBy({ userId: ps.userId });
```
Johann150: You could use `await Promise.all` here to await both promises so both database operations can be run simultaneously.

packages/backend/src/server/api/endpoints/admin/show-user.ts
```diff
@@ -34,8 +35,37 @@ export default define(meta, paramDef, async (ps, me) => {
 		throw new Error('cannot show info of admin');
 	}
 
+	if (!_me.isAdmin) {
```
Johann150, line 40: I am not sure what `_me` is for in this case as I think it should be eqivalent to `me`. We might want to remove `_me` in later refactoring, so please use `me` here instead.
```suggestion
	if (!me.isAdmin) {
```

packages/client/src/pages/user-info.vue
```diff
@@ -54,6 +54,9 @@
 				<FormButton v-if="user.host != null" class="_formBlock" @click="updateRemoteUser"><i class="fas fa-sync"></i> {{ $ts.updateRemoteUser }}</FormButton>
 			</FormSection>
 
+			<MkObjectView v-if="info && info.email !== undefined" tall :value="info">
```
Johann150: I think checking for `info.email` here is a suboptimal solution, not least because some instances do not require email addresses for signup. I checked on my own instance and in such a case `info.email` is not set so would be `undefined`.
I'm not sure if the `v-if` is needed here at all.

Johann150 (changes requested):

Johann150:
Hi, thanks for your contribution! Please fix the issues I noted above before these changes can go in.

I also slightly modified your original comment so the issue is properly linked.

packages/backend/src/server/api/endpoints/admin/show-user.ts
PikaDude: The _me is used as it was already used in the lines before ¯\\\_(ツ)\_/¯

packages/client/src/pages/user-info.vue
PikaDude: v-if is used to avoid showing the following data which is entirely redundant to moderators
```ts
return {
    isModerator: user.isModerator,
    isSilenced: user.isSilenced,
    isSuspended: user.isSuspended,
};
```

packages/backend/src/server/api/endpoints/admin/show-user.ts
```diff
@@ -23,9 +23,9 @@ export const paramDef = {
 
 // eslint-disable-next-line import/no-default-export
 export default define(meta, paramDef, async (ps, me) => {
-	const user = await Users.findOneBy({ id: ps.userId });
+	const [user, profile] = await Promise.all([Users.findOneBy({ id: ps.userId }), UserProfiles.findOneBy({ userId: ps.userId })]);
```
Johann150: I think spreading this over multiple lines makes this easier to read.
```suggestion
	const [user, profile] = await Promise.all([
	 Users.findOneBy({ id: ps.userId }),
	 UserProfiles.findOneBy({ userId: ps.userId })
	]);
```

Johann150 (approved):
Small stylistic nitpick, other thank that looks good.

Johann150:
Something else I noticed while testing is that from the UI its a bit weird to have two `MkObjectView`s right next to each other because it is not even very clear that they are two separate elements. Maybe some margin or border between them would be a nice addition.

PikaDude:
> Something else I noticed while testing is that from the UI its a bit weird to have two `MkObjectView`s right next to each other because it is not even very clear that they are two separate elements. Maybe some margin or border between them would be a nice addition.

Is this really a problem though? Currently it sort of looks like there's two JavaScript objects or something when the two ObjectViews are next to each other as they are. Because technically speaking they are referring to the same user/source, so it would make sense that they would be related together and look like they're apart of the same parent object.

packages/backend/src/server/api/endpoints/admin/show-user.ts
```diff
@@ -23,9 +23,12 @@ export const paramDef = {
 
 // eslint-disable-next-line import/no-default-export
 export default define(meta, paramDef, async (ps, me) => {
-	const user = await Users.findOneBy({ id: ps.userId });
+	const [user, profile] = await Promise.all([
+	 Users.findOneBy({ id: ps.userId }),
+	 UserProfiles.findOneBy({ userId: ps.userId })
```
Johann150: sorry mixed up tabs and spaces, my suggested change breaks the lint

```suggestion
		Users.findOneBy({ id: ps.userId }),
		UserProfiles.findOneBy({ userId: ps.userId })
```

PikaDude:
@johann150 based on the reply I made above to your comment, is this PR okay to merge?

syuilo:
👍🏻

ThatOneCalculator:
Can I ask what `carefulBot` means?

Johann150:
It is a flag that the backend uses to determine whether a follow should be automatically accepted.
https://github.com/misskey-dev/misskey/blob/aea2f01ef7909885cf20df9bc2d27bb77863b35b/packages/backend/src/services/following/create.ts#L167
I don't think it is hooked up in the client.

PR #8767, opened 2022-05-31
Extract commonly used test logic to commands

pixeldesu:
# What

This PR extracts a lot of the commonly used logic into [Cypress commands](https://docs.cypress.io/api/cypress-api/custom-commands).

* `cy.resetState()` to reset the instance database and IndexedDB
* `cy.registerUser(username, password, isAdmin)` to create a (admin) user
* `cy.login(username, password)` to log in a user

# Why

This reduces code duplication, especially considering we want to write more tests in the future, which require a lot of the same setup code in the beginning.

# Additional info (optional)

Not every instance of logging in users or registering them has been replaced. I only replace them in `before/after{Each}` hooks because if we specifically test if registering/logging in works, that should not be hidden behind an abstraction.

Tests currently fail after a fixed point because the sign-in limiter kicks in, this is already being discussed in #7986.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,7 @@ on:
     branches:
       - master
       - develop
+      - feature/test-commands
   pull_request:
 
 jobs:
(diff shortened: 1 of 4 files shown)


Johann150 (approved):

syuilo:
🆒

PR #8771, opened 2022-05-31
fix: always remove completed tasks

Johann150:
# What
Always remove completed items from the task queues.

# Why
Do not take up unnecessary resources in Redis.

All other cases of adding tasks to a queue that I could find also used `removeOnComplete: true` so I assume leaving them out in these cases was not intended.


pixeldesu (approved):

syuilo:
👍🏻

PR #8772, opened 2022-05-31
fix(client): correctly handle MiAuth URLs with query string

### LINKED ISSUE #8608: MiAuth doesn't support callbacks with query segment
## 💡 Summary

As part of the MiAuth flow, misskey expects apps to perform a redirect to the permission grant page at `https://{host}/miauth/{session}` with a `callback` URL that the user will land on after permission is granted.

Currently when the user clicks accept [the relevant code](https://github.com/misskey-dev/misskey/blob/1f222e6cd19dcf2f4bb679e810090f11ce6fdc6a/packages/client/src/pages/miauth.vue#L85) just blindly appends `?session=...`, which - if used with redirect URLs like `https://blahaj.club/login/callback?type=misskey` - will result in a URL that (while *technically* valid) is hard to parse correctly.

This behavior has been present ever since MiAuth was introduced with 6be127e18bdbea6031698baeb8632f917a8346eb.

Because of how long this went (seemingly) unnoticed I'm not entirely sure if this is a bug or if MiAuth simply doesn't support this usecase. If it _is_ a valid bug I'd be happy to fix it.

## 🥰 Expected Behavior

The query segment is retained and correctly appended to (if present)

## 🤬 Actual Behavior

The callback URL ends up having two `?`s and requires mildly annoying workarounds to parse correctly.

## 📝 Steps to Reproduce

- Use the `/miauth/{session}` endpoint with a redirect URL like `https://blahaj.club/login/callback?type=misskey`

ghost:
# What

Patches the MiAuth flow to correctly append `session` to URLs that already have a query string in them.
A long-form explanation of the problem can be found in #8608.

# Why

Closes #8608 

# Additional Notes

Strictly speaking this could be considered a breaking API change since people (like myself) who added special misskey workarounds in their codebase for the previous broken behavior will not be able to use new MiAuth sessions after this update rolls out. Then again, given how long this has gone seemingly unnoticed, it might as well have no noticeable effect for most of the developers building on top of MiAuth. Hard to say really.

I'll leave deciding whether or not this deserves a changelog entry up to you.

### DIFF
--- a/packages/client/src/pages/miauth.vue
+++ b/packages/client/src/pages/miauth.vue
@@ -42,6 +42,7 @@ import MkSignin from '@/components/signin.vue';
 import MkButton from '@/components/ui/button.vue';
 import * as os from '@/os';
 import { login } from '@/account';
+import { appendQuery, query } from '@/scripts/url';
 
 export default defineComponent({
 	components: {
@@ -82,7 +83,9 @@ export default defineComponent({
 
 			this.state = 'accepted';
 			if (this.callback) {
-				location.href = `${this.callback}?session=${this.session}`;
+				location.href = appendQuery(this.callback, query({
+					session: this.session
+				}));
 			}
 		},
 		deny() {


Johann150 (approved):
Thanks for your contribution! The changes look good to me.

pixeldesu (approved):

syuilo:
🙏

PR #8778, opened 2022-06-01
refactor: configurable error images

### LINKED ISSUE #7043: Embed not found image in each domain instead of external domain?
## 💡 Summary

The "not-found" image is external and not local

## 🙂 Expected Behavior

Image should point to instance folder not external

## ☹️ Actual Behavior

Image pointing to external

## 📝 Steps to Reproduce

1. Visit `https://misskey.io/anything-here`
2. View image url

## 📌 Environment

Misskey 12.65.3

### LINKED ISSUE #8356: meta.errorImageUrl is unused
## 💡 Summary

The database field `meta.errorImageUrl` is unused.

1. Use it in the client to solve #7043. It should also be added back to the admin menu.
2. Remove the dead code and database field.

## 📌 Environment

Misskey version: all 12.x.x

Johann150:
# What
Make the error images configurable. They are configurable in `.config/default.yml`. With the default values the behaviour stays the same as it is currently.

The `mascotImageUrl` and `errorImageUrl` columns are removed from the `meta` database table because they were not used before. The possibility to adjust them via the admin API is also removed.

# Why
Some people want to either theme their instance with special images or do not find the currently used images appropriate for their use cases. Some people also do not want their instances to link to external resources.

fix #7043
fix #8356
related: https://github.com/misskey-dev/misskey/pull/3589#issuecomment-446121344

### DIFF
--- a/.config/example.yml
+++ b/.config/example.yml
@@ -146,3 +146,9 @@ id: 'aid'
 
 # Upload or download file size limits (bytes)
 #maxFileSize: 262144000
+
+# Images that will be shown by the web client, for example for error pages.
+#images:
+#  error: 'https://xn--931a.moe/assets/error.jpg'
+#  not_found: 'https://xn--931a.moe/assets/not_found.jpg'
+#  info: 'https://xn--931a.moe/assets/info.jpg'
--- a/packages/backend/migration/1654124623992-remove-unused-image-urls.js
+++ b/packages/backend/migration/1654124623992-remove-unused-image-urls.js
@@ -1,0 +1,14 @@
+export class removeUnusedImageUrls1654124623992 {
+	name = 'removeUnusedImageUrls1654124623992'
+
+	async up(queryRunner) {
+		await queryRunner.query(`ALTER TABLE "meta" DROP COLUMN "mascotImageUrl"`);
+		await queryRunner.query(`ALTER TABLE "meta" DROP COLUMN "errorImageUrl"`);
+	}
+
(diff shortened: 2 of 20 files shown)


Johann150:
I tested it with some different images:

<details>
<summary>screenshot: empty announcements list</summary>

image
</details>

<details>
<summary>screenshot: not found page</summary>

image
</details>

<details>
<summary>screenshot: error when trying to view a note that does not exist</summary>

image
</details>

pixeldesu (approved):

syuilo:
What is the reason for making it configurable in the configuration file instead of the DB?

Johann150:
I think the images are something that will usually be configured only once, so not using the database in this case can improve performance because it does not have to be unnecessarily reloaded from the database.

syuilo:
Since instance icon is stored in database, it would be better to unify one or the other in the future.

syuilo:
Since those images are prefetched in `backend/src/server/web/base.pug`, it is necessary to update that as well.

Johann150:
prefetch has been updated too

syuilo:
I think that the error image settings are client-specific settings, and it is a bit unnatural to have them in the server configuration file, and it would be more convenient to have them in the GUI, so I will implement the settings in the database.
Thanks anyway.

Johann150:
Next time just clearly tell me you do not want it instead of asking for adjustments. 💢

syuilo:
Sorry, did not have time to consider that in depth.

Issue #8785, opened 2022-06-04
federation: use `source` instead of custom `_misskey_content`


Johann150:
## Summary
It would be more in line with the ActivityPub specification to use the `source` attribute defined by ActivityPub (see [ActivityPub 2018 § 3.3](https://www.w3.org/TR/2018/REC-activitypub-20180123/#source-property)) to convey the original MFM markup. The custom `_misskey_content` should be deprecated in favor of it. From the specification:

<blockquote>

The `source` property is intended to convey some sort of source from which the `content` markup was derived, as a form of provenance, or to support future editing by clients. In general, clients do the conversion from `source` to `content`, not the other way around.

**Example 8:**
```json
{
  "@context": ["https://www.w3.org/ns/activitystreams",
               {"@language": "en"}],
  "type": "Note",
  "id": "http://postparty.example/p/2415",
  "content": "<p>I <em>really</em> like strawberries!</p>",
  "source": {
    "content": "I *really* like strawberries!",
    "mediaType": "text/markdown"}
}
```
</blockquote>

For MFM it could look something like this:
```json5
{
  "@context": [
    "https://www.w3.org/ns/activitystreams",
    // ...
  ],
  "type": "Note",
  "id": "http://misskey.example/notes/913n7143qa",
  "content": "<p>I <em>really</em> like strawberries!</p>",
  "source": {
    "content": "I *really* like strawberries!",
    "mediaType": "text/x.misskeymarkdown"
  },
  // ...
}
```
(The `text/x.misskeymarkdown` is just a suggestion that is compliant with [RFC 6838 § 3.4](https://datatracker.ietf.org/doc/html/rfc6838#section-3.4).)

Johann150:
The question is how to accomodate older versions of Misskey with this. I think for the time being we may want to send both for the time being. After some time has passed, the `_misskey_content` attribute could then be removed completely.

Johann150:
> you could/should probably use json-ld contexts to provide the migration path via aliases
> @CosmicToast https://github.com/misskey-dev/misskey/pull/8787#issuecomment-1153997532

In the case of Misskey and I think some other fediverse software ActivityPub is handled as JSON, not fully JSON-LD processed. So while I agree it would be good to have I don't think it is very important for our use case.

CosmicToast:
probably still a solid thing to do given it's how the spec works, I think

Johann150:
Thinking about it a bit more I'm not sure what exactly you mean with "migration path".

[...]

PR #8787, opened 2022-06-05
enhance(federation): use ActivityPub defined property in favour of proprietary property.

Johann150:
# What
Add ActivityPub `source` property.

For now this continues to write and read the `_misskey_content` property. This change is therefore fully backward compatible.

# Why
related to #8785. This PR does not completely fix that issue yet: The `_misskey_content` attribute should be removed in the future. I think we should wait *at least* 3 months after initial release of this change before removing `_misskey_content`.


tamaina:
p1.a9z.devに適用してみた

syuilo:
👍🏻

CosmicToast:
you could/should probably use json-ld contexts to provide the migration path via aliases

PR #8788, opened 2022-06-07
chore: fix some lints automatically

Johann150:
# What
Fix lints that were automatically fixable with `eslint --fix` in the client package.

# Why
fix lints

# Additional info
Running `eslint --fix` would also alter the spacing in the following areas. Because it would mess up the current alignment I did not commit this.
https://github.com/misskey-dev/misskey/blob/aea2f01ef7909885cf20df9bc2d27bb77863b35b/packages/client/src/scripts/hpml/lib.ts#L133-L169
https://github.com/misskey-dev/misskey/blob/aea2f01ef7909885cf20df9bc2d27bb77863b35b/packages/client/src/scripts/hpml/index.ts#L17-L23

### DIFF
--- a/packages/client/src/components/captcha.vue
+++ b/packages/client/src/components/captcha.vue
@@ -27,8 +27,7 @@ type CaptchaContainer = {
 };
 
 declare global {
-	interface Window extends CaptchaContainer {
-	}
+	type Window = CaptchaContainer
 }
 
 const props = defineProps<{
--- a/packages/client/src/components/captcha.vue
+++ b/packages/client/src/components/captcha.vue
@@ -27,7 +27,7 @@ type CaptchaContainer = {
 };
 
 declare global {
-	type Window = CaptchaContainer
+	interface Window extends CaptchaContainer { }
 }
 
 const props = defineProps<{
--- a/packages/client/.eslintrc.js
+++ b/packages/client/.eslintrc.js
@@ -15,6 +15,12 @@ module.exports = {
 		'plugin:vue/vue3-recommended',
 	],
 	rules: {
+		'@typescript-eslint/no-empty-interface': [
+			'error',
+			{
+				'allowSingleExtends': true,
+			},
+		],
 		// window の禁止理由: グローバルスコープと衝突し、予期せぬ結果を招くため
 		// data の禁止理由: 抽象的すぎるため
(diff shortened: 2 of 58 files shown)


pixeldesu (approved):

packages/client/src/components/captcha.vue
```diff
@@ -27,8 +27,7 @@ type CaptchaContainer = {
 };
 
 declare global {
-	interface Window extends CaptchaContainer {
-	}
+	type Window = CaptchaContainer
```
acid-chicken: I don’t think this change is semantically identical.

Johann150: As I said these changes were all automatical with `eslint --fix`. This is the relevant lint annotation for this change:

```
/home/runner/work/misskey/misskey/packages/client/src/components/captcha.vue
Error:   30:12  error  An interface declaring no members is equivalent to its supertype  @typescript-eslint/no-empty-interface
```

description relevant to the lint: https://typescript-eslint.io/rules/no-empty-interface/

acid-chicken: I saw, but the description in the link you provided;

> An empty interface is equivalent to its supertype. If the interface does not implement a supertype, then the interface is equivalent to an empty object (`{}`). In both cases it can be omitted.

is not strictly accurate.

https://www.typescriptlang.org/play?#code/JYOwLgpgTgZghgYwgAgEIE8CS5rycgbwChlTkBnAewFsIAuZAchmAgBsATcxogXyKJh0ABxQBRAB6QQ5YJRDIAvIRJkA2gH1koJjEqVGyAD5MARnCiMAugxABXaqeh8BoSLEQoM2d3hQQpCBAuZElpWXkVMmQAehjkMAALYHJtVIRKKCgIBDAXAUERL3QAQTZgOFTlYmjKJOgGZlZObnyhUTRS8sqlUMCZOQU4hOTUlOQQOuQMrJy8oA

acid-chicken: typescript-eslint/typescript-eslint#1640

Johann150: I will configure the rule with `allowSingleExtends: true` as suggested in that issue so the lint does not show up as wrong any more.

syuilo:
🙏

PR #8789, opened 2022-06-07
refactor: follow button

### LINKED ISSUE #7638: The follow-button does not cancel properly under a specific situation
## 💡 Summary

When the remote user is on an instance that is down or there is some other delayed connection between the remote and local instances, the follow-button for the local user does not display properly when cancelling a follow request.  While the follow-button is displaying "Processing", after clicking once or several times, the button continues displaying "Processing".

On the backend, for the server side, the event has properly cancelled the follow request. The front end is not updating properly.

For an end user who doesn't understand the issue, they may click the "Processing" button several times and not realize that the cancelled event has occurred. And by clicking "Processing" repeatedly, the user is sending a random number of requests to follow and unfollow without a  proper update on the front end.  

## 🙂 Expected Behavior

For a remote user who is on an instance that is down or delayed, and the local user has clicked "Follow", 
While the follow-button is displaying "Processing", the local user will click once and the button turns back into "Follow".

## ☹️ Actual Behavior

While the follow-button is displaying "Processing", the local user will click once or several times, the button continues displaying "Processing".

## 📝 Steps to Reproduce

To recreate the issue, we need a user on a server that is down. We can use @disney@birdsite.skye.cx while logged in at https://misskey.io
To recreate this from different instance, birdsite.skye.cx may not work. You would need a different server that is currently down but has previously federated with the instance. And a user from that down server.

1. From the front end of misskey, open the search box and enter @disney@birdsite.skye.cx  (you may need to refresh once more due to "error, try again").
2. Click "Follow" once. The button will turn into "Processing"
3. Wait a few seconds. 
4. Click "Processing" *once* to cancel the attempt to follow. The button will remain as "Processing".  Note; on the backend, the local instance has cancelled the request to follow, but the front end/client ui has not properly refreshed. 
5. Reload the page and the button will now say "Follow".

## 📌 Environment

Misskey 12.83 thru 12.87
May not be limited to these client versions. 

##  Proposed solution

This is working for me on my instance; 
within the file; src/client/components/follow-button.vue
between lines 112 and 113, add a line;
						this.hasPendingFollowRequestFromYou = false;						

@@ -110,6 +110,7 @@ export default defineComponent({
						await os.api('following/requests/cancel', {
							userId: this.user.id
						});
						this.hasPendingFollowRequestFromYou = false;						
					} else if (this.user.isLocked) {
						await os.api('following/create', {
							userId: this.user.id

Johann150:
# What
refactor follow button component

# Why
- improve code quality
- use `$ref` sugar as specified in CONTRIBUTING
- fix #7638

### DIFF
--- a/packages/client/src/components/follow-button.vue
+++ b/packages/client/src/components/follow-button.vue
@@ -85,11 +85,6 @@ async function onClick() {
 					userId: props.user.id
 				});
 				hasPendingFollowRequestFromYou.value = false;
-			} else if (props.user.isLocked) {
-				await os.api('following/create', {
-					userId: props.user.id
-				});
-				hasPendingFollowRequestFromYou.value = true;
 			} else {
 				await os.api('following/create', {
 					userId: props.user.id
--- a/packages/client/src/components/follow-button.vue
+++ b/packages/client/src/components/follow-button.vue
@@ -43,9 +43,9 @@ const props = withDefaults(defineProps<{
 	large: false,
 });
 
-const isFollowing = ref(props.user.isFollowing);
-const hasPendingFollowRequestFromYou = ref(props.user.hasPendingFollowRequestFromYou);
-const wait = ref(false);
+let isFollowing = $ref(props.user.isFollowing);
+let hasPendingFollowRequestFromYou = $ref(props.user.hasPendingFollowRequestFromYou);
+let wait = $ref(false);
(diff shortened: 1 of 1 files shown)


pixeldesu (approved):
👍 LGTM

One thing that should be adjusted however is removing the `ref` import, which is now unused.

syuilo:
🙏

PR #8791, opened 2022-06-07
refactor: use awaitAll to reduce duplication

Johann150:
# What
Refactor the `/api/users/stats` endpoint to reduce code duplication. By making use of `awaitAll`, most of the property names only need to be written once instead of twice.

# Why
improve code quality


pixeldesu (approved):
👍 LGTM

syuilo:
🙏

PR #8793, opened 2022-06-07
fix(client): only enable hotkeys for logged in users

### LINKED ISSUE #8491: ログインしていなくてもログインページで投稿フォームが表示される
## 💡 Summary

タイトルの通り
https://github.com/misskey-dev/misskey/issues/5851
と同様ですが…

## 🥰 Expected Behavior
（仕様でなければ）出ない
## 🤬 Actual Behavior
出る
## 📝 Steps to Reproduce

0. ログイン済みであればログアウトする
1. ログインページで n または p を押す

## 📌 Environment

Misskey version: 12.110.0
Your browser: Chromium, Safari

pixeldesu:
# What

This PR disables hotkeys for non-logged in users.

# Why

Most of them aren't really useful for site visitors anyway and the post dialog does not work if a user is not logged in.

Fixes #8491

### DIFF
--- a/packages/client/src/init.ts
+++ b/packages/client/src/init.ts
@@ -335,16 +335,16 @@ for (const plugin of ColdDeviceStorage.get('plugins').filter(p => p.active)) {
 	});
 }
 
+const hotkeys = {
+	'd': (): void => {
+		defaultStore.set('darkMode', !defaultStore.state.darkMode);
+	},
+	's': search,
+};
+
 if ($i) {
-	// shortcut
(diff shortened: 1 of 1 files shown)


Johann150:
Since the search API is available to all unauthenticated users it would make sense to also allow it for unauthenticated users. And toggling light/dark mode also makes sense for unauthenticated users.

Maybe better to use `pleaseLogin` at the start of the post key handler or immediately returning from that if not signed in.

Johann150 (approved):

syuilo:
👍

PR #8797, opened 2022-06-08
fix: use autocomplete=new-password for signin

### LINKED ISSUE #8795: Email / password fields are not auto-completed correctly on Sign-up form
## 💡 Summary

Sign-up form has fields of email address and password (and their confirmation), which is not auto-completed from 1Password or other autofill correctly because value of `[autocomplete]` attribute is invalid.

image

## 🥰 Expected Behavior

Value of `[autocomplete]` should be `new-password`, and should be auto-completed correctly from password managers.

## 🤬 Actual Behavior

Value of `[autocomplete]` is random decimal, and not auto-completed correctly. 

## 📝 Steps to Reproduce

1. Open the sign-up form.
2. See the DOM.

## 📌 Environment

https://misskey.io/

Johann150:
# What
Fix autocomplete for signin.

# Why
fix #8795


syuilo:
🙏

PR #7518, opened 2021-05-17
Add image description support

nullobsi:
## Summary
Resolve #7145
Resolve #6860

Adds support for both federation and addition of image descriptions using the ALT and TITLE properties on img.
Adds support to attach image descriptions to drive files and store them under comment in the database.

PR #8802, opened 2022-06-08
fix: alt text for videos

Johann150:
# What
Properly display alt text for videos.

I assume this was overlooked in #7518 since that was focused on image descriptions. But the federation and API is in place to easily make this work for videos as well.

# Why
improve accessibility


syuilo:
🙏

PR #8833, opened 2022-06-13
enhance: show software in instance ticker

### LINKED ISSUE #8832: show what software an instance is using
## Summary

Show what software another instance is running. This can be useful because some ActivityPub software does not support some features, e.g. Mastodon does not support emoji reactions or some MFM markup, Pleroma does not support custom emoji reactions etc.

This could be done by showing the software name in the instance ticker bar.

Johann150:
# What
Show the software of an instance in the instance ticker.

# Why
fix #8832

# Additional info
The API already provides this information so no changes to the backend are necessary.

### DIFF
--- a/packages/client/src/components/instance-ticker.vue
+++ b/packages/client/src/components/instance-ticker.vue
@@ -2,6 +2,7 @@
 <div class="hpaizdrt" :style="bg">
 	<img v-if="instance.faviconUrl" class="icon" :src="instance.faviconUrl"/>
 	<span class="name">{{ instance.name }}</span>
+	<span v-if="instance.softwareName" class="software">({{ instance.softwareName }})</span>
 </div>
 </template>
 
@@ -14,6 +15,7 @@ const props = defineProps<{
 		faviconUrl?: string
 		name: string
 		themeColor?: string
+		softwareName?: string
 	}
 }>();
 
@@ -21,7 +23,8 @@ const props = defineProps<{
 const instance = props.instance ?? {
 	faviconUrl: '/favicon.ico',
 	name: instanceName,
-	themeColor: (document.querySelector('meta[name="theme-color-orig"]') as HTMLMetaElement)?.content
+	themeColor: (document.querySelector('meta[name="theme-color-orig"]') as HTMLMetaElement)?.content,
+	softwareName: 'misskey',
 };
 
 const themeColor = instance.themeColor ?? '#777777';
@@ -34,6 +37,7 @@ const bg = {
 <style lang="scss" scoped>
 .hpaizdrt {
 	$height: 1.1rem;
+	position: relative;
 
 	height: $height;
 	border-radius: 4px 0 0 4px;
@@ -57,12 +61,17 @@ const bg = {
 		height: 100%;
 	}
 
-	> .name {
+	> .name, .software {
 		margin-left: 4px;
 		line-height: $height;
 		font-size: 0.9em;
 		vertical-align: top;
 		font-weight: bold;
+	}
+
+	> .software {
+		position: absolute;
+		right: .3em;
 	}
 }
 </style>
--- a/packages/client/src/components/instance-ticker.vue
+++ b/packages/client/src/components/instance-ticker.vue
@@ -2,7 +2,7 @@
 <div class="hpaizdrt" :style="bg">
 	<img v-if="instance.faviconUrl" class="icon" :src="instance.faviconUrl"/>
 	<span class="name">{{ instance.name }}</span>
-	<span v-if="instance.softwareName" class="software">({{ instance.softwareName }})</span>
+	<span v-if="instance.softwareName" v-tooltip="instance.softwareName + ' ' + instance.softwareVersion" class="software">{{ instance.softwareName }}</span>
 </div>
 </template>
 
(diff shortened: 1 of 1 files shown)


Johann150:
<details>
<summary>screenshot</summary>

image
</details>

I'm unsure over the design. It was suggested that the text does not look great because it does not have a proper background so maybe it would make sense to remove the gradient for the instance theme colour here.

packages/client/src/components/instance-ticker.vue
```diff
@@ -14,14 +15,16 @@ const props = defineProps<{
 		faviconUrl?: string
 		name: string
 		themeColor?: string
+		softwareName?: string
```
Johann150: I noticed there are no semicolons after the types so I did not add one here either. Should they be added?

ThatOneCalculator:
Does the software name discern between forks? For example, Mastodon doesn't support emoji/custom emoji reactions but Fedibird (Mastodon fork) does

Jeder321:
> Does the software name discern between forks? For example, Mastodon doesn't support emoji/custom emoji reactions but Fedibird (Mastodon fork) does

this makes use of software part of the nodeinfo, fedibird doesn't use it to discern itself from upstream mastodon at all
image
glitch-soc, however, does
image
but as you can see, this doesn't show the version string

ThatOneCalculator:
Maybe it could look in `metadata[forks]`? 
image

So it'd appear as `mastodon (fedibird)`

Jeder321:
this would require backend changes if i understand correctly, and also description of `metadata`  is `Free form key value pairs for software specific values. Clients should not rely on any specific key present.` as seen [here](http://nodeinfo.diaspora.software/ns/schema/2.0), not exactly sure if this is a great idea

ThatOneCalculator:
Maybe it could just be if in `metadata`, show the fork name *if* `forks` is present.

Johann150:
As Jeder321 said previously, the NodeInfo specification says:

> **metadata** (object)
> Free form key value pairs for software specific values. Clients should not rely on any specific key present.
-- https://nodeinfo.diaspora.software/schema.html

The last sentence rules out what you want to do in my understanding. So I agree with Jeder321 that this does not sound like a good idea.

It was also shown that e.g. for the glitch fork the software is shown in the version string. Although that might be also useful in cases I think showing the full version string will decrease the usefulness because it will be so much potentially hard to read text, especially on mobile devices.

tamaina:
- ソフトウェア名をそのまま表示するのはダサい
- とりあえずオプションにしてほしい
- ソフトウェアのアイコンが表示されると見栄えが良くなるかも？

ThatOneCalculator:
> ソフトウェアのアイコンが表示されると見栄えが良くなるかも？

I feel like that's already (partially) handled by the instance favicon, it'd be weird to see the same icon twice for a lot of posts. 

> とりあえずオプションにしてほしい

I agree. Having this as an opt-in setting would be nice.

Jeder321:
> ソフトウェア名をそのまま表示するのはダサい

what do you mean?
otherwise, i agree with calc

Johann150:
Showing icons for software would also mean we have to keep a list of known software and their icons, which sounds like it could be a lot of work because anyone could come up with a new AP server software at any time. And maybe they do not even have an icon.

tamaina:
> > ソフトウェア名をそのまま表示するのはダサい
> 
> what do you mean?

It is tacky to display the software name as it is [like the screenshot](https://github.com/misskey-dev/misskey/pull/8833#issuecomment-1154434837).

ダサい (tacky to) https://japaslang.com/?keyword=%E3%83%80%E3%82%B5%E3%81%84

Jeder321:
yeah it looks weird, but tbh i can't really imagine how to show it better without hiding it which would be undesirable

Jeder321:
image
maybe something like this would look better?

tamaina:
>  maybe something like this would look better?

Works well on desktop, but can be cramped on mobile and deck.

Jeder321:
hm, removing parentheses and capitalising first letter seems to be enough? looks better than current one imo
image
but for me "tacky" doesn't say much about what's wrong with it lol, so i have no clue what i am doing

Johann150:
As I suggested above, removing the background gradient is also an option. For example:
image

futchitwo:
If the software name were to be displayed in that position, it would look more natural if the gradient is left in place and the text color is the same color as the user name, etc.

ThatOneCalculator:
> As I suggested above, removing the background gradient is also an option. 

If we were to do this, the banner color should be at 50%-ish opacity, as a full bar like that on every post can be jarring, especially if they're bright colors.

Johann150:
New design change looks like this:

image

If you hover over the software name you will see a tooltip with the software name and also version. This should allow e.g. glitch to be discerned. I thought that the version number can be quite ugly and long so I did not want to include it in the label that is always shown.

image

futchitwo:
Cool tooltip!
ツールチップをインスタンス名のホバーで表示できるようにして、右側のソフトウェア名消しても良いかも
見た目が従来と同じだと「とりあえずオプション」にする必要もなくなる

Johann150:
I think having it displayed immediately is a nice feature because you do not have to explicitly check.

I can see how it may be problematic on smaller UI layouts. But on touch devices where hovering is often not possible, this information would not be available at all.

PR #6696, opened 2020-09-10
Embedded all attachment, renotes and discussion history into rss feed content & improve title

CGsama:
The feed npm module dose not support multiple files, and enclosure format may not parse correct by apps, so that embedding all files into content. Also the email field is required by some app, so add a dummy one, could improved by use user email if have one.

## Summary

Resolve #7004

PR #8834, opened 2022-06-14
enhance: Embedded all attachment, renotes and discussion history into rss feed content & improve title

### LINKED ISSUE #7004: Want enhancement for RSS feed generation
## Summary

I know that the RSS feature is kind of useless in 2020, but there still some old dinosaur (like me) wish to use it. Checked the feed generating code haven't update for years. Wish the feed can contain more note info in it, e.g. all media files, discussion history, and renote content.

CGsama:
# What

The feed npm module dose not support multiple files, and enclosure format may not parse correct by apps, so that embedding all files into content. Also the email field is required by some app, so add a dummy one, could improved by use user email if have one.

The feed also can track back x posts for reply and renote, by using query string e.g. https://xxx/@xxx.atom?history=5 will include 5 most recent notes in the thread

# Why

Resolve #7004

# Additional info (optional)

Change originally made for #6696 at v11, now port to v12 to match file structure, working fine on my instance for years

### DIFF
--- a/packages/backend/src/server/web/feed.ts
+++ b/packages/backend/src/server/web/feed.ts
@@ -13,7 +13,7 @@ export default async function(user: User, threadDepth = 5, history = 20, noteint
 
 	const profile = await UserProfiles.findOneByOrFail({ userId: user.id });
 
-	let searchCriteria = {
+	const searchCriteria = {
 		userId: user.id,
 		visibility: In(['public', 'home']),
 	};
--- a/packages/backend/src/server/web/feed.ts
+++ b/packages/backend/src/server/web/feed.ts
@@ -44,8 +44,8 @@ export default async function(user: User, withAll = false, history = 5) {
 		let contentStr = await noteToString(note, true);
 		let next = note.renoteId ? note.renoteId : note.replyId;
 		let depth = history;
-		while(depth > 0 && next && withAll){
-			let finding = await findById(next);
+		while (depth > 0 && next && withAll) {
+			const finding = await findById(next);
 			contentStr += finding.text;
 			next = finding.next;
 			depth -= 1;
@@ -60,41 +60,41 @@ export default async function(user: User, withAll = false, history = 5) {
 		});
 	}
 
-	async function noteToString(note, isTheNote = false){
-		const author = isTheNote ? null : await Users.findOneBy({id: note.userId});
+	async function noteToString(note, isTheNote = false) {
+		const author = isTheNote ? null : await Users.findOneBy({ id: note.userId });
 		let outstr = author ? `${author.name}(@${author.username}@${author.host ? author.host : config.host}) ${(note.renoteId ? 'renotes' : (note.replyId ? 'replies' : 'says'))}: <br>` : '';
 		const files = note.fileIds.length > 0 ? await DriveFiles.findBy({
 			id: In(note.fileIds),
 		}) : [];
 		let fileEle = '';
-		for (const file of files){
-			if(file.type.startsWith('image/')){
+		for (const file of files) {
+			if (file.type.startsWith('image/')) {
 				fileEle += ` <br><img src="${DriveFiles.getPublicUrl(file)}">`;
-			}else if(file.type.startsWith('audio/')){
+			} else if (file.type.startsWith('audio/')) {
 				fileEle += ` <br><audio controls src="${DriveFiles.getPublicUrl(file)}" type="${file.type}">`;
-			}else if(file.type.startsWith('video/')){
+			} else if (file.type.startsWith('video/')) {
 				fileEle += ` <br><video controls src="${DriveFiles.getPublicUrl(file)}" type="${file.type}">`;
-			}else{
+			} else {
 				fileEle += ` <br><a href="${DriveFiles.getPublicUrl(file)}" download="${file.name}">${file.name}</a>`;
 			}
 		}
 		outstr += `${note.cw ? note.cw + '<br>' : ''}${note.text || ''}${fileEle}`;
-		if(isTheNote){
-			outstr += ` <span class="${(note.renoteId ? 'renote_note' : (note.replyId ? 'reply_note' : 'new_note'))} ${(fileEle.indexOf('img src') != -1 ? 'with_img' : 'without_img')}"></span>`;
+		if (isTheNote) {
+			outstr += ` <span class="${(note.renoteId ? 'renote_note' : (note.replyId ? 'reply_note' : 'new_note'))} ${(fileEle.indexOf('img src') !== -1 ? 'with_img' : 'without_img')}"></span>`;
 		}
 		return outstr;
 	}
 
-	async function findById(id){
+	async function findById(id) {
 		let text = '';
 		let next = null;
-		const findings = await Notes.find({where: {id: id, visibility: In(['public', 'home'])}, order: { createdAt: -1 }, take: 20});
-		for (const aFind of findings){
+		const findings = await Notes.find({ where: { id: id, visibility: In(['public', 'home']) }, order: { createdAt: -1 }, take: 20 });
+		for (const aFind of findings) {
 			text += `<hr>`;
 			text += await noteToString(aFind);
 			next = aFind.renoteId ? aFind.renoteId : aFind.replyId;
 		}
-		return {text: text, next: next};
+		return { text: text, next: next };
 	}
 
 	return feed;
--- a/packages/backend/src/server/web/feed.ts
+++ b/packages/backend/src/server/web/feed.ts
@@ -49,8 +49,22 @@ export default async function(user: User, history = 5, noteintitle = false) {
 			depth -= 1;
 		}
 
-		let title = `${author.name} ${(note.renoteId ? 'renotes' : (note.replyId ? 'replies' : 'says'))}`;
-		title += noteintitle ? `: ${note.cw ? note.cw : (note.text ? note.text : 'post a new note')}` : '';
+		let title = `${author.name} `;
+		if (note.renoteId) {
+			title += 'renotes';
+		} else if (note.replyId) {
+			title += 'replies';
+		} else {
+			title += 'says';
+		}
+		if (noteintitle) {
+			const content = note.cw ?? note.text;
+			if (content) {
+				title += `: ${content}`;
+			} else {
+				title += 'something';
+			}
+		}
 
 		feed.addItem({
 			title: title.replace(/[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]/g, '').substring(0,100),
--- a/packages/backend/src/server/web/feed.ts
+++ b/packages/backend/src/server/web/feed.ts
@@ -2,10 +2,9 @@ import { Feed } from 'feed';
 import { In, IsNull } from 'typeorm';
 import config from '@/config/index.js';
 import { User } from '@/models/entities/user.js';
-import { Users, Notes, DriveFiles, UserProfiles } from '@/models/index.js';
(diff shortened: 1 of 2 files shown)


packages/backend/src/server/web/feed.ts
```diff
@@ -2,20 +2,22 @@ import { Feed } from 'feed';
...
-import { Notes, DriveFiles, UserProfiles, Users } from '@/models/index.js';
+import { Users, Notes, DriveFiles, UserProfiles } from '@/models/index.js';
+import { In, IsNull } from 'typeorm';
 
-export default async function(user: User) {
+export default async function(user: User, withAll = false, history = 5) {
 	const author = {
 		link: `${config.url}/@${user.username}`,
-		name: user.name || user.username,
+		email: `${user.username}@${config.host}`,
```
Johann150, line 10: A fedi account is not an email address, so this seems a bit weird. People might try to send email to these addresses which would be inappropriate.

packages/backend/src/server/web/feed.ts
```diff
@@ -27,7 +29,7 @@ export default async function(user: User) {
 		title: `${author.name} (@${user.username}@${config.host})`,
 		updated: notes[0].createdAt,
 		generator: 'Misskey',
-		description: `${user.notesCount} Notes, ${profile.ffVisibility === 'public' ? user.followingCount : '?'} Following, ${profile.ffVisibility === 'public' ? user.followersCount : '?'} Followers${profile.description ? ` · ${profile.description}` : ''}`,
+		description: `${user.notesCount} Notes, ${user.followingCount} Following, ${user.followersCount} Followers${profile.description ? ` · ${profile.description}` : ''}`,
```
Johann150: Removing the visibility check is most definitely not a good idea. This is not okay.

packages/backend/src/server/web/feed.ts
```diff
@@ -39,20 +41,61 @@ export default async function(user: User) {
...
+		while (depth > 0 && next && withAll) {
+			const finding = await findById(next);
+			contentStr += finding.text;
+			next = finding.next;
+			depth -= 1;
+		}
 
 		feed.addItem({
-			title: `New note by ${author.name}`,
+			title: `${author.name} ${(note.renoteId ? 'renotes' : (note.replyId ? 'replies' : 'says'))}: ${note.cw ? note.cw : (note.text ? note.text : 'post a new note')}`.replace(/[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]/g, '').substring(0,100),
```
Johann150: Using so many ternary expressions makes this code hard to read. Maybe it is better to generate this text in a separate variable.

I am also not sure why you include the full text of the note in the title already. That is what the `content` is for.

packages/backend/src/server/web/feed.ts
```diff
@@ -39,20 +41,61 @@ export default async function(user: User) {
...
 			link: `${config.url}/notes/${note.id}`,
 			date: note.createdAt,
-			description: note.cw || undefined,
-			content: note.text || undefined,
-			image: file ? DriveFiles.getPublicUrl(file) || undefined : undefined,
+			description: note.cw ? note.cw.replace(/[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]/g, '') : undefined,
+			content: `${contentStr.replace(/[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]/g, '')}`
```
Johann150: There is no need to use a template string here.
```suggestion
			content: contentStr.replace(/[\x00-\x08\x0b\x0c\x0e-\x1f\x7f]/g, '')
```

packages/backend/src/server/web/index.ts
```diff
@@ -198,7 +198,7 @@ router.get('/@:user.atom', async ctx => {
 
 // RSS
 router.get('/@:user.rss', async ctx => {
-	const feed = await getFeed(ctx.params.user);
+	const feed = await getFeed(ctx.params.user, ctx.query.history ? true : false, parseInt(ctx.query.history) ? parseInt(ctx.query.history) : 5);
```
Johann150: same as above

packages/backend/src/server/web/index.ts
```diff
@@ -210,7 +210,7 @@ router.get('/@:user.rss', async ctx => {
 
 // JSON
 router.get('/@:user.json', async ctx => {
-	const feed = await getFeed(ctx.params.user);
+	const feed = await getFeed(ctx.params.user, ctx.query.history ? true : false, parseInt(ctx.query.history) ? parseInt(ctx.query.history) : 5);
```
Johann150: same as above

packages/backend/src/server/web/index.ts
```diff
@@ -173,20 +173,20 @@ router.get('/api.json', async ctx => {
...
 	});
 
-	return user && await packFeed(user);
+	return user && await packFeed(user, withAll, history);
 };
 
 // Atom
 router.get('/@:user.atom', async ctx => {
-	const feed = await getFeed(ctx.params.user);
+	const feed = await getFeed(ctx.params.user, ctx.query.history ? true : false, parseInt(ctx.query.history) ? parseInt(ctx.query.history) : 5);
```
Johann150: I think the history parameter should be sanitized. I also think there should be an upper limit to not make this exploitable in a DoS attack.

```suggestion
	let history = parseInt(ctx.query.history, 10);
	if (isNaN(history) || history <= 0 || history > 10) {
		history = 5;
	}
	const feed = await getFeed(ctx.params.user, !!history, history);
```

Maybe consider if the 2nd parameter is necessary in this case. Or maybe also do this parsing and validating in the `getFeed` function since it is reused code.

packages/backend/src/server/web/feed.ts
```diff
@@ -2,20 +2,22 @@ import { Feed } from 'feed';
 import { In, IsNull } from 'typeorm';
 import config from '@/config/index.js';
 import { User } from '@/models/entities/user.js';
-import { Notes, DriveFiles, UserProfiles, Users } from '@/models/index.js';
+import { Users, Notes, DriveFiles, UserProfiles } from '@/models/index.js';
+import { In, IsNull } from 'typeorm';
```
Johann150: Duplicate import, this is already imported in line 2.

packages/backend/src/server/web/feed.ts
```diff
@@ -2,20 +2,22 @@ import { Feed } from 'feed';
 import { In, IsNull } from 'typeorm';
 import config from '@/config/index.js';
 import { User } from '@/models/entities/user.js';
-import { Notes, DriveFiles, UserProfiles, Users } from '@/models/index.js';
+import { Users, Notes, DriveFiles, UserProfiles } from '@/models/index.js';
```
Johann150: Not sure why you reordered these imports.

packages/backend/src/server/web/feed.ts
```diff
@@ -39,20 +41,61 @@ export default async function(user: User) {
...
+	async function findById(id) {
+		let text = '';
+		let next = null;
+		const findings = await Notes.find({ where: { id: id, visibility: In(['public', 'home']) }, order: { createdAt: -1 }, take: 20 });
+		for (const aFind of findings) {
+			text += `<hr>`;
+			text += await noteToString(aFind);
+			next = aFind.renoteId ? aFind.renoteId : aFind.replyId;
+		}
+		return { text: text, next: next };
```
Johann150:
```suggestion
		return { text, next };
```

packages/backend/src/server/web/feed.ts
```diff
@@ -39,20 +41,61 @@ export default async function(user: User) {
...
+
+	async function findById(id) {
+		let text = '';
+		let next = null;
+		const findings = await Notes.find({ where: { id: id, visibility: In(['public', 'home']) }, order: { createdAt: -1 }, take: 20 });
+		for (const aFind of findings) {
+			text += `<hr>`;
+			text += await noteToString(aFind);
+			next = aFind.renoteId ? aFind.renoteId : aFind.replyId;
+		}
```
Johann150: The `id` column should be unique so what are the `take: 20` and loop for?

packages/backend/src/server/web/feed.ts
```diff
@@ -2,20 +2,22 @@ import { Feed } from 'feed';
...
+		name: user.name || user.username
 	};
 
 	const profile = await UserProfiles.findOneByOrFail({ userId: user.id });
 
 	const notes = await Notes.find({
 		where: {
 			userId: user.id,
-			renoteId: IsNull(),
+			//renoteId: IsNull(),
```
Johann150: If you do not want to filter this, just remove the line instead of commenting it out.

Johann150 (dismissed):

packages/backend/src/server/web/feed.ts
CGsama: the problem is some rss software requires an email field, otherwise will think as invalid feed, thats why i put here

packages/backend/src/server/web/feed.ts
CGsama: similar reason to the email, some app will just show the title but ignore content, so user will see "xxx says:" again and again, e.g. rss reader build in misskey, guess parameterize the include text would be better

packages/backend/src/server/web/feed.ts
CGsama: true, the check wasn't there in v11

packages/backend/src/server/web/feed.ts
CGsama: somethine was added after v12 lol

packages/backend/src/server/web/index.ts
CGsama: I would say 30 is a resonable max number? some thread may getting very long

packages/backend/src/server/web/feed.ts
CGsama: have a noteintitle query param, default will only show name and renote/reply/say

packages/backend/src/server/web/feed.ts
```diff
@@ -39,20 +39,64 @@ export default async function(user: User) {
...
+		let depth = history;
+		while (depth > 0 && next) {
+			const finding = await findById(next);
+			contentStr += finding.text;
+			next = finding.next;
+			depth -= 1;
+		}
+
+		let title = `${author.name} ${(note.renoteId ? 'renotes' : (note.replyId ? 'replies' : 'says'))}`;
+		title += noteintitle ? `: ${note.cw ? note.cw : (note.text ? note.text : 'post a new note')}` : '';
```
Johann150: This still uses a lot of ternary expressions and is hard to read. I think something like this is easier to read:

```suggestion
		let title = `${author.name} `;
		if (note.renoteId) {
			title += 'renotes';
		} else if (note.replyId) {
			title += 'replies';
		} else {
			title += 'says';
		}
		if (noteintitle) {
			const content = note.cw ?? note.text;
			if (content) {
				title += `: ${content}`;
			} else {
				title += 'something';
			}
		}
```

CGsama:
query example: https://xxx.xxx/@xxx.atom?thread=3&history=5&noteintitle&norenotes&noreplies

thread: how deep a thread can go (reply history)
history: how many notes totaly
noteintitle: put note content in title
norenotes: no renotes in feed
noreplies: no noreplies in feed

CGsama:
anything needs to change/fix?

CGsama:
Any updates here? Thx

CGsama:
rebase to latest

tamaina (approved):

tamaina:
Close because it has been moved to FeedService and conflicted heavily. 
(I wonder how @syuilo can do such a terrible thing)

PR #8837, opened 2022-06-14
enhance: use tag for federating quotes

### LINKED ISSUE #8722: Make quote renotes work with other Fediverse software
## Summary

Quote-renotes are shown as a reply to the quoted note in Misskey, but users of Mastodon and Pleroma don't see the reply and don't get a notification. One of the Mastodon devs suggested that Misskey should change the activity it sends. I don't know much about this stuff, so I'll just link the post here: https://github.com/mastodon/mastodon/issues/18473#issuecomment-1133799674

Johann150:
# What
Use a `Link` tag to federate quotes instead of the custom `_misskey_quote` property.

# Why
related to #8722

# Additional info
This uses the same namespace URL that the `_misskey_quote` property uses in JSON-LD for the `rel` attribute.

With this change the deprecated properties will still be sent so this is for now fully backwards compatible. It should be reviewed at a later date whether the `_misskey_quote` property (and maybe the `quoteUrl` property as well) are still necessary or should be removed to reduce network traffic.

### DIFF
--- a/packages/backend/src/remote/activitypub/models/tag.ts
+++ b/packages/backend/src/remote/activitypub/models/tag.ts
@@ -24,10 +24,20 @@ export function extractQuoteUrl(tags: IObject | IObject[] | null | undefined): s
 		.filter(isLink)
 		.filter(link => link.mediaType === 'application/activity+json');
 
+	// Multiple quotes are not supported, try to reduce to only 1.
 	if (quotes.length > 1) {
-		// There are multiple quotes? Not supported.
 		// Check if there maybe is one intended for Misskey.
 		quotes = quotes.filter(link => toArray(link.rel).includes('https://misskey-hub.net/ns#_misskey_quote'));
+	}
+	if (quotes.length > 1) {
+		// Deduplicate by href value.
+		const hrefs = new Set();
+		quotes = quotes.filter(link => {
+			if (hrefs.has(link.href)) return false;
+
+			hrefs.add(link.href);
+			return true;
+		});
 	}
 
 	if (quotes.length === 1) {
--- a/packages/backend/src/remote/activitypub/models/tag.ts
+++ b/packages/backend/src/remote/activitypub/models/tag.ts
@@ -1,5 +1,5 @@
 import { toArray } from '@/prelude/array.js';
-import { IObject, isHashtag, IApHashtag } from '../type.js';
+import { IObject, isHashtag, IApHashtag, isLink, ILink } from '../type.js';
 
 export function extractApHashtags(tags: IObject | IObject[] | null | undefined) {
 	if (tags == null) return [];
@@ -15,4 +15,18 @@ export function extractApHashtags(tags: IObject | IObject[] | null | undefined) 
 export function extractApHashtagObjects(tags: IObject | IObject[] | null | undefined): IApHashtag[] {
 	if (tags == null) return [];
 	return toArray(tags).filter(isHashtag);
+}
+
+export function extractQuoteUrl(tags: IObject | IObject[] | null | undefined): string | null {
+	if (tags == null) return null;
+	const quotes: ILink[] = toArray(tags)
+		.filter(isLink)
+		.filter((link) => toArray(tag.rel).includes('https://misskey-hub.net/ns#_misskey_quote'));
+
+	if (quotes.length === 1) {
+		return quotes.href;
+	} else {
+		// either multiple quotes (unsupported) or no quote
+		return null;
+	}
 }
--- a/packages/backend/src/remote/activitypub/models/tag.ts
+++ b/packages/backend/src/remote/activitypub/models/tag.ts
@@ -19,9 +19,16 @@ export function extractApHashtagObjects(tags: IObject | IObject[] | null | undef
 
 export function extractQuoteUrl(tags: IObject | IObject[] | null | undefined): string | null {
 	if (tags == null) return null;
-	const quotes: ILink[] = toArray(tags)
+
+	let quotes: ILink[] = toArray(tags)
 		.filter(isLink)
-		.filter((link) => toArray(tag.rel).includes('https://misskey-hub.net/ns#_misskey_quote'));
+		.filter(link => link.mediaType === 'application/activity+json');
+
+	if (quotes.length > 1) {
+		// There are multiple quotes? Not supported.
+		// Check if there maybe is one intended for Misskey.
+		quotes = quotes.filter(link => toArray(link.rel).includes('https://misskey-hub.net/ns#_misskey_quote'));
+	}
 
 	if (quotes.length === 1) {
 		return quotes.href;
--- a/packages/backend/src/remote/activitypub/renderer/note.ts
+++ b/packages/backend/src/remote/activitypub/renderer/note.ts
@@ -111,10 +111,8 @@ export default async function renderNote(note: Note, dive = true, isTalk = false
 		...apemojis,
 	];
 
-	const attachment = files.map(renderDocument);
-
 	if (quote) {
-		attachment.push({
+		tag.push({
 			type: 'Link',
 			rel: 'https://misskey-hub.net/ns#_misskey_quote',
 			mediaType: 'application/activity+json',
@@ -160,7 +158,7 @@ export default async function renderNote(note: Note, dive = true, isTalk = false
 		to,
 		cc,
 		inReplyTo,
-		attachment,
+		attachment: files.map(renderDocument),
 		sensitive: note.cw != null || files.some(file => file.isSensitive),
 		tag,
 		...asPoll,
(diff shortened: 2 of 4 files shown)


packages/backend/src/remote/activitypub/renderer/note.ts
```diff
@@ -111,6 +111,16 @@ export default async function renderNote(note: Note, dive = true, isTalk = false
 		...apemojis,
 	];
 
+	if (quote) {
+		tag.push({
+			type: 'Link',
+			rel: 'https://misskey-hub.net/ns#_misskey_quote',
```
ThatOneCalculator, line 117: Seems to be a dead/invalid link? 
image

mei23:
attachment だけは意味的にありえないと思う。
というか他人の投稿を添付した扱いになるのは困る。

あと、`_misskey_quote` がダメな理由がわからない。
`quoteUrl` は ActivityStream を勝手に拡張しているかもしれないけど、`_misskey_quote` は一応 valid なはず。

noellabo:
うん、attachmentではないよね。

ちなみにFedibirdの引用はquoteUrlからquoteUriに変更してこうなってる。
```
@context: [{
  fedibird: "http://fedibird.com/ns#",
  quoteUri: "fedibird:quoteUri"
}],
quoteUri: "https://misskey.io/notes/91zs39f68n",
```

mei23:
これらは全て valid であって 独自拡張 でもあるので、どれかに寄せる理由は何？という感じなんだわ。
`"http://fedibird.com/ns#quoteUri": ...`
`"https://misskey-hub.net/ns#_misskey_quote": ...`
`tag rel="https://misskey-hub.net/ns#_misskey_quote"`

mei23:
そもそも `quoteUri`, `_misskey_quote` は invalid という誤認を元に話を進めきているのが、変な感じ。
https://github.com/mastodon/mastodon/issues/18473#issuecomment-1133874921
↓
https://github.com/misskey-dev/misskey/issues/8722
↓

Johann150:
@mei23 The discussion is taking place in the linked issue, why do you not participate there? Just complaining to me is not that helpful.

> そもそも quoteUri, _misskey_quote は invalid という誤認を元に話を進めきているのが、変な感じ。

... as I said please read the issue, that topic is covered there and you can comment on it there:

> I'm not sure why the current implementation seems to be completely out of the picture, seeing that it is seemingly starting to be adopted by other ActivityPub implementations (Fedibird, Pleroma, as I also laid out on the Mastodon issue).
-- https://github.com/misskey-dev/misskey/issues/8722#issuecomment-1134008835

Johann150:
I am also not sure why you are complaining about that here because this pull request does not remove or modify anything related to `quoteUrl` and `_misskey_quote`.

mei23:
議論はIssueの方でするけど、この Pull Request が Issueのどこで合意を経て `tag` にしたのか `attachment` にしたのかがわからない。

mei23:
> I am also not sure why you are complaining about that here because this pull request does not remove or modify anything related to quoteUrl and _misskey_quote.

Because it was written here that it will be removed later
> the _misskey_quote property (and maybe the quoteUrl property as well) should be removed at a later time.

mei23:
変える必要がなさそうで仕様も決まってないならこのプルリクは何をするものなの？
Draftでもないならもうこの仕様にしようとしていることじゃないの？
とかすごく納得がいかないわ

mei23:
https://github.com/misskey-dev/misskey/pull/8837#issuecomment-1166894145

とりあえず、Misskeyの互換用のも変えておいたほうが良さそう
`https://www.w3.org/ns/activitystreams#quoteUrl`
↓
`http://fedibird.com/ns#quoteUri`

rinsuki:
そもそもquoteUrlはfedibirdじゃない別のMastodonのfork用だったはず (今でも生きてる鯖があるかは不明)

packages/backend/src/remote/activitypub/renderer/note.ts
tamaina: no problem

mei23: We are using URIs to make the namespace unique. So there is no problem if you can't access it.

ThatOneCalculator: Ah, got it

Johann150:
> Because it was written here that it will be removed later
>> the _misskey_quote property (and maybe the quoteUrl property as well) should be removed at a later time.

I have updated the description.

packages/backend/src/remote/activitypub/renderer/note.ts
Johann150: For completeness: This is as specified for ActivityPub vocabulary:
> ### `rel`
> A link relation associated with a [Link](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-link). The value MUST conform to both the [[HTML5](https://www.w3.org/TR/html5/)] and [[RFC5988](https://datatracker.ietf.org/doc/html/rfc5988#section-4)] "link relation" definitions.
-- https://www.w3.org/TR/activitystreams-vocabulary/#dfn-rel

The "link relation" definition from RFC 5988 allows two relation types, relevant is this one:

> ### 4.2. Extension Relation Types
> Applications [...] can use an extension relation type, which is a URI [...]. Although the URI can point to a resource that contains a definition of the semantics of the relation type, clients SHOULD NOT automatically access that resource [...].
-- https://datatracker.ietf.org/doc/html/rfc5988#section-4.2

Johann150:
I have added deduplication based on the `href` field, as this implementation parses both the `tag` and the `attachment` field and in the issue it was suggested that the object may be added to both.

I have also reverted back to using `tag` as it was suggested as the "technically correct" solution.

packages/backend/src/remote/activitypub/models/tag.ts
```diff
@@ -16,3 +16,34 @@ export function extractApHashtagObjects(tags: IObject | IObject[] | null | undef
...
+			return true;
+		});
+	}
+
+	if (quotes.length === 1) {
+		return quotes.href;
+	} else {
+		// either multiple quotes (unsupported) or no quote
+		return null;
+	}
```
silverpill, line 48: Other servers may allow multiple quotes per post. Perhaps it would be better to take the first link from the array if `quotes.length >= 1`?

packages/backend/src/remote/activitypub/models/tag.ts
```diff
@@ -16,3 +16,34 @@ export function extractApHashtagObjects(tags: IObject | IObject[] | null | undef
 	if (tags == null) return [];
 	return toArray(tags).filter(isHashtag);
 }
+
+export function extractQuoteUrl(tags: IObject | IObject[] | null | undefined): string | null {
+	if (tags == null) return null;
+
+	let quotes: ILink[] = toArray(tags)
+		.filter(isLink)
+		.filter(link => link.mediaType === 'application/activity+json');
```
trwnh, line 25: Is this step necessary? `mediaType` isn't required for full semantic understanding. It seems adequate to check only for Links, which ideally have the specified `rel` that you check for immediately below this.

Consider the following JSON:

```json
"tag": [
  {
    "type": "Link",
    "href": "URI",
    "name": "RE: URI",
    "rel": "https://misskey-hub.net/ns#_misskey_quote"
  }
]
```

The current code would ignore this even though it is clearly intended as a misskey quote.

trwnh: One other thing worth pointing out is that [on the associated FEP, I proposed](https://codeberg.org/fediverse/fep/pulls/13#issuecomment-601588) using `rel` instead of `mediaType` as a less ambiguous way of detecting when a Link refers to an ActivityStreams object. Something like this:

```json
"tag": [
  {
    "type": "Link",
    "href": "URI",
    "name": "RE: URI",
    "rel": ["https://www.w3.org/ns/activitystreams", "https://misskey-hub.net/ns#_misskey_quote"]
  }
]
```

Given that the FEP is for "Object Links"

trwnh: Logically I would restructure the flow of this function along the following lines:

1) Check if there's a Link with link-rel of _misskey_quote. If yes, stop.
2) ~~Check if there are any Links with link-rel of activitystreams. If yes, parse for `name` within `content`. If a match is found, add a quote of that object (and optionally replace `name` in `content` with a link targeting `href`).~~

trwnh: Update on the FEP again: given that clients MUST specify `Accept: application/ld+json; profile="https://www.w3.org/ns/activitystreams"` it might make sense to use this `mediaType` instead of `application/activity+json`, since it is possible that an ActivityPub server might exist where `application/activity+json` is not supported (it is only a SHOULD and not a MUST).

Johann150:

Issue #8806, opened 2022-06-09
Use remote versions of follow/reply/renote when not logged in


soupglasses:
When you are browsing on a remote instance `remote.tld/notes/...` without being logged in and you see a note you would like to interact with (reply, renote, or follow), you will currently get a "Please log in" pop up to log into `remote.tld`.

This is not intuitive, as very few people are going to be logged out of their own home server instance, and it is much more likely that those who are not logged in are coming from a different home server.

I propose a similar solution to how Mastodon handles this with remote follows and remote interactions. When you are not logged in to the instance, and use any of the buttons (renote/follow/reply), prompt for `username@domain.tld` and redirect to the other home server.

Johann150:
Regarding remote follows, this is a duplicate of #8114 and/or #2518. Also related: #2217

Johann150:
Misskey can already receive this kind of interaction using the `authorize-follow` client route. What we would have to implement is for the client to be able to determine the interaction URL that the user should be forwarded to.

This can in general be determined using webfinger and looking for the item with `rel` property set to `http://ostatus.org/schema/1.0/subscribe`. Misskey also already provides this value for local accounts.

I'm not sure if webfinger always has the CORS headers set so I assume that the client can not do this entirely on its own and a new API endpoint would need to be added to carry out the webfinger request and return the `template` URL.

Johann150:
I think we should add a new API endpoint `/api/ap/remote-interact` or something like that which does the webfinger part. In the future we might want to cache that, but having the API endpoint provides enough abstraction to allow us to change that in the future.

You pass it an `id` or URL similar to `/api/ap/show` as well as the instance name of the user that wants to interact with it remotely. It returns the URL you have to redirect the client to.

When that is done we can see where this can best be added into the client.

PR #8838, opened 2022-06-14
refactor: simplify ap/show with DbResolver

Johann150:
# What
1. Using the existing code in DbResolver we can avoid separate code for parsing the URIs in this endpoint.
2. Use the signed in user account for packing Notes and Users.

# Why
1. Reduce code duplication and improve maintainability.
2. Use proper visibility checks and allow e.g. followers to see follower only notes.

I moved this to a separate pull request because I would like to reuse the code for implementing #8806.

# Additional info
This commit was extracted from #8641 as a separate change (mostly 8ddc4a15515da8db9b8c1183918ebda71ec9543f) as I found this should be refactored regardless of that pull request.

As a side effect of moving those changes out to a separate commit, the `try...catch` statement in `mergePack` is introduced because of changes that will be made by #8641.


syuilo:
🆒

PR #8904, opened 2022-06-29
 Prevent access to user pages when not logged in [v2]

### LINKED ISSUE #8377: User settings shouldn't be accessible when logged out

Johann150:
# What
Prevent non-logged in users from accessing parts of the website that require a user object or are in general not supposed to be visible to off-site users.

# Why
fix #8377
close #8716


syuilo:
👍🏻

PR #8907, opened 2022-06-29
fix: restore menu entries

Johann150:
# What
Add ability to determine the initial tab for `/my/notifications` using URL hash.

Add back the menu entries for mentions, direct notes, federation, custom emojis and the scratchpad that were removed in commit 0f1c0a42a253dcc5758e4811aaa116708d482305.

# Why
> I think some of the menu items you removed here, especially mentions, direct notes and federation, are commonly used. So I think they should be reintroduced.
-- https://github.com/misskey-dev/misskey/commit/0f1c0a42a253dcc5758e4811aaa116708d482305#commitcomment-77260033

I also noticed there was no Changelog entry for these user-facing changes so I thought that it was the intention to add them back.

### DIFF
--- a/packages/client/src/components/launch-pad.vue
+++ b/packages/client/src/components/launch-pad.vue
@@ -62,7 +62,7 @@ const modal = $ref<InstanceType<typeof MkModal>>();
 
 const menu = defaultStore.state.menu;
 
-const items = Object.keys(menuDef).filter(k => !menu.includes(k)).map(k => menuDef[k]).filter(def => def.show == null ? true : def.show).map(def => ({
+const items = Object.keys(menuDef).filter(k => !menu.includes(k)).map(k => menuDef[k]).filter(def => (def.show ?? true) && !(def.optional ?? false)).map(def => ({
 	type: def.to ? 'link' : 'button',
 	text: i18n.ts[def.title],
 	icon: def.icon,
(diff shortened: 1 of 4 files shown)


ThatOneCalculator:
How about emojis? It's great for ppl who wanna ~~steal~~ borrow other instance's emotes

syuilo:
The change is to solve the problem of [too many menu items](https://github.com/misskey-dev/misskey/issues/6389), so it will not make much sense when restored.
If there are too many menu items, it will be confusing for users who are not used to them, and for those who are used to them, it will be difficult to find the desired item.

ThatOneCalculator:
I personally think adding back emojis is a good idea. Mentions and direct notes aren't really needed since they're now in notifications

Johann150:
> aren't really needed since they're now in notifications

maybe, but

> for those who are used to them, it will be difficult to find the desired item

- - -
> solve the problem of too many menu items

Maybe we could make some menu entries that can be used if users configure the menu to display them. But otherwise not display them on the launch pad ("More" menu entry) otherwise.

I just think it is not a great idea to remove previously configured menu entries.

ThatOneCalculator:
I think making those optional is good :) LGTM!

ThatOneCalculator:
@Johann150 After more testing on this PR, the notification dot over "more" won't go away unless you add back DMs and mentions, and look at them separately. I think adding emojis and groups back is good, but not DMs or mentions.

ThatOneCalculator:
Given that emojis are now in instance info and groups aren't included in this pr, I don't think it's the best idea.

PR #8930, opened 2022-07-03
feat: styled error screen

ThatOneCalculator:
# What
Makes the error screen look like this:
image

# Why
Looks nicer and less scary for less technical users

# Additional info (optional)
Codepens:
- Error: https://codepen.io/ThatOneCalculator/pen/KKopYJe
- Offline (not implemented): https://codepen.io/ThatOneCalculator/pen/RwMPmpZ

For the error screeen, I can't seem to find a way to make Misskey purposefully error so that I can test this.


ThatOneCalculator:
Service worker code currently doesn't work, some help would be appreciated 🙏

ThatOneCalculator:
~~Service worker code currently doesn't work (just shows default browser offline screen), some help would be appreciated 🙏 (cc @tamaina since I know you've worked on the service worker)~~

I've decided to not include the offline screen in this PR. I've tested and the regular error screen works in practice. I can just put the offline screen in another PR.

syuilo:
cool✨
I think error.css is better than boot.css.

ThatOneCalculator:
Renamed!

syuilo:
🙏🏻

ThatOneCalculator:
@syuilo any reason why this was reverted?

syuilo:
This does not seem to work. How did you verify that it works?

ThatOneCalculator:
I tested it on my instance, and got an error (on accident) and it worked, but maybe I made a mistake. Is there a good way to force an error to happen so I can test it further?

PR #8933, opened 2022-07-04
Update default menu options

ThatOneCalculator:
# What
Changes menu options default

# Why
- Removes featured as that's no longer a page
- More realistic choices for what people will most be using


syuilo:
Removing the featured is fine, but we do not want to change the other items. In particular, we do not want the messaging to stand out as it is not well maintained.
🙏

ThatOneCalculator:
Are there any plans to improve messaging in the future? It's one of my favorite misskey features

PR #8949, opened 2022-07-06
enhance: show recipients of notes with specified visibility

### LINKED ISSUE #8414: 公開範囲:ダイレクトで投稿した時に、宛先を後からわかるようにしてほしい
## Summary

現状は、ダイレクト投稿で送信した後に誰に対してノートを送信したのかわからないので、あとからわかるようにしてほしい。
image
image

Johann150:
# What
- refactor visibility component to a separate file to make it reusable
- rename renote tooltip component to reuse it
- add a tooltip to specified recipients indicator

# Why
fix #8414 

# Additional info
Like for renotes, only the first 10 users will be shown. If there are more, a number indicating how many more users there are will be shown.


packages/client/src/components/visibility.vue
```diff
@@ -0,0 +1,48 @@
...
+			users,
+			count: props.note.visibleUserIds.length,
+			targetElement: specified.value,
+		}, {}, 'closed');
+	});
+}
+</script>
+
+<style lang="scss" scoped>
+.visibility, .localOnly {
```
syuilo, line 44: The class selector for the root element must be a random string or use CSS Modules, otherwise it will affect the same name class of other components even with the scoped attribute.

syuilo: I will fix after merge.

Johann150: Is that not what the `scoped` attribute is for?

syuilo:
🙏🏻

PR #8951, opened 2022-07-06
fix(client): Broken chats

### LINKED ISSUE #8857: Sending a chat doesn't show the message you sent until reloading a page
## 💡 Summary

After sending a chat, the message bubble of what you sent doesn't appear until you reload the page.

## 🥰 Expected Behavior

The message you sent shows as soon as you send it

## :rage: Actual Behavior

https://user-images.githubusercontent.com/44733677/174722742-d8c018dd-7eb7-40a3-91b8-0ba3725e64ca.mp4

## 📝 Steps to Reproduce

1. Chat with someone
2. Send message

## 📌 Environment

Misskey version: 12.111.1-test.1
Your OS: Linux
Your browser: Firefox

ltlapy:
# What & Why

- Append `overflow: scroll` in style
  - Satisfy the condition of pagination prepend
  - fix purging old-message on buttom of the screen is not working
  - Close #8857
- Let pagination prepend work even `getScrollContainer(rootEl.value)` is null
  - Prevent potential bugs like this case

# Additional info (optional)

### DIFF
--- a/packages/client/src/pages/messaging/messaging-room.vue
+++ b/packages/client/src/pages/messaging/messaging-room.vue
@@ -292,6 +292,7 @@ definePageMetadata(computed(() => !fetching ? user ? {
 <style lang="scss" scoped>
 .mk-messaging-room {
 	position: relative;
+	overflow: scroll;
 
 	> .body {
 		.more {
--- a/packages/client/src/components/ui/pagination.vue
+++ b/packages/client/src/components/ui/pagination.vue
@@ -192,21 +192,23 @@ const prepend = (item: Item): void => {
 	if (props.pagination.reversed) {
 		if (rootEl.value) {
 			const container = getScrollContainer(rootEl.value);
-			if (container == null) return; // TODO?
-
-			const pos = getScrollPosition(rootEl.value);
-			const viewHeight = container.clientHeight;
-			const height = container.scrollHeight;
-			const isBottom = (pos + viewHeight > height - 32);
-			if (isBottom) {
-				// オーバーフローしたら古いアイテムは捨てる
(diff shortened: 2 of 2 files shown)


ThatOneCalculator:
New messages come in live BUT it erases the prev message. Testing on https://stop.voring.me

ThatOneCalculator:
Also there seems to be some weird overlapping problems


This is unscrolled.

packages/client/src/pages/messaging/messaging-room.vue
```diff
@@ -292,6 +292,7 @@ definePageMetadata(computed(() => !fetching ? user ? {
 <style lang="scss" scoped>
 .mk-messaging-room {
 	position: relative;
+	overflow: scroll;
```
Johann150, line 295: I think using `overflow-y` here is a better idea. Otherwise browsers will always show an unnecessary scroll bar for horizontal scrolling.

Or maybe use `overflow: auto;` instead to let the browser decide when scroll bars are necessary.

ThatOneCalculator:
The problem of messages being cut off only seems to happen on mobile. Works fine on desktop.

ltlapy:
I think i'm cannot handle with these problems, those are more complicated than i thought.
Thanks to feedbacks, and Sorry for late replying.

ThatOneCalculator:
> I think i'm cannot handle with these problems, those are more complicated than i thought.
> Thanks to feedbacks, and Sorry for late replying.

I don't think it should be too hard to fix, just a tiny bit of positioning. If you re-open this PR I'd be happy to take a look and see if I can make a solution :)

PR #8955, opened 2022-07-07
enhance: autocomplete in messaging

ThatOneCalculator:
# What
Autocomplete in messaging, just like posts.

# Why
Fixes TODO, no reason *not* to have this functionality

# Additional info (optional)
Tested and working on https://stop.voring.me


Johann150:
I think the reason there is no autocomplete in chats is that
- user mentions and hashtags do not really work in chats, so it makes little sense to use them
- emojis are not federated properly (yet), see #8052

ThatOneCalculator:
Fair enough. It's still nice to have for chatting with local users and MFM completion, suppose I can just keep this open until #8052 is resolved :)

PR #8957, opened 2022-07-07
refactor: signup component as composition api

Johann150:
# What
Refactor the signup component to Composition API.

# Why
Composition API refactoring.


syuilo:
🙏🏻

PR #8977, opened 2022-07-09
enhance: read theme color nodeinfo

### LINKED ISSUE #8897: Pull theme-color from nodeinfo instead of HTML/PWA config
## 💡 Summary

Currently, `theme-colour`, used for the banner on posts, [is pulled from](https://github.com/misskey-dev/misskey/blob/develop/packages/backend/src/services/fetch-instance-metadata.ts#L211):
- HTML meta tags on the `/` path
- A PWA-specific `manifest.json` property

These should be considered instance-internal paths and should not be used for federated information

## 🥰 Expected Behavior

Theme-color should have a fallback (or, priority?) in the nodeinfo - `metadata.nodeAccentColor`? (take your pick on the actual property)

Johann150:
# What
Provide the theme color in the nodeinfo and read it from there when fetching metadata of another instance.

# Why
fix #8897


syuilo:
👍🏻

PR #8990, opened 2022-07-13
Collapse long notes

### LINKED ISSUE #8559: option to collapse long notes

Johann150:
# What
Fixes the issues from #8561.

# Why
fix #8559


syuilo:
I tried it and it didn't work, it seems that as soon as collapsed is false, isLong is also false.

syuilo:
👍🏻

PR #8996, opened 2022-07-14
enhance: make showing ads optional

ThatOneCalculator:
# What
Adds a setting whether to display ads or not, stored on the account

# Why
- Some people don't like ads
- https://stop.voring.me/notes/92ntoujxxe


ThatOneCalculator:
Confirmed all issues fixed and it's behaving as expected.

Luhav (approved):
Indeed it does work as intended without any noticeable issues.

packages/backend/migration/1657759324648-show-ads.js
```diff
@@ -0,0 +1,11 @@
...
+	name = 'showAds1657759324648'
+
+	async up(queryRunner) {
+		await queryRunner.query(`ALTER TABLE "user_profile" ADD "showAds" boolean NOT NULL DEFAULT false`);
+	}
+
+	async down(queryRunner) {
+			await queryRunner.query(`ALTER TABLE "user_profile" DROP "showAds"`);
+	}
+}
```
Johann150: Not sure why this is necessary when you store the setting on the client and not on the server.

ThatOneCalculator: In store.ts, does `where: "account"` not store it on the server?

Johann150: It is stored on the server, but not in the `user_profile` table. It will be [stored as a registry item](https://github.com/misskey-dev/misskey/blob/a4b5a0072d427ba2be4170041f98c11f8ea13e44/packages/client/src/pizzax.ts#L118-L122).

PR #9057, opened 2022-08-16
enhance(client): mobile twitter url can be used as widget

caipira113:
# What

We made the Twitter widget available even when the url is [mobile.twitter.com](mobile.twitter.com).

Untitled-1

# Why

If you look at your timeline, when you share a tweet, you see people sharing it as a mobile link. I wanted to apply it because the Twitter widget does not appear on mobile links.


syuilo:
👍🏻

PR #9159, opened 2022-11-10
[Fix] - Bug fix for Pinned Users

squidicuzz:
### Issue Details/Description
Noticed that pinned users were not showing on instance after latest update.  Console/Log shows error thrown API call for this.

### Resolution
It seems this issue was introduced recently due to a typo..
This Pull Request fixes issues with Pinned users not loading due to the bug in `pinned-users.ts` API endpoint for looking up of pinned users for the instance. 

## Tested Changes
- Updated file `packages/backend/src/server/api/endpoints/pinned-users.ts`
- Tested on local develop instance, and production instance.
- The API error shown in console was resolved.
- Pinned users are shown once again on instance.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ You should also include the user name that made the change.
 ### Improvements
 
 ### Bugfixes
+- Server: Bug fix for Pinned Users lookup on instance @squidicuzz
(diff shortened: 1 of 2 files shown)


syuilo:
👍

PR #9171, opened 2022-11-13
fix typo on CleanRemoteFilesProcessorService

hekacode:
# What

Change tiny typo (cahced ➡️ cached)

# Why

Was looking through source code to better understand how Misskey handles remote file deletion, and this line caught my attention.

# Additional info (optional)

After the recent surge of interest in ActivityPub implementations, I really love how Misskey works, it's cute and pleasing to use! I would like to make more contributions here and there 😄


syuilo:
🙏

PR #9179, opened 2022-11-15
fix(client): userpage ui

Candinya:
# What

Tweak UI, fix incorrect display issues in user's (public) page like https://nya.one/@Candinya

# Why

## Before
image
image
image
image

## After
**Please ignore the color and logo differences (they are included in the commit)**
image
image
image
image

# Additional info (optional)

If any changed page is designed to not have a width limit (or a different limit), please inform me and I'll restore that


syuilo:
🙏🏻

PR #9192, opened 2022-11-17
Person: support movedTo

Mstrodl:
#5475

# What

Supports movedTo attribute on `Person`

# Why

I tried to follow a friend on the wrong account! Now, a warning will appear to tell me they have a different account.

# Additional info (optional)

image

### DIFF
--- a/locales/en-US.yml
+++ b/locales/en-US.yml
@@ -148,6 +148,7 @@ autoAcceptFollowed: "Automatically approve follow requests from users you're fol
 addAccount: "Add account"
 loginFailed: "Failed to sign in"
 showOnRemote: "View on remote instance"
+accountMoved: "User has moved to a new account."
 general: "General"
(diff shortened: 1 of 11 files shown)


Mstrodl:
Note: this doesn't touch the renderer at all, and users on misskey can't set it for themselves. This is only for remote users for now.

Mstrodl:
Also, I noticed the wrong locales tag

I don't know japanese, sorry

tamaina:
I have added one commit, please try it.

tamaina:
In this PR, you have not written the Move activity process, so I think movedTo is only written when a new user is acquired. (If actor update is emitted, it's not a problem).

tamaina:
Actor Updateを処理するのはこのPRで実装したい

tamaina:
🙏🙏🙏

Issue #9202, opened 2022-11-19
server won't startup without @tensorflow/tfjs-node


grainrigi:
## 💡 Summary

With #8984 and #8986, tensorflow has been moved into optionalDependencies and the server should work fine without `@tensorflow/tfjs-node`. However, without the package, initial module resolution still fails since the server statically imports `nsfwjs`, which imports `@tensorflow/tfjs` statically.

## 🥰 Expected Behavior

Even if `@tensorflow/tfjs-node` is not installed at all, the server boots up without any problem.

## 🤬 Actual Behavior

Server fails to boot with the following error:

```
Error: Cannot find module '@tensorflow/tfjs'
Require stack:
- /misskey/packages/backend/node_modules/nsfwjs/dist/index.js
    at Function.Module._resolveFilename (node:internal/modules/cjs/loader:933:15)
    at Function.Module._load (node:internal/modules/cjs/loader:778:27)
    at Module.require (node:internal/modules/cjs/loader:1005:19)
    at require (node:internal/modules/cjs/helpers:102:18)
    at Object.<anonymous> (/misskey/packages/backend/node_modules/nsfwjs/dist/index.js:62:23)
    at Module._compile (node:internal/modules/cjs/loader:1105:14)
    at Object.Module._extensions..js (node:internal/modules/cjs/loader:1159:10)
    at Module.load (node:internal/modules/cjs/loader:981:32)
    at Function.Module._load (node:internal/modules/cjs/loader:822:12)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/translators:170:29)
    at ModuleJob.run (node:internal/modules/esm/module_job:198:25)
    at async Promise.all (index 0)
    at async ESMLoader.import (node:internal/modules/esm/loader:385:24)
    at async workerMain (file:///misskey/packages/backend/built/boot/worker.js:9:5)
    at async default (file:///misskey/packages/backend/built/boot/index.js:25:9) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [ '/misskey/packages/backend/node_modules/nsfwjs/dist/index.js' ]
}
```

## 📝 Steps to Reproduce

1. edit `scripts/install-packages.js` and add `--ignore-optional` option to the yarn install of `packages/backend`
2. build the docker image
3. bring up with docker-compose

## 📌 Environment

Misskey version: 12.119.0
[...]

PR #9206, opened 2022-11-20
chore: Add missing peer dependency for nsfwjs

grainrigi:
# What

Add `@tensorflow/tfjs`, a peer dependency of `nsfwjs`, as a dependency of `packages/backend`.

# Why

As described in #9202, optional support for tensorflow is incomplete and the server won't work without `@tensorflow/tfjs-node`.
In platforms which don't support libtensorflow, this would be a problem because `@tensorflow/tfjs-node` cannot be installed due to missing prebuilt binaries.

### DIFF
--- a/packages/backend/package.json
+++ b/packages/backend/package.json
@@ -31,6 +31,7 @@
 		"@peertube/http-signature": "1.7.0",
 		"@sinonjs/fake-timers": "10.0.0",
 		"@syuilo/aiscript": "0.11.1",
+		"@tensorflow/tfjs": "4.0.0",
 		"ajv": "8.11.2",
 		"archiver": "5.3.1",
 		"autobind-decorator": "2.4.0",


tamaina:
Is this change necessary in yarn v3 (develop branch)?

grainrigi:
Sorry, I misunderstood about optionalDependencies.
I didn't need to use `--ignore-optional` because installation error for optionalDependencies does not affect the exit code of `yarn install` itself.
Without using `--ignore-optional`, `yarn install` completes successfully, and the server worked without any problem!
(Why this works is that trying to install `@tensorflow/tfjs-node` causes the installation of all its dependency (including `@tensorflow/tfjs`), and installation failure of `@tensorflow/tfjs-node` does not remove the dependency installed at that time.)
It might also work well in yarn v3.

Finally, thanks for reviewing this!

grainrigi:
FYI, I tried to test with yarn v3, but `yarn install` ran out all virtual memory and could not confirm if it really works.
(yarn v3 uses really much virtual memory(about 256g on my x86_64 machine), which i386 cannot afford (max. 4GB virtual memory))

tamaina:
`yarn install --immutable` may not consume that much memory. (However, it consumes a lot of memory to change yarn.lock.)

grainrigi:
Unfortunately, `--immutable` didn't really help. (still causes OOM)
For now I'll stick to yarn v1 on my i386 server.
Thank you!

PR #9238, opened 2022-11-28
fix(client): make note visibility inheritance behave correctly when default is set to followers-only

zotanmew:
# What
Fixes an edge case where posts might accidentally end up unlisted instead of follower-only when responding to an unlisted note

# Why
I repeatedly posted notes more public than I intended because of this bug, I'm sure others encountered the same issue


tamaina:
I changed it to support the case of specified.

tamaina (approved):

syuilo:
👍

PR #9247, opened 2022-12-01
fix(server): "forkbomb" DOS mitigation

Skehmatics:
# What

Limit the amount of side-effects / recursions that a single activity resolution can cause.
# Why

A DOS attack is possible by using a specially crafted activitypub server that will serve up an infinite stream of users with featured collections that all reference other users with the same setup. This occurred in the wild on November 30th, 2022 starting sometime around 5PM EST with instances appropriately named `misskey-forkbomb.*`
# Additional info (optional)

The limit (100) was set for no particular reason, and could be modified / parameterized if necessary. The resolver was also chosen for the place for this logic, as it is commonly shared all throughout AP resolution

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,7 +25,7 @@ You should also include the user name that made the change.
 - Server: 引用内の文章がnyaizeされてしまう問題を修正 @kabo2468
 - Server: Bug fix for Pinned Users lookup on instance @squidicuzz
 - Client: インスタンスティッカーのfaviconを読み込む際に偽サイト警告が出ることがあるのを修正 @syuilo
-- Server: Mitigate AP reference chain DOS vector
+- Server: Mitigate AP reference chain DOS vector @skehmatics
 
 ## 12.119.0 (2022/09/10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 ### Improvements
 
 ### Bugfixes
-- 
+-
 
 You should also include the user name that made the change.
 -->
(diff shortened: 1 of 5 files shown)


acid-chicken (approved):

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@ You should also include the user name that made the change.
 - Server: 引用内の文章がnyaizeされてしまう問題を修正 @kabo2468
 - Server: Bug fix for Pinned Users lookup on instance @squidicuzz
 - Client: インスタンスティッカーのfaviconを読み込む際に偽サイト警告が出ることがあるのを修正 @syuilo
+- Server: Mitigate AP reference chain DOS vector @skehmatics
```
acid-chicken:
```suggestion
- Server: Mitigate AP reference chain DoS vector @skehmatics
```

packages/backend/src/core/remote/activitypub/ApResolverService.ts
```diff
@@ -116,6 +117,10 @@ export class Resolver {
 			throw new Error('cannot resolve already resolved one');
 		}
 
+		if (this.history.size > this.recursionLimit) {
+			throw new Error('hit recursion limit');
```
acid-chicken:
```suggestion
			throw new Error(`hit recursion limit: ${this.utilityService.extractDbHost(value)}`);
```

I think appending the hostname may help for moderation.

ThatOneCalculator:
I've tested this on Calckey (and stop.voring.me), works as expected.

syuilo (approved):

tamaina:
マージしないのかしら

ThatOneCalculator:
@syuilo this should seriously be merged. Maybe even backported to a 12.119.1 release before 12.120.0.

PR #9300, opened 2022-12-09
fix: typo

xtexx:
Fix: `metohd` -> `method`
Chore: add some return type declarations to functions


syuilo:
👍

xtexx:
Why is the `/deploy` comment not working?

PR #9311, opened 2022-12-12
Fix lint errors (except `@typescript-eslint/prefer-nullish-coalescing`)

saschanaz:
# What

Fixes linter errors

# Why

To make the CI happy

# Additional info (optional)

Mostly just formatting changes. Excluded `@typescript-eslint/prefer-nullish-coalescing` as I'm not confident enough that migrating to `??` is safe. I'd rather just disable it given that it has been failing for a while.


syuilo:
👍

PR #9314, opened 2022-12-12
refactor(sw): Typecheck the service worker

saschanaz:
# What

Run `tsc --noEmit` in the service worker build step

# Why

Currently the service worker build never typechecks, since esbuild and typescript-eslint don't do such job.

esbuild: https://esbuild.github.io/content-types/#typescript

>However, esbuild does not do any type checking so you will still need to run tsc -noEmit in parallel with esbuild to check types. This is not something esbuild does itself.

typescript-eslint: https://typescript-eslint.io/linting/troubleshooting#why-dont-i-see-typescript-errors-in-my-eslint-output

>TypeScript's compiler (or whatever your build chain may be) is specifically designed and built to validate the correctness of your codebase. Our tooling does not reproduce the errors that TypeScript provides, because doing so would slow down the lint run [1], and duplicate the errors that TypeScript already outputs for you.

Adding this step adds tons of TS errors 😱

# Additional info (optional)

### DIFF
--- a/packages/sw/package.json
+++ b/packages/sw/package.json
@@ -4,7 +4,7 @@
 	"scripts": {
 		"watch": "node build.js watch",
 		"build": "node build.js",
-		"lint": "eslint --quiet src/**/*.ts"
+		"lint": "tsc --noEmit && eslint --quiet src/**/*.ts"
 	},
 	"dependencies": {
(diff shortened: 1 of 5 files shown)


tamaina:
👍

PR #9316, opened 2022-12-12
Call `tsc --noEmit` also for backend/client

saschanaz:
# What

Call `tsc --noEmit` also for backend/client

# Why

Backend explicitly ignores typescript errors by `|| echo done.` for some reason, and also client because vite does not do typecheck.

https://vitejs.dev/guide/features.html#typescript

>Vite only performs transpilation on .ts files and does NOT perform type checking. It assumes type checking is taken care of by your IDE and build process (you can run tsc --noEmit in the build script or install vue-tsc and run vue-tsc --noEmit to also type check your *.vue files).

# Additional info (optional)


tamaina:
👍

PR #9319, opened 2022-12-13
Fix MetaService unit tests

saschanaz:
# What

1. Removes `mockImplementation` calls
2. Adds `restoreMocks: true` configuration

# Why

1. The mock function by `spyOn` should already call the original implementation. Using `mockImplementation` is redundant and is also wrong, since it should do `originalFunction.call(db, args)` rather than calling `originalFunction` directly.
2. As described in the comment, somehow `spy.mockRestore()` didn't work and kept the spy alive. The `restoreMocks` configuration does the work.

# Additional info (optional)


syuilo:
🙏🏻🙏🏻🙏🏻

PR #9320, opened 2022-12-13
Use 127.0.0.1 for misskey/test.yml

saschanaz:
# What

Replaces `localhost` with `127.0.0.1` on `.github/misskey/test.yml`

# Why

Docker on WSL2 on Windows somehow can't access containers via localhost but only can via 127.0.0.1. See also: https://github.com/microsoft/WSL/issues/4983

This change also makes things consistent with docker-compose:

https://github.com/misskey-dev/misskey/blob/f30d54fe889c4da039c83bae0a91e624de689143/packages/backend/test/docker-compose.yml#L4-L7

# Additional info (optional)


tamaina:
これなんでself-requestしたんだ…？

tamaina:
👍

PR #9335, opened 2022-12-17
Remove redundant ts-node things

saschanaz:
# What

Removes ts-node dependency and loader

# Why

b75184ec8e3436200bacdcd832e3324702553d20 removed the usage. (It was to run `mocha` but now `@swc/jest` is doing the same job.)

# Additional info (optional)


syuilo:
👍🏻

PR #9340, opened 2022-12-18
refactor(client): add proper types to `never[]`

saschanaz:
# What

Added proper types to arrays with no types

# Why

To fix "Property 'foo' does not exist on type 'never'" TS2339 errors when accessing those arrays

# Additional info (optional)


syuilo:
🙏🙏🙏

PR #9341, opened 2022-12-18
chore: deploy vue-tsc for client typecheck

saschanaz:
# What

Use `vue-tsc --noEmit` instead of `tsc --noEmit` (which is provided by [Volar](https://github.com/johnsoncodehk/volar/tree/master/vue-language-tools/vue-tsc))

# Why

`tsc` does not support Vue files and thus is ignoring them. Currently [Misskey uses Volar for VSCode extension](https://github.com/misskey-dev/misskey/blob/bb3d274db64fe662ad01780ca5eadbc263f6f759/.vscode/extensions.json#L6), so it should be consistent to use the same tool provided by them.

# Additional info (optional)

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -4,7 +4,7 @@
 	"scripts": {
 		"watch": "vite build --watch --mode development",
 		"build": "vite build",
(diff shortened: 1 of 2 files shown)


syuilo:
🆒

Issue #9345, opened 2022-12-18
Add unit tests for client package


saschanaz:
I'd like to add some tests since there's no tests AFAICT.

pixeldesu:
Well, there are no direct unit tests of components, but there are E2E tests for the client in `cypress/`

saschanaz:
Good point, changing the title.

syuilo:
Does this mean testing each Vue component?

syuilo:
If so, I don't know because I have never written unit tests for Vue, but I feel it would be a difficult task since many components rely on global state.

saschanaz:
I'd start with splitting some Jest-runnable subfeatures used by Vue components and testing them, e.g. image compression configuration function I added in #9334, rather than testing full Vue components themselves. Such thing should be easier with Cypress I guess.

tamaina:
Vue/ViteのテストはVitestでテストを書けば良さそう?

モジュールのmockもできる

https://vitest.dev/api/vi.html#vi-domock

acid-chicken:
Related to #10336

acid-chicken:
でもそれはそれとしてグローバルステートはストアとかは良いけど Web API (HTTP API じゃない方) の直接依存は剥がしたい

saschanaz:
https://github.com/misskey-dev/misskey/tree/develop/packages/frontend/test ここにユニットテスト少しあります

tamaina:
とりあえずcloseして良さそう

PR #9346, opened 2022-12-18
enhance(client): case insensitive emoji search

### LINKED ISSUE #9279: Make emoji search case insensitive
## Summary

Android keyboards by default makes the first character uppercase. That makes the emoji search fail easily.

1. Search "Blobcat", no result
2. Search "blobcat", now there are results

saschanaz:
# What

MkEmojiPicker now searches emojis in case insensitive way

# Why

Mobile OSes by default writes words with autocapitalization and thus mobile users have to explicitly switch to lowercase to properly search for emojis. HTML [`autocapitalize`](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize) attribute is still Chromium only, so case insensitivity is the next option.

# Additional info (optional)

(Filed #9345 for test)

Fixes #9279

### DIFF
--- a/packages/client/src/components/MkEmojiPicker.vue
+++ b/packages/client/src/components/MkEmojiPicker.vue
@@ -135,7 +135,7 @@ watch(q, () => {
 		return;
 	}
 
-	const newQ = q.value.replace(/:/g, '');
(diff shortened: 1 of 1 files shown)


syuilo:
👍

PR #9348, opened 2022-12-18
Fix peers API returning suspended instances

ineffyble:
# What

Exclude suspended instances from peers returned from `/v1/instance/peers`.

# Why

Suspended instances are not actually federated peers of the server and should not appear in the peer list. Removing them matches the behaviour of the Mastodon API that this endpoint is matching.

# Additional info (optional)

I don't have a working dev environment but I think this should just work.


syuilo:
👍

syuilo:
👍

PR #9350, opened 2022-12-19
enhance(client): Page embed html block

### LINKED ISSUE #9333: Embed HTML in Misskey Pages
## Summary
Add a block option to pages to allow static HTML to be displayed

Would enable more customized pages or embedding of existing html content from other sources without using MFM. However, this would not be compatible with AiScript or the other scripting abilities in pages because of sand-boxing.

## High Level Design
New block type, `EmbedBlock` created in client that contains srchtml contents, optional height of the iframe to display, and optional background color.

Use a sand-boxed iframe to isolate user HTML from rest of site and disable script execution inside.

ssmucny:
# What

Adding a new page block of type 'Embed' for external web pages and HTML.

# Why

Resolve #9333 

# Additional info (optional)

Localization is not completed. Placeholder values are used now. Please let me know if a separate PR is needed for this.

Changes are all in client.

HPML does not interpolate variables in embed block.

# UI examples

## URL
Screenshot_20221218_191403
Screenshot_20221218_191434

## HTML
Screenshot_20221218_192212
Screenshot_20221218_192231

### DIFF
--- a/locales/en-US.yml
+++ b/locales/en-US.yml
@@ -1441,7 +1441,6 @@ _pages:
     section: "Section"
     image: "Images"
     button: "Button"
-    embed: "Embed"
     if: "If"
     _if:
       variable: "Variable"
@@ -1509,9 +1508,6 @@ _pages:
       title: "Title"
       values: "List of choices separated by line breaks"
       default: "Default value"
-    _embed:
-      height: "Height"
-      backgroundColor: "Backround Color"
   script:
     categories:
       flow: "Flow control"
--- a/locales/en-US.yml
+++ b/locales/en-US.yml
@@ -1510,10 +1510,8 @@ _pages:
       values: "List of choices separated by line breaks"
       default: "Default value"
     _embed:
-      url: "Embed URL"
-      html: "Custom HTML"
-      height: "Height Override"
-      backgroundColor: "Backround Color Override"
+      height: "Height"
+      backgroundColor: "Backround Color"
   script:
     categories:
       flow: "Flow control"
--- a/locales/en-US.yml
+++ b/locales/en-US.yml
@@ -1441,6 +1441,7 @@ _pages:
     section: "Section"
     image: "Images"
     button: "Button"
+    embed: "Embed"
     if: "If"
     _if:
       variable: "Variable"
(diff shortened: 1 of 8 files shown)


tamaina:
これなんかセキュリティ的にすごく面倒そう

ssmucny:
With proper sand-boxing (https://cloud.google.com/blog/products/data-analytics/iframe-sandbox-tutorial), many of the security concerns with iframes can be addressed. I would recommend sand-boxing completely (no script execution, no modals, no same origin or top navigation, etc.). With that, only static HTML will be able to be displayed.

All major browsers support sand-boxing iframes, but this could be an issue if users are on a particularly old version: https://www.w3schools.com/tags/att_iframe_sandbox.asp 

Even with sand-boxing, I would like others to provide input on any other possible security concerns I am missing before a merge is approved. 🙏

acid-chicken:
I don't think it is safe to say that if only script is disabled. (e.g. <https://jakearchibald.com/2018/third-party-css-is-not-safe/>)

ssmucny:
> I don't think it is safe to say that if only script is disabled. (e.g. https://jakearchibald.com/2018/third-party-css-is-not-safe/)

Thanks for the feedback. It looks like the majority of the concerns with CSS are including it in the main page where it can mess with Misskey itself. Inside a sand-boxed iframe the main issue that I still see is the ability to monitor what a user is doing inside the iframe (hovers, typing). This is somewhat mitigated by having the sandbox not allowing forms, but that just prevents form submission. Filling out text fields is still allowed.

I will have to investigate to see if a CSP can be reliably enforced for this in iframes.

ssmucny:
I found a good example of embedding HTML from emails that I think is a good use case to enable. https://making.close.com/posts/rendering-untrusted-html-email-safely. It mentions a handful of good ideas to implement that lock down the ability for malicious actors to exploit the system.

Some key points that could be included in this PR:
- [x] sandbox iframe (This PR actually locks down the content more than may be necessary, but I'd rather be safe than sorry)
- [ ] apply CSP attribute (I don't know if iframes strictly enforce this on embedded content)
- [ ] check for unsupported browsers and display error message instead of content

PR #9351, opened 2022-12-19
refactor: fix broken Vue component custom properties

saschanaz:
# What

1. Removed DefineComponent
2. Changed `typeof i18n['locale']` to `typeof i18n['ts']`
3. Moved all custom property declarations to `vue.d.ts`
4. Imported `vue` instead of `@vue/runtime-core`

# Why

1. It has been broken as DefineComponent doesn't exist now in Vue, and importing Vue components works well without it.
2. Reflected the change from 57ec04d9ecc51060225bb15867215c7475685f92
3. Because the comments didn't like the current positions and I saw 0a235e5e482b260ab4b3597f380eb462e74707b1 from #6976 initially wanted to put them in `vue.d.ts`. It seems Volar is happy with it.
4. Importing `runtime-core` directly does not work (probably because Misskey don't install it directly). The official docs recommend `vue` so I just followed it. https://vuejs.org/api/utility-types.html#componentcustomproperties

# Additional info (optional)

### DIFF
--- a/packages/client/@types/vue.d.ts
+++ b/packages/client/@types/vue.d.ts
@@ -1,7 +1,16 @@
 /// <reference types="vue/macros-global" />
 
-declare module '*.vue' {
-	import type { DefineComponent } from 'vue';
-	const component: DefineComponent<{}, {}, any>;
-	export default component;
+import type { $i } from '@/account';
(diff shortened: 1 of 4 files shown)


syuilo:
👍👍

PR #9356, opened 2022-12-19
fix(backend): request.body may be undefined

saschanaz:
# What

Fix backend assuming POST body always exists

# Why

Because cypress was throwing `Cannot read properties of undefined (reading 'i')` before #9355. This can easily be reproduced by `curl -X POST http://localhost:61812/api/reset-db`.

# Additional info (optional)

Forgot to push this one to #9355.


syuilo:
👍

PR #9357, opened 2022-12-19
fix(client): add missing ref for MkButton

saschanaz:
# What

Added missing `ref=el` to MkButton

# Why

Because Cypress was failing because of `Cannot read properties of null (reading 'focus')`.

https://github.com/misskey-dev/misskey/blob/c3cb2189753aa74b8fc5c554577deeed89971dc6/packages/client/src/components/MkButton.vue#L51-L57

# Additional info (optional)

This now should make Cypress fully green 🎉🟩

### DIFF
--- a/packages/client/src/components/MkButton.vue
+++ b/packages/client/src/components/MkButton.vue
@@ -1,5 +1,6 @@
 <template>
 <button
+	ref="el"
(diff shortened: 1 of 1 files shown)


syuilo:
🙏

PR #9368, opened 2022-12-20
refactor(client): explicit `showing` property for tooltips

saschanaz:
# What

Explicitly defined `showing` properties for tooltip components and bound them to MkTooltip

# Why

Currently it's implicitly bound to child components by callers of `os.popup` and then by `common.vue`:

https://github.com/misskey-dev/misskey/blob/bc0e600e51ad6ed24919d862dc97c5d6faf71314/packages/client/src/components/MkReactionsViewer.reaction.vue#L82-L89

https://github.com/misskey-dev/misskey/blob/bc0e600e51ad6ed24919d862dc97c5d6faf71314/packages/client/src/ui/_common_/common.vue#L2-L8

While this works, TypeScript hates this as it's not explicit whether the child MkTooltip will ever get the required `showing` property. This patch makes it explicit, following what MkChartTooltip already does.

https://github.com/misskey-dev/misskey/blob/bc0e600e51ad6ed24919d862dc97c5d6faf71314/packages/client/src/components/MkChartTooltip.vue#L2

# Additional info (optional)

### DIFF
--- a/packages/client/src/components/MkChartTooltip.vue
+++ b/packages/client/src/components/MkChartTooltip.vue
@@ -16,7 +16,7 @@
 import { } from 'vue';
 import MkTooltip from './MkTooltip.vue';
 
-const props = defineProps<{
+defineProps<{
 	showing: boolean;
 	x: number;
 	y: number;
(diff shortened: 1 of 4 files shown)


syuilo:
good catch

PR #9378, opened 2022-12-21
ApRequestService: don't generate our own Host header

Mstrodl:
# What

Though it's needed for signatures, including our own `Host` header means `node-fetch` won't generate a new one for us when we encounter a redirect, leading to breakage!

# Why

This is necessary for compatibility with some weird instances like https://sleeping.town which redirect to https://mastodon.sleeping.town

# Additional info (optional)

Something strange, misskey.io doesn't seem to have issues resolving users, but my instance does! Not sure what to make of this...
Also changes `Host` header we generate to use the `host` property, which will include the port.
It could be interesting to see the signature stuff implemented as an agent instead, which would allow us to modify the request right before it's sent, hopefully including the exact headers we need to sign!

### DIFF
--- a/packages/backend/src/core/activitypub/ApRequestService.ts
+++ b/packages/backend/src/core/activitypub/ApRequestService.ts
@@ -47,7 +47,7 @@ export class ApRequestService {
 			method: 'POST',
 			headers: this.objectAssignWithLcKey({
 				'Date': new Date().toUTCString(),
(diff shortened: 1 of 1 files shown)


Mstrodl:
I've been using this locally for the past month and it works great. Any news?

tamaina:
Behavior may be different since we changed from node-fetch to undici

PR #9385, opened 2022-12-22
enhance(client): update emoji picker immediately on all input

saschanaz:
# What

1. Use raw `.value` instead of `v-model` for emoji-applicable form elements
2. Removed `window.clipboardData` usage
3. A few more TypeScript related refactoring

# Why

1. `v-model` does not update while composition is happening. Android always causes composition on all languages and thus Android users are affected. (Except users on Chromium-based browsers, which somehow immediately flush every input even on CJK IME. 🤔)
2. `window.clipboardData` is IE-era thing and does not exist on modern browsers (and thus TypeScript is unhappy with it)
3. To make the linter happy

# Additional info (optional)


syuilo:
👍

PR #9389, opened 2022-12-22
style(client): fix `vue/v-on-event-hyphenation`

saschanaz:
# What

Fixed `vue/v-on-event-hyphenation` warnings

# Why

To reduce ESLint warnings (which are more tricky because they somehow overlap with TypeScript errors)

# Additional info (optional)


syuilo:
👍

PR #9399, opened 2022-12-22
enhance(client): show Unicode emoji tooltip with its name

saschanaz:
# What

1. MkReactionViewer and MkEmoji now shows the tooltip with the name when including Unicode emoji, lazily computed on pointerenter
2. Some TypeScript related refactoring

# Why

1. Searching from an array with 2000 items for every emoji felt like too energy-consuming, so I decided to do it lazily on pointerenter. (But I have no actual performance profile 🤔)
  * `alt`はOSで各言語に読めるようになるので変更不要 / `alt` is not changed here as OSes should be able to read it in each language. 
3. For less warnings

# Additional info (optional)


syuilo:
- altを削除すると、テキストをコピーしたときに絵文字が消えます
- > Searching from an array with 2000 items for every emoji felt like too energy-consuming
  - これはコード内にコメントとして残しておくと良いと思います。

PR #9419, opened 2022-12-27
enhance(client): Add link to copy RSS feed in user profile menu

ssmucny:
# What

Add a menu option to copy the user RSS feed link

# Why

Make finding and using the RSS/Atom feature easier for users through GUI

# Additional info (optional)

image


syuilo:
🙏🏻

PR #9462, opened 2023-01-03
Fix: content type application/x-www-form-urlencoded is not allowed

paring-chan:
# What

fixes `Unsupported Media Type` error for `application/x-www-form-urlencoded` content type

# Why

Without this, some of tools that integrates with misskey will be broken. (example: https://misskey.tools)  
I think this is caused by the default content type on POST request of axios

### DIFF
--- a/packages/backend/src/server/ServerService.ts
+++ b/packages/backend/src/server/ServerService.ts
@@ -1,10 +1,8 @@
 import cluster from 'node:cluster';
 import * as fs from 'node:fs';
-import * as http from 'node:http';
 import { Inject, Injectable } from '@nestjs/common';
 import Fastify from 'fastify';
 import { IsNull } from 'typeorm';
-import fastifyFormBody from '@fastify/formbody';
 import { GlobalEventService } from '@/core/GlobalEventService.js';
 import type { Config } from '@/config.js';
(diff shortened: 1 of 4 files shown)


packages/backend/src/server/ServerService.ts
```diff
@@ -70,6 +70,8 @@ export class ServerService {
 			});
 		}
 
+		fastify.register(fastifyFormBody);
```
syuilo: Maybe ApiServerService is better place

HotoRas:
Note: this will fix #9448 if works as expected.
Checked on the test environment (dev build, server & domain provided by @pikokr, code on this pr) and worked quite well.

tamaina:
p1.a9z.devに適用してみたが、 #9448 (Misskey Toolsのエラー)は治らなかった

syuilo:
そもそもMisskeyAPIの仕様的に今までapplication/x-www-form-urlencodedってサポートしてたっけ？

tamaina:
https://github.com/misskey-dev/misskey/pull/3740

こんなプルリクがあった

syuilo:
see https://github.com/misskey-dev/misskey/pull/3740

PR #9479, opened 2023-01-06
Support remote objects in search

### LINKED ISSUE #9428: Search page should support remote URLs
## Summary

The search page should support remote URLs (when logged in). Combined with #9427 this would allow quickly switching to the local instance from any remote post.

(We have implementation ready to go, this issue is just following the contributing guidelines.)

Example with #9427: (keyword needs to be manually added in browser settings)

[Screencast_from_2022-12-28_21-23-21.webm](https://user-images.githubusercontent.com/1779619/209882885-85d41c1a-25e4-431f-82a9-f0a095db2876.webm)

(can also be done from remote post)

SoniEx2:
Closes #9428

# What

This allows the standalone search page to also fetch/redirect to local copies of remote objects

# Why

It enables the flow in #9428

# Additional info (optional)

### DIFF
--- a/packages/frontend/src/pages/search.vue
+++ b/packages/frontend/src/pages/search.vue
@@ -14,6 +14,7 @@ import { i18n } from '@/i18n';
 import { definePageMetadata } from '@/scripts/page-metadata';
 import * as os from '@/os';
 import { mainRouter } from '@/router';
+import { $i } from '@/account';
 
 const props = defineProps<{
 	query: string;
@@ -22,7 +23,7 @@ const props = defineProps<{
 
 const query = props.query;
 
-if (localStorage.getItem('account') != null) {
+if ($i != null) {
 	if (query.startsWith('https://') || (query.startsWith('@') && !query.includes(' '))) {
 		const promise = os.api('ap/show', {
 			uri: props.query,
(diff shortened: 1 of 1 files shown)


tamaina (approved):

packages/frontend/src/pages/search.vue
```diff
@@ -12,12 +12,34 @@ import { computed } from 'vue';
 import XNotes from '@/components/MkNotes.vue';
 import { i18n } from '@/i18n';
 import { definePageMetadata } from '@/scripts/page-metadata';
+import * as os from '@/os';
+import { mainRouter } from '@/router';
```
tamaina: It is likely that only mainRouter will be used since it is just after the browser tab is displayed, but I think it is preferable to use useRouter() as much as possible as a courtesy.

packages/frontend/src/pages/search.vue
```diff
@@ -12,12 +12,34 @@ import { computed } from 'vue';
...
 
 const props = defineProps<{
 	query: string;
 	channel?: string;
 }>();
 
+const query = props.query;
+
+if (localStorage.getItem('account') != null) {
```
syuilo:
```suggestion
if ($i != null) {
```
and
```
import { $i } from '@/account';
```

syuilo:
👍🏻

PR #9578, opened 2023-01-15
chore: update year in COPYING file

Candinya:
# What

Change end year in COPYING file from 2022 to 2023

# Why

It's 2023 now

# Additional info (optional)

Happy new year :D


syuilo:
🎉

PR #9758, opened 2023-01-31
fix(ap): ld signature normalizer

### LINKED ISSUE #9752: Misskey v13 failed to communicate with some of the mastodon sites and old misskey sites via relay
## 💡 Summary

After upgrade to v13, my Misskey instance failed to communicate with some of the other instances via relay. That is, for example, with misskey v13, most of the notes from `o3o.ca` no longer be shown in the global timeline, and the requests to `o3o.ca` showed a significant drop (red box in the picture).
图片

Meanwhile, old misskey instances (v12) could not communicate to v13 in a similar manner.

This behavior could be reproduced in a newly set instance.

## 🥰 Expected Behavior

All notes from mastodon and misskey v12 instances connected to the same relay should be shown in global timeline.

## 🤬 Actual Behavior

On a new set misskey v13 instances, a lot of notes from misskey v13 instances were shown in global timeline. No note from mastodon was shown.

On a new set misskey v12 instances, only notes from mastodon were shown in global timeline. Notes from misskey v12 instances might could also be shown in global timeline, but today active v12 instances are not common.

## 📝 Steps to Reproduce

1. Setup two new instance, use v12.119.2 and v13.2.5 respectively. The two instances use similar configs:
```
url: https://misskeytestv12.chn.moe/
port: 3004
db:
  host: 127.0.0.1
  port: 5432
  db: misskeytestv12
  user: xxx
  pass: xxxxxxxxxxxxxxx
  disableCache: true
  extra:
    statement_timeout: 600000
redis:
  host: localhost
  port: 6379
id: 'aid'
proxyRemoteFiles: true
signToActivityPubGet: true
```
2. Do not follow or search any one, Just add a relay: `https://relay.nya.one/inbox`
3. Wait for several minutes. In the global timeline of misskey v12 instance, only a lot of notes from mastodon was shown. In the global timeline of misskey v13 instance, only a lot of notes from misskey v13 was shown. 

A screenshot from v12 instance:

图片

A screenshot from v13 instance:

图片

## 📌 Environment

Misskey version: 13.2.5, running with docker
Your OS: Gentoo
Your browser: firefox

Candinya:
# What

Fix #9752

# Why

The signature function has changed during major migration, and thus breaks the json-ld signature, causing misskey's isolation with other fediverse instances.

# Additional info (optional)

### DIFF
--- a/packages/backend/src/core/activitypub/LdSignatureService.ts
+++ b/packages/backend/src/core/activitypub/LdSignatureService.ts
@@ -1,5 +1,6 @@
 import * as crypto from 'node:crypto';
 import { Inject, Injectable } from '@nestjs/common';
+import jsonld from 'jsonld';
 import { HttpRequestService } from '@/core/HttpRequestService.js';
 import { bindThis } from '@/decorators.js';
 import { CONTEXTS } from './misc/contexts.js';
@@ -84,7 +85,9 @@ class LdSignature {
 	@bindThis
 	public async normalize(data: any) {
 		const customLoader = this.getLoader();
-		return 42;
+		return await jsonld.normalize(data, {
+			documentLoader: customLoader,
+		});
 	}
 
 	@bindThis


rakino:
Applied to my instance, it works!

syuilo:
Thanks🙏

Issue #9786, opened 2023-02-03
Media player fails to load remote media due to CORS


zotanmew:
## 💡 Summary

Media player fails to load remote media due to CORS policy

## 🥰 Expected Behavior

The old browser-default player appears to have mostly ignored CORS and played everything fine

## 🤬 Actual Behavior

Player is stuck on loading, console says `has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.`

## 📝 Steps to Reproduce

1. View post from remote instance containing a video attachment
2. Try to play it

## 📌 Environment

Misskey version: 13.3.0
Your OS: ArchLinux latest
Your browser: Chrome latest stable

RoxyBoxxy:
in ur nginx configuration add 

```
add_header 'Access-Control-Allow-Origin' '*' always;
```

to your location under proxy_pass i put mine under `proxy_redirect off;`

zotanmew:
Heya, that sadly does not solve the problem as this would need to be added to the configuration of every remote instance :/

RoxyBoxxy:
huh i didnt have any problem playing content from any other instance, could you send me your profile on ur instance and i can try it out on my own

RoxyBoxxy:
I just did a test from a mastodon instance to my misskey instance and seems to work fine over here 
![image](https://user-images.githubusercontent.com/8870878/216707732-cb8d625f-7f72-4de5-b822-1cea13bf088b.png)

zotanmew:
Broken post example: https://botsin.space/@hackers_gifs/109802917994699880
Working post example: https://glitterkitten.co.uk/@doot/109802986994518206

zotanmew:
The working instance sends a cors header, the broken one doesn't. Most appear to not do this by default.

zotanmew:
Oh, also: I have remote media caching disabled. I presume this bug does not occur when remote media caching is enabled, because all requests would go to the local server.

RoxyBoxxy:
> Oh, also: I have remote media caching disabled. I presume this bug does not occur when remote media caching is enabled, because all requests would go to the local server.

yer i think i run remote media caching as i also use a cdn service for all files, also i saw that you was loading gifs? as the links you sent only contain gifs

zotanmew:
Huh. Yeah, looks like it. Pre v13.3 this worked fine as it just used the browser internal player which apparently doesn't check CORS?

RoxyBoxxy:
[...]

PR #9790, opened 2023-02-03
Fix | Vue-plyr CORS issue

RoxyBoxxy:
# What
Fixed CORS in vue-plyr causeing instance with remote media caching not being able to play videos

# Why
The CORS policy was active when instance wih remote media caching was disabled


RoxyBoxxy:
Related issue #9786

syuilo:
🙏🏻

PR #9810, opened 2023-02-05
fix docker health check

### LINKED ISSUE #9811: Docker healthcheck does not work properly
## 💡 Summary

1. curl is not installed
2. yq runs only in amd64(current dockerfile downloads the yq binary only for amd64)

## 🥰 Expected Behavior

the curl command should send the request

## 🤬 Actual Behavior

1. the yq command does not work on non amd64 servers
2. curl command fails as it does not exist

## 📝 Steps to Reproduce

1. just run container

## 📌 Environment

Misskey version: after https://github.com/misskey-dev/misskey/commit/f5993373200746856ccdcd8629d8180c8d3d1b5a

paring-chan:
this removes yq command and uses grep and awk to get port

fixes #9811

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 	; echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache \
 	&& apt-get update \
 	&& apt-get install -yqq --no-install-recommends \
-	build-essential wget ca-certificates
+	build-essential
 
 RUN corepack enable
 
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -8,9 +8,7 @@ RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
 	; echo 'Binary::apt::APT::Keep-Downloaded-Packages "true";' > /etc/apt/apt.conf.d/keep-cache \
 	&& apt-get update \
 	&& apt-get install -yqq --no-install-recommends \
-	build-essential wget ca-certificates \
-	&& wget https://github.com/mikefarah/yq/releases/latest/download/yq_linux_amd64 -O /usr/bin/yq \
-	&& chmod +x /usr/bin/yq
+	build-essential wget ca-certificates
```
massongit:
```suggestion
	build-essential
```


`wget` and `ca-certificates` can remove because those packages need to install `yq` only.

ncwhale:
Just an idea: why should we using bash when we got a lot node modules installed?

How about modify the health check to `health_check.js` and just using node to do it?

Then we do not need other things such as `curl` installed and depended.

syuilo:
🙏🏻

PR #9850, opened 2023-02-09
fix: prevent clipping audio plyr's tooltip

Candinya:
# What

Audio plyr's tooltip was wrongly clipped by this style.

# Why

To show audio plyr tooltip properly

# Additional info (optional)

Before:
image

After:
image


syuilo:
🙏

PR #9924, opened 2023-02-13
fix: Allow configuring of the host/address to listen/bind to

### LINKED ISSUE #7738: Listen address を変更可能にする
## Summary
MisskeyのListen addressは、おそらくNode.js側のデフォルトの`::`や`0.0.0.0`になっているが、
https://nodejs.org/api/net.html#net_server_listen_port_host_backlog_callback
広すぎる気がするので、デフォルトでは`127.0.0.1`にして、設定により変更できるようになどする。
→ デフォルトはそのままの方が良いかな

ledlamp:
First of all, the default behavior of node.js server is to listen on all addresses of **both** IPv4 and IPv6. However, Misskey was hard-coded to listen on `"0.0.0.0"`, which makes node.js listen on IPv4 only.

Second, the administrator should be able to configure the host, **just like any other server allows you to do**. In production, when the server is behind a reverse proxy, it is a good idea to make the backend listen only on a loopback address, so that it is not publicly exposed. There are also other possible reasons, for example, binding to port 80 of an IPv6 address to connect directly to Cloudflare, without conflicting with other web servers on the same system.

So with this PR, `host` can be configured just like `port`, and it defaults to `undefined` so node.js listens on both stacks, not just IPv4. Closes #7738

### DIFF
--- a/.config/example.yml
+++ b/.config/example.yml
@@ -30,6 +30,9 @@ url: https://example.tld/
 # The port that your Misskey server should listen on.
 port: 3000
 
+# The hostname or address to listen on (default: all)
+#host: "127.0.0.1"
+
 #   ┌──────────────────────────┐
 #───┘ PostgreSQL configuration └────────────────────────────────
 
--- a/packages/backend/src/boot/master.ts
+++ b/packages/backend/src/boot/master.ts
@@ -78,7 +78,7 @@ export async function masterMain() {
 		await spawnWorkers(config.clusterLimit);
 	}
 
-	bootLogger.succ(`Now listening on port ${config.port} on ${config.url}`, null, true);
+	bootLogger.succ(`Now listening on ${config.host ? `[${config.host}]:` : ''}${config.port} for ${config.url}`, null, true);
 }
 
 function showEnvironment(): void {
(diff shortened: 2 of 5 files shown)


.config/example.yml
```diff
@@ -30,6 +30,9 @@ url: https://example.tld/
 # The port that your Misskey server should listen on.
 port: 3000
 
+# The hostname or address to listen on (default: all)
+#host: "127.0.0.1"
```
syuilo, line 34: The name "host" has a wide range and can be confusing, so how about "listenHost" or something similar?

ledlamp: I think it is okay because I can not think of anything else it could mean in this context. It is simple and consistent with the node.js code.

If it's called "listenHost", then should "port" be called "listenPort"?

saschanaz: Maybe `hostname` is better? (Oh but there's already `host` below...)

syuilo: When people see the property config.host, they don't know what host it refers to.

> then should "port" be called "listenPort"?

I think so.

ledlamp: what about `bind` or `bindAddress`?

saschanaz: This is a server configuration, so IMO host/port shouldn't be too confusing, especially with the comments here.

saschanaz: (But then it could be confusing too because below `host` means the address the server connects to, and here it's to receive connections. Maybe bind is better then.)

syuilo:
👍🏻

PR #9936, opened 2023-02-14
fix(frontend): fix type errors from vite.json5.ts

saschanaz:
# What

* Replaced regex with line/column fields

# Why

* The regex was wrong and also causing type errors

# Additional info (optional)


tamaina:
👍👍👍

PR #10033, opened 2023-02-23
enhance: make pwa icon maskable

1000ch:
# What

Make [the current icon for PWA maskable](https://web.dev/i18n/ja/maskable-icon/).

# Why

Because an installed PWA looks like below.

And current icon seems to be maskable already so by having this change PWA users would see a better-looking misskey icon on their desktop.


syuilo:
👍🏻

PR #10088, opened 2023-02-25
Simplify search.vue (remove dead code)

SoniEx2:
This is already handled by the code above it, no need to handle it twice


syuilo:
👍🏻

PR #10156, opened 2023-03-01
refactor(frontend): add explicit `Promise<void>` return types in `os.ts`

saschanaz:
`Expected 1 arguments, but got 0. Did you forget to include 'void' in your type argument to 'Promise'?ts(2794)`のようなエラーを修正


syuilo:
👍🏻

PR #11039, opened 2023-06-21
fix(backend): Resolve missing parseObjectId in IdService

caipira113:
## What
Add parseObjectId to IdService in parse method.

## Why
I added the "objectId" because it exists for compatibility purposes and even though the parse is implemented, it was missing in the IdService.

## Additional info (optional)
None.

## Checklist
- [ ] (If possible) Add tests


syuilo:
🙏

PR #11047, opened 2023-06-25
fix(backend): APIエラーのHTTP status code変更

saschanaz:
## What

* authenticationエラーにHTTP 401
* permissonエラーにHTTP 403

## Why

https://www.rfc-editor.org/rfc/rfc9110#status.401

>The 401 (Unauthorized) status code indicates that the request has not been applied because it lacks valid authentication credentials for the target resource.
>If the request included authentication credentials, then the 401 response indicates that authorization has been refused for those credentials.

https://www.rfc-editor.org/rfc/rfc9110#name-403-forbidden

>The 403 (Forbidden) status code indicates that the server understood the request but refuses to fulfill it. A server that wishes to make public why the request has been forbidden can describe that reason in the response content (if any).
>If authentication credentials were provided in the request, the server considers them insufficient to grant access.

あとOAuthのスペックにもtokenエラーに401使うように書いてあるので（いますぐには関係ありませんが関係あるようになる）

https://datatracker.ietf.org/doc/html/rfc6750.html#section-3.1

>   invalid_token
>          The access token provided is expired, revoked, malformed, or
>          invalid for other reasons.  The resource SHOULD respond with
>          the HTTP 401 (Unauthorized) status code.  The client MAY
>          request a new access token and retry the protected resource
>          request.

403も同じく

>    insufficient_scope
>          The request requires higher privileges than provided by the
>          access token.  The resource server SHOULD respond with the HTTP
>          403 (Forbidden) status code and MAY include the "scope"
>          attribute with the scope necessary to access the protected
>          resource.

## Checklist
- [ ] Read the [contribution guide](https://github.com/misskey-dev/misskey/blob/develop/CONTRIBUTING.md)
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/server/api/ApiCallService.ts
+++ b/packages/backend/src/server/api/ApiCallService.ts
@@ -83,7 +83,7 @@ export class ApiCallService implements OnApplicationShutdown {
 			}
 		}).catch(err => {
 			if (err instanceof AuthenticationError) {
-				this.send(reply, 403, new ApiError({
+				this.send(reply, 401, new ApiError({
 					message: 'Authentication failed. Please ensure your token is correct.',
 					code: 'AUTHENTICATION_FAILED',
 					id: 'b0a7f5f8-dc2f-4171-b91f-de88ad238e14',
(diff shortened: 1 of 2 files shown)


syuilo:
👍

PR #11081, opened 2023-07-03
[enhance] Use background image on Misskey - Update style.scss

ncwhale:
## What
Add default background image styles.

For better show an image as centered, cover size, and no repeat.

And can be modified by CSS vars.

## Why

By default, if we setup an image as background for Misskey, it will just do not show it at all. When process with style settings and found that the background image was not fit to the page viewpoint.

So I create this patch for better usage of bit images as background, and also can be controlled by CSS vars as other styles.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/frontend/src/style.scss
+++ b/packages/frontend/src/style.scss
@@ -23,6 +23,10 @@
 
 html {
 	background-color: var(--bg);
+	background-size: var(--bg-size, cover);
+	background-repeat: var(--bg-repeat, no-repeat);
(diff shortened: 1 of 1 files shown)


xtexx (approved):
lgtm

ncwhale:
Fix #11084

syuilo:
デザイン上の都合から、現在壁紙の設定はサポートされなくなりました 🙏🏻

PR #11142, opened 2023-07-06
enhance: Add OGP data for notes with multiple images/videos

alexwh:
## What

Return all og:image or og:video meta tags based on the content of the note, rather than just the first match (`Array.find` vs `Array.filter`).

## Why

When misskey note links are embedded in external applications like Discord, only the first image was displayed.

## Additional info (optional)

[OGP supports multiple elements with arrays](https://ogp.me/#array).

I'm new to Pug and Misskey, so please suggest or make changes as needed (allow edits by maintainers is on).

## Checklist
- [ ] (If possible) Add tests


packages/backend/src/server/web/views/note.pug
```diff
@@ -19,15 +19,17 @@ block og
...
 	meta(property='og:description' content= summary)
 	meta(property='og:url'         content= url)
-	if video
-		meta(property='og:video:url'        content= video.url)
-		meta(property='og:video:secure_url' content= video.url)
-		meta(property='og:video:type'       content= video.type)
-		// FIXME: add width and height
-		// FIXME: add embed player for Twitter
-	if image
+	if videos
```
acid-chicken:
```suggestion
	if videos.length
```

packages/backend/src/server/web/views/note.pug
```diff
@@ -19,15 +19,17 @@ block og
...
+		each video in videos
+			meta(property='og:video:url'        content= video.url)
+			meta(property='og:video:secure_url' content= video.url)
+			meta(property='og:video:type'       content= video.type)
+			// FIXME: add width and height
+			// FIXME: add embed player for Twitter
+	if images
```
acid-chicken:
```suggestion
	if images.length
```

acid-chicken (commented):
maybe

alexwh:
Good point, I didn't know that empty arrays were truthy until now (thanks javascript).

tirr-c:
I've applied this patch to my server, but Discord still shows only the first one. Rendered pages were ok; there were multiple `og:image`'s when I checked it with curl.

alexwh:
> I've applied this patch to my server, but Discord still shows only the first one. Rendered pages were ok; there were multiple `og:image`'s when I checked it with curl.

It seems Discord's implementation of Open Graph doesn't respect image arrays, and the only way to attach multiple images to an embed is [with a bot message](https://github.com/discord/discord-api-docs/issues/1643). I can't find any documentation on Discord's side for what properties are supported or how they're implemented, other than ["it's OGP"](https://twitter.com/discord/status/952834253842153472).

I think the only way to do this specifically for Discord Open Graph embeds would be to dynamically combine together the valid images into a single image with an image manipulation library and instead use that as the `og:image` tag. [BetterTwitFix](https://github.com/dylanpdx/BetterTwitFix) does this. Since the complexity of doing this is much higher compared to this original PR, I'll just leave it as an idea for now.

I'm not sure what other effects this PR has on other sites, but the maintainers can decide if this is still worth merging or not.

PR #11275, opened 2023-07-14
enhance(backend): add unix socket support

ybw2016v:
## What

Add unix socket support, to make it possible to let misskey listen on the  unix socket, which may improve backend performance.

## Why

The unix socket is an inter-process communication mechanism and it can avoid some unnecessary checks and operations, which makes them faster and lighter than listening on the TCP/IP loopback port if the reverse proxy and misskey are running on the same machine. 

## Additional info (optional)

https://serverfault.com/questions/124517/what-is-the-difference-between-unix-sockets-and-tcp-ip-sockets

## Checklist
- [ ] (If possible) Add tests


syuilo:
👍🏻

PR #11293, opened 2023-07-16
Export notes with file detail

CGsama:
## What

Add file field to the export notes like in "show" endpoint, so user can follow the url to extract file later.

## Why

Current export feature only contains fileid not the actual file nor url, its hard for import later and archiving.

## Checklist
- [ ] (If possible) Add tests


syuilo:
👍🏻

PR #11367, opened 2023-07-24
fix(backend): Use BullMQAdapter for bull-board

hutchisr:
## What

Use BullMQAdapter instead of BullAdapter since misskey now uses bullmq instead of bull

## Why

Fixes job dashboard to no longer give HTTP 500 errors when trying to perform actions

## Checklist
- [ ] (If possible) Add tests


syuilo:
👍

PR #11483, opened 2023-08-07
feat: Add option to expand wide emojis

pbzweihander:
## What

Add an option to expand wide emojis in the emoji picker

## Why

In Misskey, we use wide emojis a lot. But if the multiple wide emojis have the same front section, it is very hard to distinguish between them.

Before:

screenshot_2023-08-07_22-29-59

It is very hard to distinguish between the first three emojis.

After:

screenshot_2023-08-07_22-30-26

Now we can clearly see which emoji is which.

## Additional info (optional)

It is my first time writing Vue code. Feel free to add any subtle comments!

## Checklist
- [ ] (If possible) Add tests


packages/frontend/src/components/MkEmojiPicker.vue
```diff
@@ -470,6 +472,15 @@ defineExpose({
 						min-width: 0;
 					}
 				}
+
+				> .body.reactionPickerExpandWide {
+					display: flex;
+					flex-wrap: wrap;
+
+					> .item {
+						aspect-ratio: unset;
```
tirr-c, line 481: Maybe it would be better flip the flag (`reactionPickerSquare = !reactionPickerExpandWide` or something?) so that we don't need to use `unset`.

```css
> .body {
  display: flex;
  /* ... */

  &.reactionPickerSquare {
    display: grid;
    /* ... */
    > .item {
      aspect-ratio: 1 / 1;
      /* ... */
    }
  }
}
```

packages/frontend/src/components/MkEmojiPicker.vue
```diff
@@ -470,6 +472,15 @@ defineExpose({
 						min-width: 0;
 					}
 				}
+
+				> .body.reactionPickerExpandWide {
+					display: flex;
```
saschanaz, line 477: This overrides `display: grid` of `asDrawer` and thus ignores `reactionPickerWidth` option, I wonder how could that be fixed 🤔

Sayamame-beans:
triage: Related to #12469

pbzweihander:
I think this PR is stale. So closing it.

PR #11657, opened 2023-08-10
feat: Removing stack trace info in production env

MomentQYC:
## What

Remove stack trace information from production errors, and replace a small number of unnecessary errors with other types of output.

## Why

This is what `instance/queue/` currently looks like when accessed without logging in (e.g. `misskey.io/queue`)
image
Compare the changed look
image
Many times in production environments stack trace information is unnecessary and a security risk.
For a small number of output changes, I think an average user would prefer a program that only outputs logs as opposed to a program that reports errors when a login action is required.

## Checklist
- [ ] (If possible) Add tests


tamaina:
I think it's better to use ApiError in API execution service

tamaina:
There should be no need to change the frontend

tamaina:
I don't know why the test is failing

tamaina:
Somehow Jest can't use `./error.js`

MomentQYC:
> There should be no need to change the frontend

The main consideration is to facilitate centralized management and reduce unnecessary errors.

MomentQYC:
> > There should be no need to change the frontend
> 
> The main consideration is to facilitate centralized management and reduce unnecessary errors.

For Example, https://github.com/misskey-dev/misskey/blob/827616f63094997bd44458b8c072a50da176af32/packages/frontend/src/scripts/please-login.ts#L25C3-L25C3

tamaina:
> facilitate centralized management and reduce unnecessary errors.

I don't want them on the frontend.

tamaina:
I think it is not preferable to remove stack traces or change Errors to console.log on the front end.

MomentQYC:
> I think it is not preferable to remove stack traces or change Errors to console.log on the front end.

I have modified it to use `console.log` in the production environment, while still throwing an error in other environments. Is this feasible?

tamaina:
Do not rewrite `throw Error`s in frontend. They are there to stop calling(upstream) functions, for example.

MomentQYC:
> Do not rewrite throw Errors in frontend. They are there to stop calling(upstream) functions, for example.

 OK, I will remove all frontend changes first.

tamaina:
This PR should not include frontend changes.

saschanaz:
I'm not sure we want `throw ErrorHandling`, because errors can be from anywhere, from libraries and native JS functions. What we probably want is to customize Fastify error handler: https://fastify.dev/docs/latest/Reference/Server/#seterrorhandler

saschanaz (changes requested):
As above.

MomentQYC:
> I'm not sure we want `throw ErrorHandling`, because errors can be from anywhere, from libraries and native JS functions. What we probably want is to customize Fastify error handler: https://fastify.dev/docs/latest/Reference/Server/#seterrorhandler

I would like to address the existing problems first and then focus on improving error-related problems in the future. Since I'm not very familiar with TypeScript, it might take some time for me to make changes to the code (if necessary).

saschanaz:
>I would like to address the existing problems first 

I highly suspect "addressing the existing problems" requires more code than writing proper error handler, which should require change in only one file.

MomentQYC:
> As above.

Like now?
Defined errors still remove tracking information as before, undefined errors are processed further.

saschanaz:
There's probably no need to do setErrorHandler in every createServer, you should be able to do it only once in ServerService.

saschanaz:
Given that this PR now sets an error handler that drops stack info, the changes in other files are not needed, right?

MomentQYC:
> Given that this PR now sets an error handler that drops stack info, the changes in other files are not needed, right?

My original intention was to minimize stack trace information in production environments as much as possible, and in particular, as much as possible, not to expose the user to stack trace information especially from the back-end part.
So I think especially in scenarios like the one in the very first screenshot, targeted removal of stack trace information is necessary, other than that it's optional.

saschanaz:
You mean there will be any situation that the stack will still be exposed even with the error handler?

MomentQYC:
> You mean there will be any situation that the stack will still be exposed even with the error handler?

When I tested along the lines of the code here in the production environment, I still found the stack trace.
https://github.com/misskey-dev/misskey/blob/ab58b651f79e182c20a238f9c07f0c0006a58599/packages/backend/src/server/web/ClientServerService.ts#L148C5-L151C6

saschanaz:
Hmm, that's specific to Misskey's bull-board usage where [the library registers its own error handler](https://github.com/felixmosh/bull-board/blob/e0c176a92eb612d73e7c402e3fa7ba59c344920e/packages/fastify/src/FastifyAdapter.ts#L151-L154) (and thus yours is being ignored). It seems reasonable to not throw an error but explicitly send an object instead in https://github.com/misskey-dev/misskey/blob/ab58b651f79e182c20a238f9c07f0c0006a58599/packages/backend/src/server/web/ClientServerService.ts#L148C5-L151C6, as [their example does](https://github.com/felixmosh/bull-board/blob/e0c176a92eb612d73e7c402e3fa7ba59c344920e/examples/with-fastify-auth/cookieAuth.js#L66), and later investigate this bull-board issue separately.

For other cases the error handler should be sufficient.

MomentQYC:
This is where it's most likely to be seen by the user, so here's where I'd like to apply additional error handling (removing the stack trace).
This has been changed to `reply.code().send()` in the latest changes.
Like this:
image

syuilo:
やる場合、new Errorの使用を禁止する何らかの機構(lintとかgithub actionsとか)が無いとこのPRの目的を維持するのは難しそう

saschanaz:
> This is where it's most likely to be seen by the user, so here's where I'd like to apply additional error handling (removing the stack trace).
> This has been changed to `reply.code().send()` in the latest changes.
> Like this:
> image

What I'm saying is that it's doable with error handler, it's the sole purpose of error handler anyway. The reason it's not working for bull-board one is explained above, and I suggest using send() only for that exception for now and reverting others. And maybe add some test by adding some test-only route.

MomentQYC:
> reverting others

Do you mean like now?

packages/backend/src/error.ts
```diff
@@ -0,0 +1,6 @@
+/*
+ * SPDX-FileCopyrightText: tamaina and other misskey contributors
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+// Never use `./error.js` because jest can't use it.
```
saschanaz: I guess this file is redundant?

packages/backend/src/misc/fastify-reply-error.ts
```diff
@@ -12,5 +12,9 @@ export class FastifyReplyError extends Error {
 		super(message);
 		this.message = message;
 		this.statusCode = statusCode;
+
+		if (process.env.NODE_ENV === 'production') {
+			Object.defineProperty(this, 'stack', { value: '' });
```
saschanaz: I guess `this.stack = undefined` is more consistent.

packages/backend/src/server/oauth/OAuth2ProviderService.ts
```diff
@@ -353,6 +353,7 @@ export class OAuth2ProviderService {
 	public async createServer(fastify: FastifyInstance): Promise<void> {
 		// https://datatracker.ietf.org/doc/html/rfc8414.html
 		// https://indieauth.spec.indieweb.org/#indieauth-server-metadata
+
```
saschanaz: Redundant

packages/backend/src/misc/error.ts
```diff
@@ -0,0 +1,20 @@
...
+		error.stack = undefined;
+	}
+	if (reply) {
+		reply.code(statusCode ?? 500);
+	}
+	return error;
+}
+
+export function ErrorHandler(error: Error, request: FastifyRequest, reply: FastifyReply): void {
+	throw ErrorHandling(error.message, reply);
```
saschanaz: Merge these two functions?

packages/backend/src/misc/error.ts
```diff
@@ -0,0 +1,20 @@
...
+ */
+import { FastifyReply, FastifyRequest } from 'fastify';
+
+export function ErrorHandling(message: string, reply?: FastifyReply, statusCode?: number): Error {
+	const error = new Error(message);
+	if (process.env.NODE_ENV === 'production') {
+		error.stack = undefined;
+	}
+	if (reply) {
+		reply.code(statusCode ?? 500);
```
saschanaz: You can skip this if you do `reply.send(error)`, the send function will pass the error to the default error handler where it'll process the status code.

saschanaz (changes requested):

packages/backend/src/misc/error.ts
MomentQYC: The lack of merging here is in consideration of possible separate application scenarios.

saschanaz: We can split it again if it's needed later of course, but in the current state I see no reason for that.

saschanaz:
Hmm, I locally tested what Fastify does by adding a random error, but it seems Fastify already does not expose error stack information by default at all. The [`details` field you saw in `/queue` was from bull-board custom error handler](https://github.com/felixmosh/bull-board/blob/e0c176a92eb612d73e7c402e3fa7ba59c344920e/packages/api/src/handlers/error.ts#L9) and nothing to do with Fastify default behavior, which only adds `statusCode`, `error`, and `message`.

tl;dr: we don't need setErrorHandler at all, this PR should be sufficient only with changes on `ClientServerService.ts`.

MomentQYC:
> we don't need setErrorHandler at all

Reverted

saschanaz (approved):
👍

syuilo:
👍🏻

PR #11711, opened 2023-08-12
fix: word mute is not applied to sub note

paring-chan:
## What

This makes sub note do the same behavior when note is muted

image

## Why

The word note feature is applied only on normal notes, not on sub notes(example: reply target)

## Additional info (optional)

None.

## Checklist
- [x] (If needed) Add story of storybook
- [x] (If possible) Add tests


tamaina (approved):

syuilo:
👍

PR #11742, opened 2023-08-19
fix(backend): change character limits to allow for long instance domains

AAGaming00:
## What
This PR changes the max length for hosts of various entities to 512 characters, and the max length of favicon and icon URLs for instances to 4096 characters, to allow federation to work properly with instances that have long domains.

## Why
Federation with instances with large domains such as [this one](https://helloeverybodymynameismarkiplierandwelcometofivenightsatfreddys.anindiehorrorgamethatyouguyssuggestedenmasseandisawthatyamimash.playeditandhesaidthatitwasreallyreallygoodsoimveryeagertoseewha.tisupandthatisaterrifyinganimatronicbear.fnaf.stream) is broken due to database fields having too short of a character limit

## Additional info (optional)
There's probably a better way to do this, but i've tested this and it works fine. I'm open to suggestions if there's a better way to fix this.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/migration/1692374635734-IncreaseHostCharLimit.js
+++ b/packages/backend/migration/1692374635734-IncreaseHostCharLimit.js
@@ -1,0 +1,47 @@
+export class IncreaseHostCharLimit1692374635734 {
+		name = 'IncreaseHostCharLimit1692374635734'
+
+		async up(queryRunner) {
+				await queryRunner.query(`ALTER TABLE "drive_file" ALTER COLUMN "userHost" TYPE character varying(512)`);
+				await queryRunner.query(`ALTER TABLE "user" ALTER COLUMN "host" TYPE character varying(512)`);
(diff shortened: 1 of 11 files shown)


AAGaming00:
long instances in usernames also seem to cause issues refreshing profile pages as the router doesn't recognize the path. unsure how to debug that.

syuilo:
例に挙げられたような長さのドメインは非常識的で実用的ではないからサポートの必要性があるのか疑問に感じる

AAGaming00:
True, but it does cause server-side errors so *shrug*

tobskep:
Yeah

tobskep:
Also long domains aren't insane, some intranets have pretty crazy domains

syuilo:
Even if a very long domain is allowed, the mentions will also become very long, causing problems when creating posts, and also making the UI very difficult to view, requiring consideration and modification in various areas.
So we have decided not to change the current character limit 🙏🏻

PR #11777, opened 2023-09-03
Respond .well-known/host-meta.json with application/json

xtexx:
According to RFC 6415 appendix-A.
>   The server
   MUST include the HTTP "Content-Type" response header field with a
   value of "application/json".  Any other "Content-Type" value (or lack
   thereof) indicates that the server does not support the JRD format.

`application/jrd+json` is only used by WebFinger (RFC 7033)


syuilo:
🙏

PR #11799, opened 2023-09-07
fix(backend): timeout when querying mentions

zekexy:
## What

I use a trick to avoid using primary key to solve a slow query problem. Is there any other better way?

## Why

For queries like this, postgresql sometimes decides to scan the primary key instead of using the index on "visibleUserIds" field
```SQL
EXPLAIN
SELECT
	"note"."id" AS "note_id"
FROM
	"note" "note"
WHERE
	'{"01GWCS86KKFQX39VH79KMD48JS"}' <@"note"."visibleUserIds"
ORDER BY
	"note"."id" DESC
LIMIT 10
```
However, postgresql doesn't know that sometimes users' data distribution is very uneven, which is the reason for the timeout problems some people have.

## Additional info (optional)

Expected to be fast, actually very slow :
image

Satisfactory speed :
image

## Checklist
- [ ] (If possible) Add tests


syuilo:
🙏🙏🙏

zekexy:
https://github.com/misskey-dev/misskey/issues/7632 may be solved

PR #12051, opened 2023-10-16
fix: Refetch user keys when HTTP Signature validation fails

erincandescent:
## What
If a user has had a key rotation, and nobody on this server follows that user, we will not receive the Update activity with the new key

Therefore, when we encounter key validation errors we should check for an up-to-date key.

## Why
Currently, federation will break with accounts on remote servers which have no local followers as the key update activity will not be received

This can happen if, for example, a Mastodon admin runs `tootctl accounts rotate`

## Additional info (optional)
Behaviour mirrors other implementations:

 * [Mastodon](https://github.com/mastodon/mastodon/blob/fc9ab61448af94513524f14f5fd287d2a26ceeab/app/controllers/concerns/signature_verification.rb#L96)
 * [Akkoma](https://akkoma.dev/AkkomaGang/http_signatures/src/branch/main/lib/http_signatures/http_signatures.ex#L46)

This is a port of [Iceshrimp PR 327](https://iceshrimp.dev/iceshrimp/iceshrimp/pulls/327), and has been verified as resolving federation issues there

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/core/activitypub/ApDbResolverService.ts
+++ b/packages/backend/src/core/activitypub/ApDbResolverService.ts
@@ -164,6 +164,19 @@ export class ApDbResolverService implements OnApplicationShutdown {
 		};
 	}
 
+	/**
+	 * Miskey User -> Refetched Key
+	 */
+	@bindThis
+	public async refetchPublicKeyForApId(user: MiRemoteUser): Promise<MiUserPublickey | null> {
+		await this.apPersonService.updatePerson(user.uri!);
+		const key = this.userPublickeysRepository.findOneBy({ userId: user.id });
+		if (key != null) {
+			await this.publicKeyByUserIdCache.set(user.id, key);
+		}
+		return key;
+	}
+
 	@bindThis
 	public dispose(): void {
 		this.publicKeyCache.dispose();
--- a/packages/backend/src/queue/processors/InboxProcessorService.ts
+++ b/packages/backend/src/queue/processors/InboxProcessorService.ts
@@ -104,7 +104,18 @@ export class InboxProcessorService {
 		}
 
 		// HTTP-Signatureの検証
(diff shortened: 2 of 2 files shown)


u1-liquid:
In my opinion, we should not automatically update the remote user's key when we fail to verify. This change would undermine the last line of defense in the ActivityPub design that protects whether the remote user is genuine. Perhaps, just like in Mastodon, it would be better to only update it when explicitly specified by the administrator.

erincandescent:
> Perhaps, just like in Mastodon, it would be better to only update it when explicitly specified by the administrator.

This is not Mastodon behaviour - [you can follow the code I linked](https://github.com/mastodon/mastodon/blob/fc9ab61448af94513524f14f5fd287d2a26ceeab/app/controllers/concerns/signature_verification.rb#L96)

erincandescent:
I agree it would be nice if there were a better way of validating key authenticity, but there isn't for users which don't have an active following relation

(If there is a following relation, then an Update activity will be received with the new key, signed with the old)

Mar0xy:
Is this actually usable if so I would go and merge it into our fork

erincandescent:
I haven't yet gotten to the bottom of why there are lint issues (I'm not a TypeScript programmer at all), but it should work.

Its already merged into the FireFish and IceShrimp forks

Mar0xy:
I have implemented it now in a bit of a different way than in this PR as can be checked via this [search query](https://github.com/search?q=repo%3Atransfem-org%2FSharkey+33eab0d178a2b20aa2935914fe7b7e97e3bea948+-+c76f0f1679459349e9362dc01a9f617a8ae8cfe3+-+71b7c31958e2ce11a4b5a11a5c282ca3bdcb41dc&type=commits)

Mainly due to the fact that the first info if it errors would throw a error causing bull to stop processing it all together also blocking the second if statement from ever running.

PR #12118, opened 2023-10-22
fix(frontend): add disableObserver property to MKPagination to avoid the glitch in admin/federation page

backrunner:
## What

This PR add a new property called `disableObserver` to `MKPagination`, it will disable the internal `IntersectionObserver` in the `MKPagination` and replace it with scroll listener chained methods to keep the `executeQueue` method working.

### Bug: Federation page glitching when the grid has three columns

The `MKPagination` contains an IntersectionObserver which will observe the `rootEl` in `MKPagination`, which works well in the timeline. But in `admin/federation`, the IntersectionObserver will cause a glitch, because the `rootMargin` is actually the container shift up in 100% of its height, and the `MKPagination` will never be intersected with the scrollable container with shifted rect.

Related code:

```ts
scrollObserver = new IntersectionObserver(entries => {
  backed = entries[0].isIntersecting;
}, {
  root: scrollableElement,
  rootMargin: getObserverRootMargin(),
  threshold: 0.01,
});
```

Because the `entries[0].isIntersecting` will be `false` constantly, in the next lines, the `scrollRemove` method will be called.

```ts
scrollRemove = (props.pagination.reversed ? onScrollBottom : onScrollTop)(contentEl, executeQueue, TOLERANCE);
```

That will cause the existed items be removed, and cause glitch on the page.

#### Bug Replication

1. Navigate to `/admin/federation`, when there's 2/3 columns, scrolling down the page.
2. The new items disappeared suddenly and just disappeared.

### Bug: incorrect offset value

There's another wrong implementation in `MKPagination`, the `offset` value won't be reset when the `scrollRemove` has been called, it will cause the disappeared items cannot be loaded again until the user refresh the whole page.

This issue has been fixed in this PR.

### About new property

New features will not be enabled until the `disableObserver` has been set to `true`, if some other scenarios have same issue, enable this property might be helpful.

The new method to detect whether the element back to the top/bottom is use a scroll listener to detect the scroll behaviour, it determine whether the user has left the top/bottom, then set `backed` to `false` (I modified the default value to `true` because the element should be the top/bottom of the container in this case, as for the original function, the `IntersectionObserver` will update the correct value to `backed` and trigger the watcher to setup the listener for `executeQueue`).

Then I using another `onScrollTop` and `onScrollBottom` (which disabled the first check when listener just setup) to detect if the user backed to top or bottom. 

### About the federation page

Because of the initial query limit is way below than the max count that the screen can display, if `executeQueue` refresh the items in `MKPagination`, to users, there's still a small glitch (last items disappeared suddenly). In my opinion, to call `executeQueue` to refresh items in this page in not that meaningful, if this step can be bypassed, the experience in this page will be better. 

### About the quality of this PR (Important)

**Because the default value of `backed` has been modified to `true`, it might cause some unexpected side effects to the original functions due to the `MKPagination` is a common component, so there should be more tests before officially publish. Currently I do not have that much time to test all the scenarios that used `MKPagination`, so this PR might be harmful to some functions.**

A harmless way to solve the issue is **just disable the `IntersectionObserver`**, if choose this way to solve the issue, you can close this PR and simply fix with another patch.

## Checklist
- [ ] (If possible) Add tests


backrunner:
I messed up the branches, reopened at #12119

PR #12248, opened 2023-11-05
fix(backend): isBot not being set on `Application` type

Mar0xy:
## What

This PR makes it so that accounts via AP will have isBot set to true if their type is "Application"

## Why

There are softwares out there which return the type of "Application" instead of "Service"
Example: https://relay.an.exchange/inbox
image

## Checklist
- [ ] (If possible) Add tests


syuilo:
👍🏻

PR #12273, opened 2023-11-06
fix: show real instance url

Candinya:
## What

implements #12226 in another way

## Why

Current logic of binding instance's frontend host with its real host will confuse users about identifying the instance's real domain. So it might be better to grab from backend.
Using a html meta tag to pass parameters, so we can prevent async requests in config.ts .

## Additional info (optional)

Need some help in fixing the test (don't know why it fails)

## Checklist
- [ ] (If possible) Add tests


syuilo:
👍👍

syuilo:
これcypressが落ちてるのはなぜかしら

Candinya:
> これcypressが落ちてるのはなぜかしら

Me too, since it's working fine on real (dev) environment, I have no idea why this breaks the cypress test

Candinya:
I think I know why - API endpoint depends on the `address`, and when this address is not available, the API fails.
Using relative paths like `location.href` or `/` for API should be fine.

image

Will create a new PR to fix this

Candinya:
fixed in #12322

PR #12493, opened 2023-11-28
fix(frontend): Update the Vitest configuration

woxtu:
## What

Vitestの設定を更新して次の警告を解決します：

```
Vitest  "deps.inline" is deprecated. If you rely on vite-node directly, use "server.deps.inline" instead. Otherwise, consider using "deps.optimizer.web.include"
```

## Checklist
- [ ] (If possible) Add tests


syuilo:
🙏

PR #11830, opened 2023-09-13
fix(backend): enhance nodeinfo by prpoagate the standart keys for homepage and repository

6543:

PR #12503, opened 2023-11-29
fix(backend): enhance nodeinfo by export instance admin via nodeAdmins key

6543:
https://codeberg.org/thefederationinfo/nodeinfo_metadata_survey


6543:
followup of #11830

packages/backend/src/server/NodeinfoServerService.ts
```diff
@@ -96,6 +96,11 @@ export class NodeinfoServerService {
 				metadata: {
 					nodeName: meta.name,
 					nodeDescription: meta.description,
+					nodeAdmins: [{
+						name: meta.maintainerName,
+						email: meta.maintainerEmail,
+					}],
+					// deprecated
```
acid-chicken, line 103: Can you please provide a source?

6543: if used internally its not deprecated else we should do so.

I try to stream line the nodeinfo across different federated projects.
the endgame is to release a new nodeinfo spec ... but that will take it's time.

If there is interest just visit https://github.com/jhass/nodeinfo/issues/70 ...

6543:
@acid-chicken it's also fin to not depricate it ... just tell :)

acid-chicken (approved):

syuilo:
🙏

PR #12619, opened 2023-12-10
Update style.scss

kasuganosora:
Use the classic font list to solve the problem that Chinese and English look narrower in width.

before
b2

after
eZ0v9GOXQHIc

### DIFF
--- a/packages/frontend/src/style.scss
+++ b/packages/frontend/src/style.scss
@@ -35,7 +35,7 @@ html {
 	accent-color: var(--accent);
 	overflow: auto;
 	overflow-wrap: break-word;
-	font-family: 'Hiragino Maru Gothic Pro', "BIZ UDGothic", Roboto, HelveticaNeue, Arial, sans-serif;
+	font-family: Roboto, HelveticaNeue, Arial, sans-serif;
 	font-size: 14px;
 	line-height: 1.35;
 	text-size-adjust: 100%;


packages/frontend/src/style.scss
```diff
@@ -35,7 +35,7 @@ html {
 	accent-color: var(--accent);
 	overflow: auto;
 	overflow-wrap: break-word;
-	font-family: 'Hiragino Maru Gothic Pro', "BIZ UDGothic", Roboto, HelveticaNeue, Arial, sans-serif;
+	font-family: Roboto, HelveticaNeue, Arial, sans-serif;
```
kakkokari-gtyih, line 38:
```suggestion
	font-family: Roboto, HelveticaNeue, 'Segoe UI Variable Text', 'Segoe UI', Arial, 'Hiragino Maru Gothic Pro', 'BIZ UDGothic', sans-serif;
```

Just change the priority and it should be fine (removing Japanese fonts makes it weird when displaying Japanese)

And I added some Windows-related fonts to make it look nicer (but it's up to you for that change)

acid-chicken:
See #6858

packages/frontend/src/style.scss
kakkokari-gtyih: > See https://github.com/misskey-dev/misskey/issues/6858

そうなんだ

PR #12743, opened 2023-12-22
fix: email verify enable logic

Candinya:
## Why

`meta.verifymailAuthKey` -> affects `verifymailApi` -> already handled above
`meta.enableActiveEmailValidation` -> affects whether enable email verify logic -> already handled in if situations

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/core/EmailService.ts
+++ b/packages/backend/src/core/EmailService.ts
@@ -167,18 +167,18 @@ export class EmailService {
 		const verifymailApi = meta.enableVerifymailApi && meta.verifymailAuthKey != null;
 		let validated;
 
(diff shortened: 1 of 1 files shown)


Mar0xy:
I feel like this should be merged for #12564 as this fixes an issue where having validation enabled but no key for verifymail set causing it to not run the else statement to check via the node package instead as it just skips everything else and always defaults to valid.

syuilo:
🙏

syuilo:
lintが失敗してそう

Candinya:
> lintが失敗してそう

should be fixed in #12761

PR #12850, opened 2023-12-29
feat: Add support for TrueMail

MomentQYC:
## What

Partially related to #12741 

## Why

Provide a freely configurable method that avoids expensive costs and achieves a close approximation.

## Additional info (optional)

1. Thanks to @MarryDream for the guide and test
2. This feature relies on this project: https://github.com/truemail-rb/truemail-rack (https://github.com/truemail-rb/truemail-rack-docker-image) ,users need to configure it correctly
3. Perhaps documentation is needed for clarification, but I'm not fluent in Japanese

## Checklist
- [x] Read the [contribution guide](https://github.com/misskey-dev/misskey/blob/develop/CONTRIBUTING.md)
- [x] Test working in a local environment
- [ ] (If needed) Add story of storybook
- [x] (If needed) Update CHANGELOG.md
- [ ] (If possible) Add tests

PR #12862, opened 2023-12-30
fix: Some fixes for #12850

MarryDream:
- refinement the error message when trueMail validation fails
- the settings of trueMail are not displayed after saving
- changing how `Active Email Validation` is saved

## What

fix: Some fixes for #12850
- refinement the error message when trueMail validation fails
- the settings of trueMail are not displayed after saving
- changing how `Active Email Validation` is saved

## Why

Addressing some of the issues that arose with #12850

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/core/EmailService.ts
+++ b/packages/backend/src/core/EmailService.ts
@@ -165,10 +165,17 @@ export class EmailService {
 			email: emailAddress,
 		});
 
+		if (exist !== 0) {
+			return {
+				available: false,
+				reason: 'used',
+			};
+		}
+
 		let validated: {
 			valid: boolean,
 			reason?: string | null,
-		};
+		} = { valid: true, reason: null };
 
 		if (meta.enableActiveEmailValidation) {
 			if (meta.enableVerifymailApi && meta.verifymailAuthKey != null) {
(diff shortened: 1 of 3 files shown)


kakkokari-gtyih:
> refinement the error message when trueMail validation fails

Just run `pnpm build-misskey-js-with-types` 🙏

MarryDream:
> > refinement the error message when trueMail validation fails
> 
> Just run `pnpm build-misskey-js-with-types` 🙏

Got it, I've executed the command and re-committed it.

packages/backend/src/core/EmailService.ts
```diff
@@ -156,7 +156,7 @@ export class EmailService {
 	@bindThis
 	public async validateEmailForAccount(emailAddress: string): Promise<{
 		available: boolean;
-		reason: null | 'used' | 'format' | 'disposable' | 'mx' | 'smtp' | 'banned' | 'network' | 'blacklist';
+		reason: null | string;
```
u1-liquid: I think this shouldn't be to changed 'just stirng', if there is another 'reason' value variants, please append it to here.

packages/backend/src/core/EmailService.ts
```diff
@@ -165,10 +165,17 @@ export class EmailService {
 			email: emailAddress,
 		});
 
+		if (exist !== 0) {
```
u1-liquid, line 168:
```suggestion
		if (exist > 0) {
```

packages/backend/src/core/EmailService.ts
```diff
@@ -185,33 +192,38 @@ export class EmailService {
...
-			null,
+			available: true,
+			reason: null,
 		};
 	}
 
 	private async verifyMail(emailAddress: string, verifymailAuthKey: string): Promise<{
 		valid: boolean;
-		reason: 'used' | 'format' | 'disposable' | 'mx' | 'smtp' | null;
+		reason: 'used' | 'format' | 'disposable' | 'mx' | 'smtp' | 'key' | null;
```
u1-liquid: If the request failed due to the API key, wouldn't it be better to throw an error rather than showing it to the general user?

packages/backend/src/core/EmailService.ts
```diff
@@ -239,6 +252,12 @@ export class EmailService {
 			related_domains: string[];
 		};
 
+		if (json.message) {
```
u1-liquid: Does 'message' indicates API key error in all cases?

packages/backend/src/core/EmailService.ts
```diff
@@ -284,22 +303,29 @@ export class EmailService {
...
 					list_match?: string;
 					regex?: string;
 					mx?: string;
 					smtp?: string;
 				} | null;
 			};
-			
-			if (json.email === undefined || (json.email !== undefined && json.errors?.regex)) {
+
+			if (json.error) {
```
u1-liquid: Does 'error' indicates API key error in all cases?

packages/misskey-js/src/autogen/apiClientJSDoc.ts
```diff
@@ -1,6 +1,6 @@
 /*
  * version: 2023.12.2
- * generatedAt: 2024-01-02T08:53:57.449Z
+ * generatedAt: 2024-01-02T13:57:41.524Z
```
u1-liquid: It seems that no actual changes have been made to the API, so it might be better not to have changes in this folder in this PR.

packages/backend/src/core/EmailService.ts
MarryDream: When the api usage exceeds the limit, will get `{ message: "error message content" }`, returned with only the `message` attribute. I've reworked the judgment method

packages/backend/src/core/EmailService.ts
MarryDream: When such an error occurs, the status code of the response will not be `200`. I looked at the implementation of the `httpRequestService` and it already throws an error in this case. So I removed the judgment here

packages/misskey-js/src/autogen/apiClientJSDoc.ts
MarryDream: Running `pnpm build-misskey-js-with-types` changed the timing of all type files. After updating the latest `develop` branch, I realized that someone else had already generated the correct types. So I'm undid the changes I made to the types files.

MarryDream:
@syuilo

syuilo:
🙏

PR #13009, opened 2024-01-15
Fix: properly handle cc followers

### LINKED ISSUE #13001: cc:Followers gets converted to DM without recipients
### 💡 Summary

i have an Akkoma MRF that drops as:Public from all activities, and so both "public" and "unlisted" posts get converted to a "followers-only" post. however, there is an issue with misskey parsing the post if it was submitted as "public". the post does not show up in misskey timelines, as it is instead converted to a DM with no recipients. (if the post was submitted as "unlisted", it works correctly after the MRF.) (also, both posts work correctly in mastodon and pleroma.)

### 🥰 Expected Behavior

parsed as a followers-only post

### 🤬 Actual Behavior

parsed as a direct post with no recipients

### 📝 Steps to Reproduce

process an activity/object like so

```
cc: <followers>
```

### 💻 Frontend Environment

```markdown
* Server URL: nya.social
* Misskey: 2023.12.1
```

_No response_

trwnh:
Fix #13001

## What

Check for cc:followers, not just to:followers

## Why

When encountering a post with no `to` and only `cc: <followers>`, it now gets parsed correctly as a followers-only post

## Checklist
- [ ] (If possible) Add tests


syuilo:
🙏

PR #13145, opened 2024-02-02
keep cached avatar&banner when refresh fails to get new values

dakkar:
## What
when the remote explicitly tells us a user image is gone, we remove our cached value, but if we fail to get the image, we keep whatever value we already have

## Why
this should minimise the problem of avatars randomly disappearing

## Additional info (optional)

tested this way:
* fetched a remote user, avatar was set ok
* removed avatar id+url from the db, refreshed remote data, avatar was set again correctly
* set banner url in the db to a random value, refreshed remote data, banner url was null-ed correctly
* faked an error in the code that fetches remote images, refreshed remote data, the avatar was not removed from the database

I initially wrote this patch for Sharkey, see https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/381 but I think it would be useful for Misskey as well.

## Checklist
- [ ] (If possible) Add tests


kakkokari-gtyih:
> Thank you for sending us a great Pull Request! 👍 Please regenerate misskey-js type definitions! 🙏
> 
> example:
> 
> ```shell
> pnpm run build-misskey-js-with-types
> ```

I think you don't have to do regenerate for this

syuilo:
@misskey-dev/dev ちょっと今PC触れないから誰か pnpm run build-misskey-js-with-types してくれると助かる

tamaina:
wakatta

tamaina:
done

syuilo:
🙏

kakkokari-gtyih:
（多分developのmisskey-jsのバージョンを現行バージョンに合わせないと一生CI通らない）

PR #13146, opened 2024-02-02
ignore `instance.actor` when checking if there are local users

dakkar:
We've seen this happen a few times:

* there was some AP software at $some_domain
* it gets replaced by Misskey
* before the first user can be created, an AP activity comes in
* Misskey resolves the activity
* to do this, it creates the `instance.actor` to sign its request
* now there *is* a local user, so the `meta` endpoint returns `requireSetup:false`
* the admin is very confused

This commit factors out the check, and doesn't count the `instance.actor` as a real user.

(also, the new query used for the check is faster: it's `select 1 from user where … limit 1` instead of `select count(*) from user where …`)

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/core/InstanceActorService.ts
+++ b/packages/backend/src/core/InstanceActorService.ts
@@ -4,7 +4,7 @@
  */
 
 import { Inject, Injectable } from '@nestjs/common';
-import { IsNull } from 'typeorm';
+import { IsNull, Not } from 'typeorm';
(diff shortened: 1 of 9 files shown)


syuilo:
👍

Johann150:
probably resolves #8685

dakkar:
> probably resolves #8685

yep! that's exactly the problem I wrote this to fix

PR #13166, opened 2024-02-05
feat: Re-support for Elasticsearch

MomentQYC:
## What

Re-support for ElasticSearch Search
Earlier versions of ElasticSearch and OpenSearch are not currently supported due to [some official ElasticSearch behavior](https://github.com/elastic/elasticsearch-js/issues/1519#issuecomment-899296380).

## Why

MeiliSearch currently has issues with certain aspects (such as cumbersome upgrade steps). A solution could be provided by re-supporting ElasticSearch.

## Checklist
- [ ] (If possible) Add tests


u1-liquid:
I believe that for indexing purposes, it's more flexible to route indexing through Logstash or a similar intermediary rather than calling Elasticsearch directly from Misskey.

Due to Elasticsearch's architecture, concentrating all data into a single index is not advisable for both performance and management reasons.
It's preferable to split indices by month or day and use a wildcard (*) to query them collectively. 
Additionally, setting up an indices policy could automatically remove outdated data from remote servers.

Attempting to finely tune this setup within Misskey could become inefficient, akin to developing a template engine.
Instead, implementing a feature to transfer json_lines over TCP/UDP for use with Logstash / vector.dev would be more beneficial.

MomentQYC:
> I believe that for indexing purposes, it's more flexible to route indexing through Logstash or a similar intermediary rather than calling Elasticsearch directly from Misskey.
> 
> Due to Elasticsearch's architecture, concentrating all data into a single index is not advisable for both performance and management reasons. It's preferable to split indices by month or day and use a wildcard (*) to query them collectively. Additionally, setting up an indices policy could automatically remove outdated data from remote servers.
> 
> Attempting to finely tune this setup within Misskey could become inefficient, akin to developing a template engine. Instead, implementing a feature to transfer json_lines over TCP/UDP for use with Logstash / vector.dev would be more beneficial.

I have currently implemented index splitting by month and the ability to query them simultaneously using a wildcard, allowing for more efficient data retrieval.
I'm not really good at Elasticsearch-related development, and I don't have any friends around me who are particularly proficient in this area, so there's not much I can do at the moment.

MomentQYC:
I'm not quite sure what to do to fix the failed check anymore.

MomentQYC:
Temporarily closed

PR #13581, opened 2024-03-16
feat: check stream heartbeat after visibility changed to 'visible'

backrunner:
## What

- Add ability to send heartbeat to stream interface when visibility changed to 'visible' (frequency limited)

## Why

- Timer might be suspended or slowed down when the page turned to background, when it becomes visible again, the whole run might beyond 1 minute, which might cause some unexpected disconnections even if user just left the page with a short time period.

## Additional info (optional)

It's better to add an option to allow send heartbeat via websocket in background, for a small instance, it might bring a better experience to users.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/frontend/src/stream.ts
+++ b/packages/frontend/src/stream.ts
@@ -8,7 +8,12 @@ import { markRaw } from 'vue';
 import { $i } from '@/account.js';
 import { wsOrigin } from '@/config.js';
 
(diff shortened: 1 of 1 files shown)


syuilo:
🙏

syuilo:
なんかマージできない

syuilo:
できた

PR #13588, opened 2024-03-17
fix(backend): フォローリクエストを作成する際に既存のものは削除するように

kakkokari-gtyih:
## What

> In rare cases, a follow request can fail (due to DB error, Crash or API failure) and leave a dangling follow request.
> 
> Subsequent follows will throw a query error when it tries to add the follow request back to the database again and prevent the follower from following that followee forever.
> 
> This change makes it so that when a new follow request is created, the old one (if it exists) is removed first.

Cherry-picked from https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/440

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,8 @@
 
 ### Server
 - Enhance: エンドポイント`antennas/update`の必須項目を`antennaId`のみに
-- Fix: フォローリクエストを作成する際に既存のものは削除するように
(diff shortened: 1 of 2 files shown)


syuilo:
これは効かない感じかしら
https://github.com/misskey-dev/misskey/blob/b65203c9f852a29a3a6e7ce81c6761e9ac228bf3/packages/backend/src/models/FollowRequest.ts#L11

syuilo:
🙏

PR #13701, opened 2024-04-13
fix(backend): FileServerServiceでレンジリクエストの場合に適切なレスポンスコードが返らない問題を修正

kakkokari-gtyih:
## What

Cherry-picked from https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/490

> When configured to proxyRemoteFiles, Misskey would download the remote file for each request, then stream it out again. As part of this, it tries to correctly handle the Range header. But this was not really working. Here I fix two problems:
> 
> - two code paths were handling ranged requests but not setting the response code to 206, confusing the clients
> - remote files don't have their size stored in the database (correctly, because it can change!), but also we weren't detecting the size after downloading them, which disabled the range handling (again, sensible: you can't send a portion of a file if you don't know how bit the file is!)
> 
> Tested with firefox, by disabling all caching, and checking the network inspector to see what requests and responses were actually performed. Before, Sharkey always sent 200 and the whole file; after, it sends 206 and the section that the browser actually requested.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/server/FileServerService.ts
+++ b/packages/backend/src/server/FileServerService.ts
@@ -264,7 +264,6 @@ export class FileServerService {
 					const parts = range.replace(/bytes=/, '').split('-');
 					const start = parseInt(parts[0], 10);
 					let end = parts[1] ? parseInt(parts[1], 10) : file.file.size - 1;
-					console.log(end);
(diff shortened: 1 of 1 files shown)


syuilo:
🙏

PR #13742, opened 2024-04-24
feat: improve emoji endpoint

Hoto-Cocoa:
## What

I changed `GET /emoji/:path` endpoint logic.

## Why

It was have some problems:
* If there is at sign more than 2, Misskey tries the first chunk as the name and the second chunk as the server instead of throwing an error.
* As `path.split('@')[1]` is always a string, TypeScript assumes that `host` cannot be null/undefined, so `host == null` causes a type warning.

Those problems are not critical, but the second problem causes an incorrect type warning that is annoying.

I used the `shift()` method for the name and the `pop()` method for the host. Using `shift()` twice works the same, but I think shift/pop is a bit more intuitive.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/server/ServerService.ts
+++ b/packages/backend/src/server/ServerService.ts
@@ -120,12 +120,20 @@ export class ServerService implements OnApplicationShutdown {
 				return;
 			}
 
-			const name = path.split('@')[0].replace(/\.webp$/i, '');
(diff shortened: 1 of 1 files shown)


syuilo:
🙏🏻

PR #13875, opened 2024-05-24
[fix] json module imports for node 22

squidink7:
## What
nodejs 22 uses `with` rather than `assert` for importing json modules, which prevents misskey from building

## Why
Misskey should build without errors on the latest version of Node

## Checklist


syuilo:
Note: Misskey is currently targeting Node.js 20(LTS).

syuilo:
🙏🏻

PR #13966, opened 2024-06-11
fix(backend): Set comment to `NULL` instead of `Denormalized`

Mar0xy:
## What

The DROP migration listed setting the comment on the channelId column to the same as the UP migration which is invalid.

It should be `NULL` instead to clear out the comment

Initial concern about this was raised here https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/537#note_3712

## Why

(Look at the text above)

## Checklist

### DIFF
--- a/packages/backend/migration/1716129964060-ChannelIdDenormalizedForMiPoll.js
+++ b/packages/backend/migration/1716129964060-ChannelIdDenormalizedForMiPoll.js
@@ -15,7 +15,7 @@ export class ChannelIdDenormalizedForMiPoll1716129964060 {
 
     async down(queryRunner) {
         await queryRunner.query(`DROP INDEX "public"."IDX_c1240fcc9675946ea5d6c2860e"`);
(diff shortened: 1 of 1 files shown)


anatawa12 (commented):
https://github.com/typeorm/typeorm/blob/e7649d2746f907ff36b1efb600402dedd5f5a499/src/driver/postgres/PostgresQueryRunner.ts#L1138-L1151

It looks this is behavior of typeorm and there are meny similar down migrations so I think we don't have to fix this (if down migration works)

これはtypeormの実装起因っぽいですしすでにいろいろ似てるのあるし、(down migrationが動くなら)直さなくていい気がします

https://github.com/misskey-dev/misskey/blob/c51347d78bc6dd30b6b4db2af64f0ea4bc83091e/packages/backend/migration/1680931179228-account-move.js#L16-L21
https://github.com/misskey-dev/misskey/blob/c51347d78bc6dd30b6b4db2af64f0ea4bc83091e/packages/backend/migration/1607353487793-isExplorable.js#L15-L19

Mar0xy:
I would have done t his just to keep consistency but oh well

PR #14079, opened 2024-06-23
parse `notRespondingSince` from redis instance cache

dakkar:
## What
Correctly extract `notRespondingSince` from the Redis cache. It needs to be parsed.

## Why
If we don't do this, we'll get a string, and `DeliverProcessorService` will error out `i.notRespondingSince.getTime is not a function`

## Checklist
- [ ] (If possible) Add tests


syuilo:
🙏🏻

PR #14200, opened 2024-07-14
enhance(frontend): Allow negative delay in MFM

easrng:
## What

Allow negative times in MFM 

## Why

I made this change so MFM animations can be offset from each other and still start immediately (for example `$[bounce 1️⃣]$[bounce.delay=-0.5s 2️⃣]`)

## Checklist
- [ ] (If possible) Add tests


syuilo:
👍

PR #14269, opened 2024-07-19
feat: add the option to override cat ear colors

### LINKED ISSUE #14178: Add the ability to change cat ear colors
### Summary

After enabling "Mark this account as a cat", I would like three extra settings: "Manually set cat ear color", "Set outer cat ear color", and "Set inner cat ear colors". The last two would only show when "Manually set cat ear color" is enabled. This would allow setting both cat ear colors:

### Purpose

Personally, the algorithm used to generate the cat ear colors of my profile picture uses a color that I do not really like, and I don't want to change my profile picture. I also would like to set the inner cat ear color to change it to another shade of blue.

(I use a fork of Misskey called Sharkey, but I think it is better to request this to Misskey so that all Misskey forks support it)

I have also found another user wanting this: https://lethallava.land/notes/9jc6rl2gvqkq6v2e

- [X] Yes, I will implement this by myself and send a pull request

BlankEclair:
## What

This PR adds the ability to set custom cat ear colors. The colors are currently not federated ([per suggestion](https://github.com/misskey-dev/misskey/issues/14178#issuecomment-2239098070)).

## Why

I'd like to set the outer cat ear color because the algorithm of using the average color from the avatar blurhash is not adequete for profile pictures that are mostly transparent, like mine.

I'd also like to set the inner cat ear color because it is highly visible in my case, and I want it to match with the outer color.

## Additional info (optional)

`pnpm lint` failed in misskey-reversi even though I didn't touch it, so I couldn't lint the code.
The tests pass even when I try to intentionally break the e2e/users.ts test, so I can't test if the test added works properly.

The default outer and inner cat ear colors are hardcoded in a variety of places, but I'm not sure how to put them all in one place. I don't know if I should add them in `packages/misskey-js/src/consts.ts` as it doesn't seem like the place for it.

Japanese text is provided by a friend of mine who is learning Japanese.

Closes #14178

## Checklist
- [x] (If needed) Add story of storybook
- [x] (If possible) Add tests


BlankEclair:
No longer interested in this ^^;

PR #14314, opened 2024-07-27
fix memory leak and inconsistent behavior in memory caches

### LINKED ISSUE #14310: Memory caches have an infinite lifetime and grow to consume infinite RAM
### 💡 Summary

Several memory caches do not have an eviction policy and can grow to contain millions of entries per active process. This is highly inefficient and is a partial cause of the slow memory leak that affects some instances. In particular, these caches have an infinite eviction time:
* User objects (this cache is duplicated 4 additional times with varying keys)
* Public keys (this cache is duplicated with an alternate key)
* Client app metadata

As a solution, caches should define a maximum capacity and reasonable eviction time. If the capacity is exceeded, then the oldest (least recently used) entry should be evicted to make room. This will prevent caches from growing indefinitely while still ensuring a high cache hit rate.

---

This bug was first discovered in the Sharkey fork, but exists in Misskey as well. For additional details, see [Sharkey issue #600](https://activitypub.software/TransFem-org/Sharkey/-/issues/600) and [Sharkey MR #580](https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/580).

Related to #14311.

### 🥰 Expected Behavior

Caches should have a limited size and reasonable eviction method to prevent infinite memory usage.

### 🤬 Actual Behavior

Memory caches are unbounded and consume infinite server memory.

### 📝 Steps to Reproduce

Run a federated instance for an extended period of time and observe memory usage. Background memory will increase indefinitely as caches are filled with remote users, public keys, and other data.

### 💻 Frontend Environment

```markdown
N/A - this is a backend issue. The frontend environment does not matter.
```

```markdown
N/A - this effects all instances running a Misskey version newer than March 2021. The backend environment is irrelevant.
```

- [X] Yes, I will patch the bug myself and send a pull request

### LINKED ISSUE #14311: Memory cache eviction is inefficient and error-prone
### 💡 Summary

Cache eviction from `MemoryKVCache<T>` is currently based on a hard-coded 3 minute timer. When this timer ticks, the `gc()` method loops through all entries and removes any that have been evicted. This has a few issues:
1. The 3-minute interval adds considerable jitter to eviction times. Some caches have lifetimes as short as 10 minutes, which translates to a **30% margin of error**.
2. The full loop wastes CPU resources on larger caches, like user & public keys.
3. If the cache is not properly disposed, then it creates a memory leak and timer leak.

This can be solved by modifying the implementation to create a new timer per entry. When one of these timers ticks, it removes the associated row without any loops or checks. The timer ID can be stored within the cache entry and automatically cancelled when the entry is deleted. This has three benefits:
1. Eviction timing is much more accurate.
2. The full loop is removed.
3. If the cache is not disposed, then all entries will slowly evict until the cache is empty. Then it will be garbage collected and not produce a leak.

---

This bug was first discovered in the Sharkey fork, but exists in Misskey as well. For additional details, see [Sharkey issue #601](https://activitypub.software/TransFem-org/Sharkey/-/issues/601) and [Sharkey MR #580](https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/580).

Related to #14310.

### 🥰 Expected Behavior

Cache eviction should be predictable and not consume significant system resources.

### 🤬 Actual Behavior

Cache eviction is slow and inconsistent.

### 📝 Steps to Reproduce

Run a federated instance for an extended period of time and wait for caches to fill. The "garbage collection" process will consume a notable amount of CPU time every 3 minutes or so.

### 💻 Frontend Environment

```markdown
N/A - this is a backend issue. The frontend environment does not matter.
```

```markdown
N/A - this effects all instances running a Misskey version newer than March 2021. The backend environment is irrelevant.
```

- [X] Yes, I will patch the bug myself and send a pull request

warriordog:
## What

This PR resolves #14310 and #14311 by refactoring `MemoryKVCache<T>`. The new implementation supports capacity limits with Least Recently Used (LRU) eviction, and no longer requires garbage collection or manual disposal. All references to this class have been updated to specify a reasonable capacity limit.

Additionally, all cache lifetimes and capacity limits have been exposed as configuration values. This allows the limits to be adjusted by instance admins to account for available system resources, which reduces the chance of cache-related resource exhaustion. Larger instances can even *increase* the limits, thus improving cache hit rate and reducing churn.

## Why

The linked issues pose a serious reliability issue that is affecting many instances, including my own. Resolving the memory leaks will improve stability, uptime, and performance.

## Additional info (optional)

This change is upstreamed from the Sharkey fork. Please see [Sharkey MR #580](https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/580) for additional details and context.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/.config/docker_example.yml
+++ b/.config/docker_example.yml
@@ -175,6 +175,130 @@ id: 'aidx'
 # deliverJobMaxAttempts: 12
 # inboxJobMaxAttempts: 8
 
+caches:
+  # Lifetime of cached ID->user lookups in memory. Default: 43200000 (12 hours)
+  userByIdMemoryLifetime: 43200000
+  # Maximum number of ID->user lookups cached in memory. Default: 15000
+  userByIdMemoryCapacity: 15000
+
+  # Lifetime of cached ID->local user lookups in memory. Default: 43200000 (12 hours)
+  localUserByIdMemoryLifetime: 43200000
+  # Maximum number of ID->local user lookups cached in memory. Default: 10000
+  localUserByIdMemoryCapacity: 10000
+
+  # Lifetime of cached token->user lookups in memory. Default: 43200000 (12 hours)
+  userByTokenMemoryLifetime: 43200000
+  # Maximum number of token->user lookups cached in memory. Default: 10000
+  userByTokenMemoryCapacity: 10000
+
+  # Lifetime of cached URI->user lookups in memory. Default: 43200000 (12 hours)
+  userByUriMemoryLifetime: 43200000
+  # Maximum number of URI->user lookups cached in memory. Default: 10000
+  userByUriMemoryCapacity: 10000
+
+  # Lifetime of cached user profiles in redis. Default: 1800000 (30 minutes)
+  userProfileRedisLifetime: 1800000
+  # Lifetime of cached user profiles in memory. Default: 60000 (1 minute)
+  userProfileMemoryLifetime: 60000
+  # Maximum number of user profiles cached in memory. Default: 10000
+  userProfileMemoryCapacity: 10000
+
+  # Lifetime of cached user keypairs in redis. Default: 86400000 (1 day)
+  userKeyPairRedisLifetime: 86400000
+  # Lifetime of cached user keypairs in memory. Default: 43200000 (12 hours)
+  userKeyPairMemoryLifetime: 43200000
+  # Maximum number of user keypairs cached in memory. Default: 10000
+  userKeyPairMemoryCapacity: 10000
+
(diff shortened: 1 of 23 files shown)


syuilo:
Thanks for the PR!
Since it seems redundant to make everything configurable about the cache used inside the system, and since this is likely to change as development progresses, it would be better to make it possible to set some single factor that applies to the entire system, such as the “rate limit” setting in the role policy.

warriordog:
This has been replaced by #14363

PR #14363, opened 2024-08-04
fix(backend): memory leak in memory caches

### LINKED ISSUE #14310: Memory caches have an infinite lifetime and grow to consume infinite RAM

### LINKED ISSUE #14311: Memory cache eviction is inefficient and error-prone

warriordog:
## What

This PR resolves #14310 and #14311 by replacing all "infinite" cache lifetimes and cleaning up `MemoryKVCache<T>`. Expiration lifetimes of all memory / redis caches have been reset to a reasonable number and normalized between similar caches.
Redis caches have a synchronization fix, and an optimization is applied to `MemoryKVCache.gc()`.

## Why

The linked issues pose a serious reliability issue that is affecting many instances, including my own. Resolving the memory leaks will improve stability, uptime, and performance.

Fix #14310
Fix #14311

## Additional info (optional)

This change is upstreamed from the Sharkey fork. Please see [Sharkey MR #580](https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/580) for additional details and context.

This replaces #14314, which was an earlier attempt at a similar fix.

## Checklist
- [ ] (If possible) Add tests


packages/backend/src/core/RelayService.ts
```diff
@@ -35,7 +35,7 @@ export class RelayService {
 		private createSystemUserService: CreateSystemUserService,
 		private apRendererService: ApRendererService,
 	) {
-		this.relaysCache = new MemorySingleCache<MiRelay[]>(1000 * 60 * 10);
+		this.relaysCache = new MemorySingleCache<MiRelay[]>(1000 * 60 * 10); // 10s
```
kakkokari-gtyih:
```suggestion
		this.relaysCache = new MemorySingleCache<MiRelay[]>(1000 * 60 * 10); // 10m
```

packages/backend/src/core/RoleService.ts
```diff
@@ -127,10 +127,8 @@ export class RoleService implements OnApplicationShutdown, OnModuleInit {
 		private moderationLogService: ModerationLogService,
 		private fanoutTimelineService: FanoutTimelineService,
 	) {
-		//this.onMessage = this.onMessage.bind(this);
-
-		this.rolesCache = new MemorySingleCache<MiRole[]>(1000 * 60 * 60 * 1);
-		this.roleAssignmentByUserIdCache = new MemoryKVCache<MiRoleAssignment[]>(1000 * 60 * 60 * 1);
+		this.rolesCache = new MemorySingleCache<MiRole[]>(1000 * 60 * 60); // 1h
+		this.roleAssignmentByUserIdCache = new MemoryKVCache<MiRoleAssignment[]>(1000 * 60 * 5); // 1h
```
kakkokari-gtyih:
```suggestion
		this.roleAssignmentByUserIdCache = new MemoryKVCache<MiRoleAssignment[]>(1000 * 60 * 5); // 5m
```

packages/backend/src/queue/processors/DeliverProcessorService.ts
```diff
@@ -45,7 +45,7 @@ export class DeliverProcessorService {
 		private queueLoggerService: QueueLoggerService,
 	) {
 		this.logger = this.queueLoggerService.logger.createSubLogger('deliver');
-		this.suspendedHostsCache = new MemorySingleCache<MiInstance[]>(1000 * 60 * 60);
+		this.suspendedHostsCache = new MemorySingleCache<MiInstance[]>(1000 * 60 * 60); // 1m
```
kakkokari-gtyih:
```suggestion
		this.suspendedHostsCache = new MemorySingleCache<MiInstance[]>(1000 * 60 * 60); // 1h
```

packages/backend/src/server/NodeinfoServerService.ts
```diff
@@ -134,7 +134,7 @@ export class NodeinfoServerService {
 			return document;
 		};
 
-		const cache = new MemorySingleCache<Awaited<ReturnType<typeof nodeinfo2>>>(1000 * 60 * 10);
+		const cache = new MemorySingleCache<Awaited<ReturnType<typeof nodeinfo2>>>(1000 * 60 * 10); // 10s
```
kakkokari-gtyih:
```suggestion
		const cache = new MemorySingleCache<Awaited<ReturnType<typeof nodeinfo2>>>(1000 * 60 * 10); // 10m
```

u1-liquid:
See also: https://github.com/MisskeyIO/misskey/pull/174 https://github.com/MisskeyIO/misskey/pull/289

packages/backend/src/misc/cache.ts
```diff
@@ -77,14 +83,14 @@ export class RedisKVCache<T> {
 
 		// Cache MISS
 		const value = await this.fetcher(key);
-		this.set(key, value);
+		await this.set(key, value);
```
syuilo, line 90: awaitする必要あるかしら？

kakkokari-gtyih: > we don't technically need to, but fire-and-forget async stuff is always risky. What if the write fails? This worker would have a value in its memory cache, but the others would then re-calculate a new value.

> And without this, it's impossible for any code to require a consistent cache state. There's some ongoing work to synchronize cache logic through IPC, and some of it is built to notify with an ID and then re-fetch from redis. If this returns early, then it's conceivable that the notification will reach another worker before redis is updated. That could result in inconsistent state.

> Also, there won't be concurrent calls to `set` but there could be concurrent writes to Redis if this method is called twice in a row.

https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/587#note_4597

syuilo: 理由があるならコメントで残しておいた方が良いかも

warriordog: @syuilo sorry, I don't quite understand. Are you asking me to document the reason with code comments?

kakkokari-gtyih: > @syuilo sorry, I don't quite understand. Are you asking me to document the reason with code comments?

right

warriordog: thanks for clarifying! I'll add comments.

warriordog: I added comments to the `fetch` methods. Please let me know whether these make sense.

syuilo:
🙏🏻

PR #14716, opened 2024-10-06
feat: Initial support for TTS functionality

MomentQYC:
## What

This is an initial implementation of the TTS functionality, which can subsequently be built upon to improve ease of use, add providers (the existing ones can actually be deployed on our own), and even be set up by the user.

## Why

Hopefully it can be an emergency measure for some of the population at some point.

## Additional info (optional)

It is currently only a preliminary realization, so only a rough framework has been formed.

## Checklist
- [ ] (If possible) Add tests

Thanks to @MarryDream  for the guidance in implementing the feature.


MomentQYC:
> I think we should discuss the feature first (especially large scale one like this) in an issue before actually implement it

#14721

PR #14825, opened 2024-10-23
fix(backend): Accept arrays in ActivityPub `icon` and `image` properties

### LINKED ISSUE #14823: Remote actors' `icon` and/or `image` are rejected if they are arrays
### 💡 Summary

When looking at the profile of a Bluesky actor bridged through Bridgy Fed, the banner image is not shown.

The reason is that a remote actor's `icon` or `image` is rejected by Misskey if it's an array rather than a single `Image` object.
(Arrays are permitted according to the Activity vocabulary: https://www.w3.org/TR/activitystreams-vocabulary/#dfn-icon)

This was reported to us as https://github.com/snarfed/bridgy-fed/issues/1408.

### 🥰 Expected Behavior

Misskey should use the first valid entry in the array in this case.

### 🤬 Actual Behavior

Remote actors' `icon` and/or `image` are rejected if they are arrays.

### 📝 Steps to Reproduce

1. Look at the profile `@snarfed.org@bsky.brid.gy` through Misskey, for example at https://misskey.io/@snarfed.org@bsky.brid.gy.
2. The banner image is missing.

(Note that this may start working due to a workaround in Bridgy Fed, but it would still be nice to fix the underlying bug.)

### 💻 Frontend Environment

* Model and OS of the device(s): PC, Windows 11 Pro 23H2 22631.4317
* Browser: Floorp 11.19.1 (64-Bit)
* Server URL: misskey.io
* Misskey: 2024.5.0-io.3d

* Misskey:
* Node:
* PostgreSQL:
* Redis:
* OS and Architecture:

Tamschi:
## What

This fixes #14823 by using the first avatar `icon` or banner `image` object of a remote ActivityPub Actor if an array of objects is provided.

Note that this supports only [Image](https://www.w3.org/TR/activitystreams-vocabulary/#dfn-image)s with `url` property.
Link objects with `href` rather than `url`, which are also valid in this position, as well as plain strings, are skipped when encountered as array items.
(Misskey also accepts a plain string when provided directly as `icon` or `image`, if I'm not mistaken.)

## Why

This fixes a partial incompatibility with Bridgy Fed that was reported to us at https://github.com/snarfed/bridgy-fed/issues/1408.

## Additional info (optional)

I tried to match the existing code style as well as possible, but let me know if anything is amiss.

Apologies for not authoring tests or updating CHANGELOG.md.

I'm aware that testing is the larger part of the work here, but I don't have the spare time to learn enough about Misskey's repository structure on my own to do so. I'm also not sure where to place the changelog entry.

I wanted to provide at least the fix itself here, since I already found the root issue.
With further instruction, I should be able to add the missing tests eventually, but that would take some time since I'm relatively busy.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/core/activitypub/models/ApPersonService.ts
+++ b/packages/backend/src/core/activitypub/models/ApPersonService.ts
@@ -232,6 +232,12 @@ export class ApPersonService implements OnModuleInit {
 		if (user == null) throw new Error('failed to create user: user is null');
 
 		const [avatar, banner] = await Promise.all([icon, image].map(img => {
+			// icon and image may be arrays
+			// see https://www.w3.org/TR/activitystreams-vocabulary/#dfn-icon
+			if (Array.isArray(img)) {
+				img = img.find(item => item && item.url) ?? null;
+			}
+			
 			// if we have an explicitly missing image, return an
 			// explicitly-null set of values
 			if ((img == null) || (typeof img === 'object' && img.url == null)) {


syuilo:
🙏🏻

PR #14826, opened 2024-10-23
fix: encode RSS URIs with escape sequences before fetching

### LINKED ISSUE #9517: Status bar does not use the parameters in an RSS link
## 💡 Summary

According to the Hacker News RSS [manual](https://hnrss.github.io/#search-feeds), subscribe to this link 

`https://hnrss.org/show?points=100&link=comments`

Clicking the items should bring me directly to the Hacker News website instead of the submitted article’s URL

## 🥰 Expected Behavior

RSS  <link> should be queried with the parameters in the URL.

## 🤬 Actual Behavior

It ignores the parameters `&parameter` after the `?first` parameter.

## 📝 Steps to Reproduce

1. Open the settings panel
2. Open status bar
3. Use the link `https://hnrss.org/show?points=100&link=comments`
4. Click save
5. Click any item on the status bar
6. It jumps to the article instead of HN comments page. 

## 📌 Environment

Misskey version: [13.0.0-beta.37](https://github.com/misskey-dev/misskey/releases/tag/13.0.0-beta.37)
Your OS: `Ubuntu 22.04.1 LTS`
Your browser: `Firefox-1.8.01`

pinapelz:
## What

Encodes all uris with built-in browser function before calling backend so that query parameters are handled correctly. The issue has some details but to summarize the problem:

Example: `https://hnrss.org/show?points=100&link=comments`

Passed into API: `https://misskey.io/api/fetch-rss?url=https://hnrss.org/show?points=100&link=comments`

The RSS feed gets seen as `https://hnrss.org/show?points=100` by the backend because it will treat the `&link=comments` as apart of  `https://misskey.io/api/fetch-rss` and not the RSS feed link. (Since the query parameter of the backend comes first)

Solution implemented in this PR is to call `encodeURIComponent` making all special characters into an escape sequence which eliminates the issue:

`https://misskey.io/api/fetch-rss?url=https%3A%2F%2Fhnrss.org%2Fshow%3Fpoints%3D100%26link%3Dcomments`

## Why

Closes #9517

## Additional info (optional)

- I already considered encoding the uri in [fetch-rss.ts](https://github.com/misskey-dev/misskey/blob/develop/packages/backend/src/server/api/endpoints/fetch-rss.ts) and then fetching, but this isn't possible because the frontend makes calls to this endpoint by calling `/api/fetch-rss?url=` which will already break any query parameters before it reaches `fetch-rss.ts`.

- This can be solved if the user encodes the URI first before entering it into the RSS widgets/status-bar but it is not obvious to do this and I believe that it should be handled by Misskey.

## Checklist
- [ ] (If possible) Add tests


syuilo:
🙏🏻

PR #14844, opened 2024-10-26
fix: Unable to crop avatar in some cases

### LINKED ISSUE #14843: Unable to crop avatar in some cases
### 💡 Summary

Unable to crop avatar when accessing a domain name that does not match the instance domain name

### 🥰 Expected Behavior

see above

### 🤬 Actual Behavior

see above

### 📝 Steps to Reproduce

_No response_

### 💻 Frontend Environment

* Misskey: 2024.10.1

Not related

MomentQYC:
## What

close #14843

## Why

#14843

## Checklist
- [ ] (If possible) Add tests


MomentQYC:
> The third argument is a flag to force the same origin, so it should be working as-is.

In my use case, it is displaying the instance domain name instead of the current domain name, so the feature doesn't work.

MomentQYC:
As well as and this: https://developer.mozilla.org/en-US/docs/Web/HTML/CORS_enabled_image#security_and_tainted_canvases

MomentQYC:
> Are you running the frontend on a separate domain from the backend? That's not the vanila Misskey behavior and is not guaranteed to work.

My use case is to reverse proxy Misskey on a different domain than the one set up

MomentQYC:
> Well, afaik that is not guaranteed to work

In fact, this type of use case is working basically well at the moment, and this PR is an attempt to make it work even better.

MomentQYC:
> Well, afaik that is not guaranteed to work

Any other issues preventing it from being merged please?

PR #14920, opened 2024-11-08
fix(backend): SQLのサニタイズを強化

kakkokari-gtyih:
## What

> To fix the problem, we need to modify the sqlLikeEscape function to escape backslashes in addition to the existing meta-characters (`%` and `_`). This can be achieved by using a regular expression that matches backslashes and the existing meta-characters, and replacing them with their escaped versions. Specifically, we should replace each backslash with a double backslash and each meta-character with a backslash followed by the character.

For more information, visit https://github.com/MisskeyIO/misskey/pull/800

## Additional info (optional)

素人目に要りそうと思って持ってきただけなので要らなさそうなら閉じてください

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/misc/sql-like-escape.ts
+++ b/packages/backend/src/misc/sql-like-escape.ts
@@ -4,5 +4,5 @@
  */
 
 export function sqlLikeEscape(s: string) {
-	return s.replace(/([%_])/g, '\\$1');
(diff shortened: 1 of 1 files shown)


syuilo:
🙏

PR #15022, opened 2024-11-22
fix(backend): Inboxのエラーをthrowせずreturnしている問題を修正

kakkokari-gtyih:
## What

> Fixes a type confusion bug where exceptions are incorrectly returned instead of thrown. The calling code will attempt to process the Error as an Activity, causing it to fail with cryptic and confusing errors.

## Why

Fix https://activitypub.software/TransFem-org/Sharkey/-/issues/796
エラーがエラーとしてログに残るようになる

## Checklist
- [ ] (If possible) Add tests


syuilo:
🙏

PR #15190, opened 2024-12-29
feat(misskey-js): allow setting `binaryType` of WebSocket connection

cannorin:
## What

This PR allows developers who use `misskey-js` to customize [`binaryType`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/binaryType) of the underlying WebSocket connection when using the streaming API.

## Why

I wanted to use this library on [bun](https://bun.sh/) to create a bot application with streaming capability.

However, bun does not support WebSocket connections with `binaryType` set to `blob` (probably just yet), which happens to be the default value.

See: https://github.com/oven-sh/bun/issues/8721

## Additional info (optional)

I'm not very sure if testing this over a mocked WebSocket server (`new WS('wss://misskey.test/streaming')`) would be helpful, as what really matters here is whether the real Misskey servers accept WebSocket connections of `arraybuffer` binary type.

As far as I can test, it is working without any issues when connecting to my own Misskey instance.

## Checklist
- [ ] (If possible) Add tests


CHANGELOG.md
```diff
@@ -30,6 +30,9 @@
 - Fix: ユーザーのプロフィール画面をアドレス入力などで直接表示した際に概要タブの描画に失敗する問題の修正( #15032 )
 - Fix: 起動前の疎通チェックが機能しなくなっていた問題を修正  
   (Cherry-picked from https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/737)
+
+### Misskey.js
+- Feat: allow setting `binaryType` of WebSocket connection
```
kakkokari-gtyih: The Changelog is in the wrong place, `Misskey.js` section should be added after the `Server` section

cannorin: Fixed!

kakkokari-gtyih (approved):

PR #15210, opened 2025-01-04
fix(backend): リプライがstreamで配信される際にリプライ元のmyReactionが正常に配信されていない問題を修正

kakkokari-gtyih:
## What

リプライとその被リプライのノートがstreamに流れる際に、被リプライのノートに`myReaction`がある（＝被リプライのノートに自分がリアクションしている）場合でもそれが正常に反映されていない問題を修正

image

↑青色の枠で囲ったノートがstreamで流れてきた際に、赤色の枠で囲ったほうのリアクションに`myReaction`があった場合でも正常に記録されていないという問題があった

## Additional info (optional)

Misskeyの標準UIでは被リプライのノートのリアクションは最小化されて表示されないので今まで影響がなかった

## Checklist
- [ ] (If possible) Add tests


dakkar:
please do *NOT* merge this *yet*

we have discovered a bug.

the source of the problem is that `StreamingApiServerService` creates a single `EventEmitter` per server process https://github.com/misskey-dev/misskey/blob/develop/packages/backend/src/server/api/StreamingApiServerService.ts#L115

so when a note is published to redis, it gets de-serialised once per server process, and then *that single objecc* is passed to all active channels on each connection.

so for example, if we had a single web server process:

* user A is looking at a timeline
* user B is looking at a timeline, doesn't matter if it's the same kind or not
* a note arrives, that is visible to both users
* the `onNote` of the channel attached to the websocket connection of user A sets the `myReaction` inside the note, asynchronously
* the `onNote` of user B does the same, *assigning a different value inside the same object*
* these two operations can happen in any order, before each `onNote` does the `send` at the end
* so the frontend can receive a note with the wrong `myReaction`

this was not a problem before these changes, because each `onNote` did a single `await` before the assignment, then did the `send` immediately; now, with more promises/await involved, things break

we're working on a fix (which will probably involve a doing shallow clone of the note object before modifying it), I'll link to our branch as soon as we've tested it

dakkar:
https://activitypub.software/TransFem-org/Sharkey/-/merge_requests/854 has been merged and should fix the data sharing bug, please cherry-pick those commits as well

packages/backend/src/server/api/stream/channel.ts
```diff
@@ -101,6 +102,44 @@ export default abstract class Channel {
...
 	public onMessage?(type: string, body: JsonValue): void;
+
+	public async assignMyReaction(note: Packed<'Note'>, noteEntityService: NoteEntityService): Promise<Packed<'Note'>> {
+		let changed = false;
+		// StreamingApiServerService creates a single EventEmitter per server process,
+		// so a new note arriving from redis gets de-serialised once per server process,
+		// and then that single object is passed to all active channels on each connection.
+		// If we didn't clone the notes here, different connections would asynchronously write
+		// different values to the same object, resulting in a random value being sent to each frontend. -- Dakkar
+		const clonedNote = { ...note };
```
kakkokari-gtyih, line 113: deepCloneを使うようにするのでも良いかも

dakkar: we didn't want to use that because it would duplicate a lot of data that is not going to be mutated… that's why notes only get shallow-cloned just before we change the myReaction attribute

packages/backend/src/server/api/stream/channel.ts
```diff
@@ -101,6 +102,44 @@ export default abstract class Channel {
...
+					changed = true;
+					clonedNote.renote = { ...note.renote };
+					clonedNote.renote.myReaction = myReaction;
+				}
+			}
+			if (note.renote?.reply && Object.keys(note.renote.reply.reactions).length > 0) {
+				const myReaction = await noteEntityService.populateMyReaction(note.renote.reply, this.user.id);
+				if (myReaction) {
+					changed = true;
+					clonedNote.renote = { ...note.renote };
```
samunohito, line 127: ①115行目の条件と123行目の条件は排他ではなく両方ともtrueになり得る
②115行目の条件がtrueの時、複製したnoteの`clonedNote.renote.myReaction`を書き換えているが、127行目で再度上書きしている

机上で追跡しただけですが、上記のような動きをしそうな気がします…？

dakkar: in one case we're assigning to `*.renote.myReaction`, in the other we're assigning to `*.renote.reply.myReaction`

PR #15535, opened 2025-02-22
fix(backend): reject symlinks on emoji import

eternal-flame-AD:
## What

Check the emoji files in import are actually regular files.

## Why

It is possible someone can upload a symlink and in turn read arbitrary files on the system. 

This is currently not exploitable because the `zip` crate version in `slacc` is too low and extracts to a `0o777` file instead of an actual symlink, but after bumping to the latest verion it will become vulnerable.

## Additional info (optional)

Maybe consider add checks in `slacc` too, might be better to check every file before extracting instead of just call `.extract()`.

## Checklist
- [ ] (If possible) Add tests


anatawa12:
I feel it's better to check if each entry is (expected to be) symlink before writing to file system in slacc than checking on javascript side

eternal-flame-AD:
I agree, I can write a PR for slacc if you accept PR there.

It isn't a security problem in the crate IMO because it is not specifically designed to handle user uploaded files. 

The behavior is [documented](https://docs.rs/zip/latest/zip/read/struct.ZipArchive.html#method.extract) and I agree it is expected behavior that when one asks for "extract this zipfile" by default it creates symlinks, and there are legitimate uses like taking a backup of a directory which can include symlinks that traverse out of the directory.

eternal-flame-AD:
I found a related but different sec concern in zip2 and I contacted the author to confirm, it would be best if we hold off updating for now.

But generally speaking I don't think we should extract zip files from any user upload, just read the content of `record.fileName` into a JS buffer and return. Zip files can encode a lot more than just file contents and the default behavior for extracting files can have many unexpected behaviors.

syuilo:
🙏🏻

Issue #14684, opened 2024-10-03
Require signed GET requests (authorized fetch) when using federation allow list


belatedly:
### Summary

I wasn't sure whether to call this a bug or not. I tested the allow/white list function in the current alpha and instances that are not entered in the allowed box can still fetch users and posts by searching the URL. Instances that are not explicitly allowed should not be able to fetch anything from the instance.

I did not receive a follow request from the instance I tested with, so it seems that this functionality is currently only rejecting certain activities from instances that are not explicitly allowed.

### Purpose

Using an allow list allows an admin to control access to their users and content, in its current state the functionality is not enforcing this. Authorized fetch would check for signatures and reject all GET requests from instances that are not explicitly allowed.

### Do you want to implement this feature yourself?

- [ ] Yes, I will implement this by myself and send a pull request

sneexy-boi:
From what I know, this is (although "experimental") added in [Sharkey](https://activitypub.software/TransFem-org/Sharkey/). That could probably be upstreamed to here.

KisaragiEffective:
> From what I know, this is (although "experimental") added in [Sharkey](https://activitypub.software/TransFem-org/Sharkey/). That could probably be upstreamed to here.

Could you point to the MR? I couldn't find one...

belatedly:
> Could you point to the MR? I couldn't find one...

Here's the original merge: https://activitypub.software/TransFem-org/Sharkey/-/commit/fd57c7e24caf936392440de3c063b5eaa9d588cb 

These are enhancements/fixes: https://activitypub.software/TransFem-org/Sharkey/-/commit/4f2fa60a72d1613ddaa315570d5754f7b54a2ac3 and https://activitypub.software/TransFem-org/Sharkey/-/commit/94fee180b92af2ff8751173d939297d584ea5244

temtemy:
There are cases where allowlist without authorized fetch still make sense. For example, an allowlist can be used as a last-ditch defense against spam, if all others (like filtering with "prohibited words") fail. But in that case the admin still doesn't want to completely shut itself off from other instances not part of the allowlist.

In fact it would be nice if the allowlist's behavior can be adjusted, like whether it should *only* block all incoming messages or *also* suspend all activities sent to servers outside the list.

temtemy:
Oh and maybe an allowlist where the Misskey will only deliver activities could be useful too... Being able to only deliver activities to the relays you've subscribed to would lessen a lot of network pushing I'd imagine.

PR #15659, opened 2025-03-13
[WIP] enhance(backend): hoist HTTP signature and note lockdown checks for existing users to ActivityPubServerService

### LINKED ISSUE #15644: ロックダウンされた/ログイン必須ノートは /outbox で見える
### 💡 Summary

タイトル通り

(I received permission to post this in public.)

### 🥰 Expected Behavior

Image

From the description one would reasonably conclude:

- This note is not visible from the _home_ instance
- This note is no longer federated in public view (a non follower should not be able to look this note up regardless of their respect of the _misskey vendor keys).
- (Ideally but not mandatory behavior) Further AP get requests need a valid signature.

### 🤬 Actual Behavior

https://mi.yumechi.jp/users/a10sac8leyy40014/outbox?page=true

The note can still be looked up, have "public" audience and can be enumerated using `/outbox?page=true`

Image

### 📝 Steps to Reproduce

Set up lockdown

Go to the `/users/<id>/outbox`

### 💻 Frontend Environment

```markdown
Not frontend problem
```

```markdown
* Installation Method or Hosting Service: podman
* Misskey: 2025.3.1
* Node: 22 
* PostgreSQL: 17
* Redis: official
* OS and Architecture: linux x86_64
```

- [x] Yes, I will patch the bug myself and send a pull request (Probably after Thurs or Friday, don't wait for me if time sensitive)

eternal-flame-AD:
## What

タイトル通りです。

Some minor sideline changes:

Removed 'Vary' header for /note/:id/activity, I think it always return AP?

## Why

- Fixes #15644 
- Makes #14684 much easier to implement
- Hoisted HTTP signature validation upwards to HTTP handler (see addn. info) to enable filtering notes by lockdown settings.

## Additional info (optional)

RSA signature validation is pretty fast in the grand scheme of RSA (in the order of 20-50kops/s IIRC), I think we can hoist it up to provide more opportunities for filtering with less impact on legitimate federation.

## Checklist
- [ ] TODO: Test working in a local environment
- [ ] TODO: Update CHANGELOG.md
- [ ] TODO: Add tests

### DIFF
--- a/packages/backend/src/core/IdService.ts
+++ b/packages/backend/src/core/IdService.ts
@@ -7,11 +7,11 @@ import { Inject, Injectable } from '@nestjs/common';
 import { ulid } from 'ulid';
 import { DI } from '@/di-symbols.js';
 import type { Config } from '@/config.js';
-import { genAid, isSafeAidT, parseAid } from '@/misc/id/aid.js';
-import { genAidx, isSafeAidxT, parseAidx } from '@/misc/id/aidx.js';
-import { genMeid, isSafeMeidT, parseMeid } from '@/misc/id/meid.js';
-import { genMeidg, isSafeMeidgT, parseMeidg } from '@/misc/id/meidg.js';
-import { genObjectId, isSafeObjectIdT, parseObjectId } from '@/misc/id/object-id.js';
+import { floorAid, ceilAid, genAid, isSafeAidT, parseAid } from '@/misc/id/aid.js';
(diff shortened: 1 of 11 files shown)


eternal-flame-AD:
I have not looked at implementation correctness at all in this stage, just to gauge if there are suggestions or objections to this approach of validating HTTP signature for existing users inline.

Issue #15853, opened 2025-04-16
Disable email notification


ghost:
### Summary

Possibility to disable email notification after every login.

### Purpose

Add option in Settings->Email->Disable login alerts

### Do you want to implement this feature yourself?

- [X] Yes, I will implement this by myself and send a pull request

PR #15871, opened 2025-04-19
Disable login email

ghost:
## What

I added the ability to disable login notifications by email. [#15853](https://github.com/misskey-dev/misskey/issues/15853)

## Why

Constant notifications are annoying and clutter up inbox, so I think users should have the freedom to turn them off.

## Checklist
- [ ] (If possible) Add tests


kakkokari-gtyih:
It is not a good security practice to completely disable login notification. A better compromise should be considered, such as only notifying users when they log in from an IP address from which they have already logged in.

miraikumiko:
> It is not a good security practice to completely disable login notification. A better compromise should be considered, such as only notifying users when they log in from an IP address from which they haven't logged in before.

It could be enabled by default and allow the user to choose to disable it. I don't think what makes decisions for the user it's good idea. For example, over the course of a week I have accumulated more than 10 emails about login, and I would like to simply disable them. I suggest making emails enabled by default and the possibility to disable it. What do you say?

Sayamame-beans:
> For example, over the course of a week I have accumulated more than 10 emails about login,

(I'm personally curious, why do new logins occur so often for you?)

miraikumiko:
> > For example, over the course of a week I have accumulated more than 10 emails about login,
> 
> (I'm personally curious, why do new logins occur so often for you?)

@Sayamame-beans for various reasons, such as testing, but I don't think it's important because it's impossible to present all possible reasons for users. Someone can log in through a private browser tab, as strange as it may sound, for reasons of anonymity, for example.

miraikumiko:
> > It is not a good security practice to completely disable login notification. A better compromise should be considered, such as only notifying users when they log in from an IP address from which they haven't logged in before.
> 
> @kakkokari-gtyih It could be enabled by default and allow the user to choose to disable it. I don't think what makes decisions for the user it's good idea. For example, over the course of a week I have accumulated more than 10 emails about login, and I would like to simply disable them. I suggest making emails enabled by default and the possibility to disable it. What do you say?

@kakkokari-gtyih I added "login" to the default values ​​of emailNotificationTypes in packages/backend/migration/1745174115062-emailNotificationTypes.js, but the migration is not applied because this column exists. I don't really know how to safely change the default values in PostgreSQL.

Mar0xy:
> A better compromise should be considered, such as only notifying users when they log in from an IP address from which they haven't logged in before.

I agree on this instead of forcing a disable just do it like most sites where they only notify if a new IP is detected on login.

> I added "login" to the default values ​​of emailNotificationTypes in packages/backend/migration/1745174115062-emailNotificationTypes.js, but the migration is not applied because this column exists. I don't really know how to safely change the default values in PostgreSQL.

Do not change existing Migration files you need to make new ones for when you edit columns or add them.
https://github.com/misskey-dev/misskey/blob/develop/CONTRIBUTING.md#migration%E4%BD%9C%E6%88%90%E6%96%B9%E6%B3%95

Issue #15852, opened 2025-04-16
Disable reconnection notification


ghost:
### Summary

Possibility to disable reconnection alert "Connection to has been lost".

### Purpose

Add option in Settings->Preferences->Other->When losing connection to the server->**Disable**

### Do you want to implement this feature yourself?

- [X] Yes, I will implement this by myself and send a pull request

miraikumiko:
@kakkokari-gtyih Hi! You marked this as a duplicate, but such functionality has not been implemented yet.
I just made [this option](https://github.com/miraikumiko/misskey/tree/disable-reconnect-alert) and would like to ask if I can make a Pull Request?

PR #15872, opened 2025-04-19
Disable reconnect alert

### LINKED ISSUE #12210: Stream indicator is bugging people
### Summary

Stream indicator is a huge modal window, barely useful when screen layout is mobile.  
The big modal is disturbing normal reading.

## First proposal

While in mobile layout, can the big modal window be replaced with a small button displaying a disconnected plug? :electric_plug: 
When you push the button you reload the page.   
If the stream restarts, the plug button disappear.

## Second proposal

Can be dismissed forever? UX needs a [x] button to normally close the modal. Then, "Do nothing" button dismiss it forever.

It's a non-sense to press a button, to do nothing. :shrug: 

https://github.com/misskey-dev/misskey/blob/develop/packages/frontend/src/ui/_common_/stream-indicator.vue

ghost:
## What

I added the ability to disable the constant pop-up window about connection loss. [#15852](https://github.com/misskey-dev/misskey/issues/15852)

## Why

Same thing, it's annoying and i think it would be good to disable it without uBlock.

## Checklist
- [ ] (If possible) Add tests


packages/frontend/src/boot/main-boot.ts
```diff
@@ -83,19 +83,21 @@ export async function mainBoot() {
...
-			if (reloadDialogShowing) return;
-			reloadDialogShowing = true;
-			const { canceled } = await confirm({
-				type: 'warning',
-				title: i18n.ts.disconnectedFromServer,
-				text: i18n.ts.reloadConfirm,
-			});
-			reloadDialogShowing = false;
-			if (!canceled) {
+		if (prefer.s.serverDisconnectedBehavior !== 'disabled') {
```
Mar0xy: imo this wasn't even needed to be edited since when it is already marked as anything other than `disabled` it would execute this is also how we solved it on Sharkey we just added an disabled option to the dropdown and that was it no edits done to `main-boot.ts`

Mar0xy: https://activitypub.software/TransFem-org/Sharkey/-/blob/develop/packages/frontend/src/boot/main-boot.ts?ref_type=heads#L75

https://activitypub.software/TransFem-org/Sharkey/-/blob/develop/packages/frontend/src/store.ts?ref_type=heads#L262

ghost: @Mar0xy Improved

PR #15915, opened 2025-04-30
feat: CREATE INDEX CONCURRENTLY for "userId" "id" composite note index if admin wish.

### LINKED ISSUE #15913: Noteのようにでかいテーブルに関するインデックスはconcurrentlyなcreate indexを指定してほしい
### Summary

Noteのようにでかいテーブルに関するインデックスはconcurrentlyなcreate indexを指定してほしい

### Purpose

ライブ更新ができるらしいので

https://www.postgresql.jp/docs/13/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY

- [ ] Yes, I will implement this by myself and send a pull request

anatawa12:
## What

MISSKEY_MIGRATION_CREATE_INDEX_CONCURRENTLY 環境変数が1に設定されてる時に、CREATE INDEX CONCURRENTLYするようにしました。

Closes #15913 since I added comment to suggest CRETE INDEX CONCURRENTLY for adding index in comment

## Why

\
サーバーが十分以上の単位で止まってしまう

## Additional info (optional)

Dev環境で migrate の動作確認はしました。

CREATE INDEX CONCURRENTLYについては公式サイトを参照してください https://www.postgresql.jp/docs/13/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY

## Checklist
- [ ] (If possible) Add tests


anatawa12:
無言でテストが落ちてる？

eternal-flame-AD:
Sorry I didn't note why I didn't write CONCURRENTLY in the original PR. 

マイグレーションスクリプトはデータベースをあるべき状態に移行したら成功する、移行出来なくなったら失敗するすべきと思います。"CONCURENTLYの注意点と利点を理解している上級データベース管理者であれば、この移行をCONCURRENTLYで手動で実行することは可能ですが、デフォルトでこれを行うべきではないと思います。

- "CREATE INDEX CONCURRENTLY"は、その名前にもかかわらず、構築後すぐには使用できず、"CREATE INDEX"**より高速ではありません**（実際には通常は低速です）。
- "CONCURRENTLY" ステートメントすぐに成功できますけど、インデックスは完全に構築されるまで使用できません。_その期間は使用できるインデックスは全く存在しない。_
- ”CONCURRENTLY” 再開するために手動介入が必要となる、high trafficウェブサイトではロック競合のリスクが高まっています (see #15005  #15574) 。テスト環境ロック競合が起こるかどうかを確実にテストすることは難しいとおもいます。


I don't think there is a good way around that except running the migration manually with CONCURRENTLY, as we have to drop an index too so I can't just issue a CREATE INDEX CONCURRENTLY (which would not be immediately usable, and may fail later) and then immediately drop the old one.

*There are too many caveats to do this, I don't think it is ideal to write CONCURRENTLY in migration script, at least without a switch like `MISSKEY_MIGRATION_CREATE_INDEX_CONCURRENTLY=1`. By default a migration script should succeed only if the database status is successfully transitioned.**

> In a concurrent index build, the index is actually entered as an “invalid” index into the system catalogs in one transaction, then two table scans occur in two more transactions. **Before each table scan, the index build must wait for existing transactions that have modified the table to terminate. After the second scan, the index build must wait for any transactions that have a snapshot (see [Chapter 13](https://www.postgresql.org/docs/17/mvcc.html)) predating the second scan to terminate, including transactions used by any phase of concurrent index builds on other tables, if the indexes involved are partial or have columns that are not simple column references**. Then finally the index can be marked “valid” and ready for use, and the CREATE INDEX command terminates. Even then, however, the index may not be immediately usable for queries: **in the worst case, it cannot be used as long as transactions exist that predate the start of the index build.**
> ...
> **If a problem arises while scanning the table, such as a deadlock or a uniqueness violation in a unique index, the CREATE INDEX command will fail but leave behind an “invalid” index. This index will be ignored for querying purposes**because it might be incomplete; however it will still consume update overhead. The psql \d command will report such an index as INVALID:
> **The recommended recovery method in such cases is to drop the index and try again** to perform CREATE INDEX CONCURRENTLY. (Another possibility is to rebuild the index with REINDEX INDEX CONCURRENTLY).
> 
> - [Building Index Concurrently](https://www.postgresql.org/docs/17/sql-createindex.html#SQL-CREATEINDEX-CONCURRENTLY)

eternal-flame-AD:
CREATE INDEX CONCURRENTLY is pretty similar to JS Promise, it isn't "faster" it is just asynchronous, but the quality of service without a usable index on "note" would be probably very poor, even if we can assume the building will eventually succeed without deadlocks (which it will happen on a % of instances), before the index is fully ready the instance would probably be not very usable anyways.

eternal-flame-AD:
On my instance the statement 'CREATE INDEX CONCURRENTLY' takes almost 3x the time, it just yields exclusive lock and waits for much more than needed, but if we write it in migration script it would block the Misskey app from starting regardless so IMO more friendly locking doesn't matter as much.

I think this can be a documentation improvement (or maybe just use CREATE INDEX IF NOT EXISTS to accommodate advanced users to create their index their preferred way) for advanced admins, rather than changing how migration script works (can fail in unintuitive ways and **require manual cleanup to reclaim the index disk space**).

```
misskey-prod@nokogiri ~> podman exec -it misskey-db psql -Umisskey -c 'vacuum (parallel 8) "note"';
VACUUM
misskey-prod@nokogiri ~> time podman exec -it misskey-db psql -Umisskey -c 'create index idx_test_node_userhost_userid O
N "note" ("userHost", "userId")';
CREATE INDEX

________________________________________________________
Executed in   19.46 secs      fish           external
   usr time   80.00 millis    0.00 micros   80.00 millis
   sys time  109.26 millis  927.00 micros  108.33 millis

misskey-prod@nokogiri ~> podman exec -it misskey-db psql -Umisskey -c 'drop index idx_test_node_userhost_userid;';
DROP INDEX
misskey-prod@nokogiri ~> podman exec -it misskey-db psql -Umisskey -c 'vacuum (parallel 8) "note"';
VACUUM
misskey-prod@nokogiri ~> time podman exec -it misskey-db psql -Umisskey -c 'create index concurrently idx_test_node_user
host_userid ON "note" ("userHost", "userId")';
CREATE INDEX

________________________________________________________
Executed in   43.49 secs      fish           external
   usr time  109.62 millis    0.01 millis  109.61 millis
   sys time  117.31 millis    1.17 millis  116.14 millis

misskey-prod@nokogiri ~> podman exec -it misskey-db psql -Umisskey -c 'drop index idx_test_node_userhost_userid;';
DROP INDEX
misskey-prod@nokogiri ~> podman exec -it misskey-db psql -Umisskey -c 'vacuum (parallel 8) "note"';
VACUUM
misskey-prod@nokogiri ~> time podman exec -it misskey-db psql -Umisskey -c 'create index idx_test_node_userhost_userid O
N "note" ("userHost", "userId")';
CREATE INDEX

________________________________________________________
Executed in   17.99 secs      fish           external
   usr time   78.42 millis    0.00 micros   78.42 millis
   sys time  104.03 millis  944.00 micros  103.08 millis
```

anatawa12:
> "CREATE INDEX CONCURRENTLY"は、その名前にもかかわらず、構築後すぐには使用できず、"CREATE INDEX"より高速ではありません（実際には通常は低速です）。

Yes I understand that. I read JP version of the doc and I think cavets to the CREATE INDEX CONCURRENTLY are payable for notes index so I propose this.


> There are too many caveats to do this, I don't think it is ideal to write CONCURRENTLY in migration script, at least without a switch like MISSKEY_MIGRATION_CREATE_INDEX_CONCURRENTLY=1.

I agree with having environment option is better. I 'll impleemnt soon.

I think it need much documentation on CHANGELOG before release.

anatawa12:
It's now under MISSKEY_MIGRATION_CREATE_INDEX_CONCURRENTLY environment flag and default behavior is kept to non-concurrent CREATE INDEX.

eternal-flame-AD:
Thanks, I still highly recommend this one-sentence I believe to be most useful for most users is:

using CONCURRENTLY is slower than not using it (https://github.com/misskey-dev/misskey/pull/15915#issuecomment-2843660605), you should only use it when you do the migration before you take the instance offline to minimize downtime, and sometimes multiple attempts may be required.

Should be something like:

CONCURRENTLY を使用すると、使用しないよりインデックス構築遅くなります。主な利点はダウンタイムを最小限に抑えるために、インスタンスをシャットダウンにする前に使用すべきです。または、大きなインスタンスで複数回の試行が必要になる可能性があります。

eternal-flame-AD:
Something like this should be more robust against issues like interrupted or deadlock during index building that are unique to asynchronous building:

An emulated deadlock or no-opportunity scenario can be replicated by:

- starting a dev instance
-  post 1 note
- start a sql session to lock that note "BEGIN TRANSACTION; SELECT * FROM "note" LIMIT 1", 
- Runt the migration concurrently
- if the administrator press Ctrl-C to interrupt the migration (or somehow disconnected), but the index building later succeeded (if you issue `ROLLBACK` after disconnecting) or timed out, later migration attemptss would be stuck at "IDX_xxxx" already exists.

```js
async up(queryRunner) {
const mayConcurrently = isConcurrentIndexMigrationEnabled();

if (mayConcurrently) {
	const hasValidIndex = await queryRunner.query(`SELECT indisvalid FROM pg_index INNER JOIN pg_class ON pg_index.indexrelid = pg_class.oid WHERE pg_class.relname = 'IDX_724b311e6f883751f261ebe378'`);
	if (!hasValidIndex || hasValidIndex[0].indisvalid !== true) {
		await queryRunner.query(`DROP INDEX IF EXISTS "IDX_724b311e6f883751f261ebe378"`);
		await queryRunner.query(`CREATE INDEX CONCURRENTLY "IDX_724b311e6f883751f261ebe378" ON "note" ("userId", "id" DESC)`);
	}
} else {
	await queryRunner.query(`CREATE INDEX IF NOT EXISTS "IDX_724b311e6f883751f261ebe378" ON "note" ("userId", "id" DESC)`);
}

await queryRunner.query(`DROP INDEX IF EXISTS "IDX_5b87d9d19127bd5d92026017a7"`);
// Flush all cached Linear Scan Plans and redo statistics for composite index
// this is important for Postgres to learn that even in highly complex queries, using this index first can reduce the result set significantly
await queryRunner.query(`ANALYZE "user", "note"`);
}

```

anatawa12:
Changelogを、この新たなオプションの注意点とともに書きました。

syuilo:
🙏🏻

PR #16185, opened 2025-06-11
enhance(backend): Allow renote of public notes with any visibility

### LINKED ISSUE #12864: allow quote renotes to be Public where the target post is Unlisted?
### 💡 Summary

For some reason even though quote's visibility is explicitly set to Public (and for some reason too it defaults to Public instead of the visibility of the target post), when the target post is Unlisted (or "Home" visibility in Misskey) the quote will also be set to Unlisted regardless of the visibility you set for it.

### 🥰 Expected Behavior

The visibility setting of the quote renote first and foremost should be by default set to the same visibility as the target note.

And quote renoting in Public should always be possible for quotable posts.

### 🤬 Actual Behavior

Default visibility is set to Public regardless of the target post's visibility

Quote renoting posts whose visibility is "Home" will always be Unlisted/Home regardless of what the visibility setting is at

### 📝 Steps to Reproduce

_No response_

### 💻 Frontend Environment

```markdown
* Model and OS of the device(s): Windows 11 22H2 x64
* Browser: Floorp 11.7.1
* Server URL: makai.chaotic.ninja
* Misskey: 2023.11.1
```

```markdown
* Misskey:
* Node:
* PostgreSQL:
* Redis:
* OS and Architecture:
```

eternal-flame-AD:
## What

Fixes #12864 

## Why

See https://github.com/misskey-dev/misskey/issues/12864#issuecomment-2961541557 , and subsequent discussions.

## Checklist
- [] Test working in a local environment
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/core/NoteCreateService.ts
+++ b/packages/backend/src/core/NoteCreateService.ts
@@ -280,13 +280,8 @@ export class NoteCreateService implements OnApplicationShutdown {
 		if (data.renote) {
 			switch (data.renote.visibility) {
 				case 'public':
-					// public noteは無条件にrenote可能
-					break;
 				case 'home':
-					// home noteはhome以下にrenote可能
-					if (data.visibility === 'public') {
-						data.visibility = 'home';
-					}
+					// notes with public audience can have any visibility
 					break;
 				case 'followers':
 					// 他人のfollowers noteはreject


eternal-flame-AD:
Hmmm, CI seems to be failing for unrelated reasons, maybe it needs to be re-run?

KisaragiEffective (changes requested):
I do not agree with this change without proper discussions in the linked issue.

PR #16404, opened 2025-08-13
enhance: performance for CleanRemoteNotesProcessorService

### LINKED ISSUE #16403: Improve NoteCleanRemoteNoteServices query performance
### Summary

Playground script with both old and new query: https://gist.github.com/eternal-flame-AD/ad7436e61de5213ffa30dbbc9d0d0caa 

Before:

```
○ Total Cost: 13,632,562,524
○ Planning Time: 2.04 ms
○ Execution Time: 23.72 s
```

After:

```
○ Total Cost: 3,040,828,040
○ Planning Time: 1.61 ms
○ Execution Time: 1.44 s
```

### Purpose

- Large perf improvement.
- The new query should now be able to be completely written in ORM instead of raw SQL.
- Related: #16395

- [x] Yes, I will implement this by myself and send a pull request

eternal-flame-AD:
## What

Fixes #16403.

Maybe good enough for #16395 .

## Checklist
- [X] Test working in a local environment (unit test only)
- [ ] (If possible) Add tests


packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
```diff
@@ -66,70 +77,112 @@ export class CleanRemoteNotesProcessorService {
...
-
-		const notesQuery = this.notesRepository.createQueryBuilder('note')
-			.addCommonTableExpression(recursiveQuery, 'related_notes', { recursive: true })
+		const removalCriteria = (qb: SelectQueryBuilder<MiNote>) => {
+			return qb
+				.where('note."id" < :newestLimit')
+				.andWhere('note."clippedCount" = 0')
+				.andWhere('note."userHost" IS NOT NULL')
+				.andWhere('NOT EXISTS (SELECT 1 FROM user_note_pining WHERE "noteId" = note."id")')
+				.andWhere('NOT EXISTS (SELECT 1 FROM note_favorite WHERE "noteId" = note."id")');
```
anatawa12: I want to add a comment indicating that there is another place (currently addSelect at line 155) where this criteria is computed, to prevent mistakes when changing conditions in the future. 
Otherwise, I would prefer to use the literal as before.

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
```diff
@@ -66,70 +77,112 @@ export class CleanRemoteNotesProcessorService {
...
+		const candidateNotesQuery = (cursorRight?: MiNote['id']) => this.db.createQueryBuilder()
+			.select('candidate_notes.id', 'id')
+			.addCommonTableExpression(
+				`(${candidateNotesQueryBase(cursorRight).getQuery()} UNION ${candidateNotesQueryInductive.getQuery()})`,
+				candidateNotesCteName,
+				{ recursive: true },
+			)
+			.from(candidateNotesCteName, candidateNotesCteName)
+			.leftJoin(candidateNotesCteName, 'unremovable', `unremovable."rootId" = "${candidateNotesCteName}"."rootId" AND unremovable."isRemovable" = FALSE`)
+			.where('unremovable."id" IS NULL');
```
anatawa12: I understand the query produces the correct results, but I’m having trouble understanding how LEFT OUTER JOIN together with IS NULL achieves that.
Therefore, I think it' good to add some comment describe how it work.

```
`candidate_notes` will have the following structure after recursive query (some columns omitted):

| id  | rootId | isRemovable |
|-----|--------|-------------|
| aaa | aaa    | TRUE        |
| bbb | aaa    | FALSE       |
| ccc | aaa    | FALSE       |
| ddd | ddd    | TRUE        |
| eee | ddd    | TRUE        |
| fff | fff    | TRUE        |
| ggg | ggg    | FALSE       |

After performing a LEFT JOIN with `candidate_notes` as `unremovable`,
the note tree containing unremovable notes will be joined.  
For removable rows, the `unremovable` columns will have `NULL` values.

| id  | rootId | isRemovable | | id   | rootId | isRemovable |
|-----|--------|-------------|-|------|--------|-------------|
| aaa | aaa    | TRUE        | | bbb  | aaa    | FALSE       |
| aaa | aaa    | TRUE        | | ccc  | aaa    | FALSE       |
| bbb | aaa    | FALSE       | | bbb  | aaa    | FALSE       |
| bbb | aaa    | FALSE       | | ccc  | aaa    | FALSE       |
| ccc | aaa    | FALSE       | | bbb  | aaa    | FALSE       |
| ccc | aaa    | FALSE       | | ccc  | aaa    | FALSE       |
| ddd | ddd    | TRUE        | | NULL | NULL   | NULL        |
| eee | ddd    | TRUE        | | NULL | NULL   | NULL        |
| fff | fff    | TRUE        | | NULL | NULL   | NULL        |
| ggg | ggg    | FALSE       | | ggg  | ggg    | FALSE       | 

Therefore, filtering rows where `unremovable.id` is `NULL`  
will produce the note tree that contains removable notes.

| id  | rootId | isRemovable | | id   | rootId | isRemovable |
|-----|--------|-------------|-|------|--------|-------------|
| ddd | ddd    | TRUE        | | NULL | NULL   | NULL        |
| eee | ddd    | TRUE        | | NULL | NULL   | NULL        |
| fff | fff    | TRUE        | | NULL | NULL   | NULL        | 
```

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
```diff
@@ -142,32 +195,39 @@ export class CleanRemoteNotesProcessorService {
 				break;
 			}
 
-			job.updateProgress((elapsed / maxDuration) * 100);
+			const progress = this.computeProgress(minId, maxId, cursorLeft);
+			const wallClockUsage = elapsed / maxDuration;
+			if (wallClockUsage > 0.5 && progress < 50 && !lowThroughputWarned) {
+				this.logger.warn(`Not projected to finish in time! (wall clock usage ${wallClockUsage * 100}% at ${progress}%, current limit ${currentLimit})`);
```
anatawa12: I think it's good to add this information to the job log `job.log` in addition to normal logger.

Admins can view the job log from the job queue page in the control panel, so it’s much easier to check.

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
```diff
@@ -177,19 +237,23 @@ export class CleanRemoteNotesProcessorService {
...
+			cursorLeft = cursorRight;
+
+			job.log(`Deleted ${noteIds.length} notes; ${Date.now() - batchBeginAt}ms`);
+
+			if (process.env.NODE_ENV !== 'test') {
+				await setTimeout(1000 * 5); // Wait a moment to avoid overwhelming the db
+			}
+		} while (cursorLeft < maxId);
 
 		this.logger.succ('cleaning of remote notes completed.');
```
anatawa12, line 277: I think it would be good to log `this.computeProgress(minId, maxId, cursorLeft)`, or include it in the job output, to check the progress at the end of the job.

anatawa12 (approved):
The overall logic and query look good to me.

I have a few minor suggestions to improve the feature and enhance code readability.

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
eternal-flame-AD: Not sure what you mean, isn't the progress always 100% at the end?

I added a log if it is timed out instead of completed

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
eternal-flame-AD: I will add an intermediate view to simplify this, and add the comments you suggested

eternal-flame-AD:
Change: added some robustness checks (shrinking batch size instead of throwing when statement timeout and continuing for transient constraint violations).

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
anatawa12: When the clean finishes due to a timeout, the while loop will break instead of returning early. Adding a log here will show the progress whether it timed out or completed 100%. However, it’s enough to log only on timeout.

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
```diff
@@ -77,73 +79,51 @@ export class CleanRemoteNotesProcessorService {
...
-				.andWhere('note."renoteId" IS NULL')
-				.andWhere(removalCriteria)
-				.orderBy('note.id', 'ASC')
-				.offset(limit)
-				.setParameters({ newestLimit, cursorLeft })
-				.getRawOne<{ id: MiNote['id'] }>().then(result => result?.id);
+		// start with a conservative limit and adjust it based on the query duration
+		const minimumLimit = 10;
+		let currentLimit = 100;
+		let cursorLeft = minId.slice(0, -1);
```
anatawa12: Can you add the content describes the reason of the slicing the id?
I think this is unnecessary.

anatawa12 (approved):
LGTM except one place.

packages/backend/src/queue/processors/CleanRemoteNotesProcessorService.ts
eternal-flame-AD: CursorLeft is open range so this ensures the first ever cursor is strictly less than the first entry .. 

I did think about just generating a 'decrement' ID but think this is good enough.

anatawa12: Oh sorry I forget it's open rather than close.
How about simply using '0' like original code? Does using '0' will degrade the performance much?

eternal-flame-AD: We can start with '0' , it shouldn't affect that much I agree with you, I was just worried about edge cases with progress computation as currently the progress is calculated w.r.t. ID space (approximately real world time) not note count (which is expensive to compute).

I will just clamp the progress to use minID then.

anatawa12 (approved):
LGTM

syuilo:
🙏🏻🙏🏻

PR #16514, opened 2025-09-03
perf(backend): optimize drive/files query for old root content

### LINKED ISSUE #14217: ドライブが開かないユーザーがいる
### 💡 Summary

色々な条件が重なるとuserId, folderId, id のIndexが採用されず、idのPK Indexが採用された結果タイムアウトするらしい

具体的には、ドライブのレコード数が大量、直近の31件のファイルのアップロードがとても少ないことが条件っぽい？
limit 31 で31件集めるのにタイムアウトする模様

### 🥰 Expected Behavior

狙ったIndexが採用される
スクリーンショット 2024-07-15 18 57 42

### 🤬 Actual Behavior

```
misskey@misskeydb ERROR: canceling statement due to statement timeout
2024-07-15 06:40:16.410 UTC [3609383] misskey@misskeydb statement:
select
	"file"."id" as "file_id",
	"file"."userId" as "file_userId",
	"file".
"userHost" as "file_userHost",
	"file"."md5" as "file_md5",
	"file"."name" as "file_name",
	"file"."type" as "file_type",
	"file"."size" as "file_
size",
	"file"."comment" as "file_comment",
	"file"."blurhash" as "file_blurhash",
	"file"."properties" as "file_properties",
	"file"."storedInter
nal" as "file_storedInternal",
	"file"."url" as "file_url",
	"file"."thumbnailUrl" as "file_thumbnailUrl",
	"file"."webpublicUrl" as "file_webpub
licUrl",
	"file"."webpublicType" as "file_webpublicType",
	"file"."accessKey" as "file_accessKey",
	"file"."thumbnailAccessKey" as "file_thumbnai
lAccessKey",
	"file"."webpublicAccessKey" as "file_webpublicAccessKey",
	"file"."uri" as "file_uri",
	"file"."src" as "file_src",
	"file"."folderI
d" as "file_folderId",
	"file"."isSensitive" as "file_isSensitive",
	"file"."maybeSensitive" as "file_maybeSensitive",
	"file"."maybePorn" as "fi
le_maybePorn",
	"file"."isLink" as "file_isLink",
	"file"."requestHeaders" as "file_requestHeaders",
	"file"."requestIp" as "file_requestIp"
from
	"drive_file" "file"
where
	"file"."userId" = $1
	and "file"."folderId" is null
order by
	"file"."id" desc
limit 31
```

### 📝 Steps to Reproduce

```
explain 
select
	"file"."id" as "file_id",
	"file"."userId" as "file_userId",
	"file".
"userHost" as "file_userHost",
	"file"."md5" as "file_md5",
	"file"."name" as "file_name",
	"file"."type" as "file_type",
	"file"."size" as "file_size",
	"file"."comment" as "file_comment",
	"file"."blurhash" as "file_blurhash",
	"file"."properties" as "file_properties",
	"file"."storedInternal" as "file_storedInternal",
	"file"."url" as "file_url",
	"file"."thumbnailUrl" as "file_thumbnailUrl",
	"file"."webpublicUrl" as "file_webpublicUrl",
	"file"."webpublicType" as "file_webpublicType",
	"file"."accessKey" as "file_accessKey",
	"file"."thumbnailAccessKey" as "file_thumbnailAccessKey",
	"file"."webpublicAccessKey" as "file_webpublicAccessKey",
	"file"."uri" as "file_uri",
	"file"."src" as "file_src",
	"file"."folderId" as "file_folderId",
	"file"."isSensitive" as "file_isSensitive",
	"file"."maybeSensitive" as "file_maybeSensitive",
	"file"."maybePorn" as "file_maybePorn",
	"file"."isLink" as "file_isLink",
	"file"."requestHeaders" as "file_requestHeaders",
	"file"."requestIp" as "file_requestIp"
from
	"drive_file" "file"
where
	"file"."userId" = '9bcuu9u19o'
	and "file"."folderId" is null
order by
	"file"."id" desc
limit 31
```

```
Limit  (cost=0.43..5562.96 rows=31 width=1806)
  ->  Index Scan Backward using "PK_43ddaaaf18c9e68029b7cbb032e" on drive_file file  (cost=0.43..338238.08 rows=1885 width=1806)
        Filter: (("folderId" IS NULL) AND (("userId")::text = '9bcuu9u19o'::text))
```

> PK_43ddaaaf18c9e68029b7cbb032e

これがIndexとして採用されている

### 💻 Frontend Environment

```markdown
* Browser:
* Server URL: 
* Misskey:
```

```markdown
https://misskey.systems/

* Misskey:  2024.5.0 or later
* Node: v20.10.0
* PostgreSQL:  PostgreSQL 15.7 (Ubuntu 15.7-1.pgdg22.04+1) on x86_64-pc-linux-gnu, compiled by gcc (Ubuntu 11.4.0-1ubuntu1~22.04) 11.4.0, 64-bit
* Redis:
* OS and Architecture:
MisskeySystems ~# lsb_release -a
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.4 LTS
Release:	22.04
Codename:	jammy
```

caipira113:
## What
Resolve #14217 

This PR speeds up the `drive/files` endpoint for users who have old content in the root folder by adding a partial index on `drive_file (userId, folderId IS NULL, id DESC)`. There are no API or behavior changes; it only adds a migration.

## Why
Some users were hitting statement timeouts because PostgreSQL chose the primary key for `ORDER BY id DESC`, leading to very wide scans when their root files were old. The new index matches the query shape so the planner can return the latest matching rows immediately.

## Additional info (optional)
On an affected account, the query improved from about 50 seconds to about 30-40 ms.

**[ TEST Query ]**
```
EXPLAIN (ANALYZE, BUFFERS) SELECT "file"."id", "file"."userId", "file"."userHost", "file"."md5", "file"."name", "file"."type", "file"."size", "file"."comment", "file"."blurhash", "file"."properties", "file"."storedInternal", "file"."url", "file"."thumbnailUrl", "file"."webpublicUrl", "file"."webpublicType", "file"."accessKey", "file"."thumbnailAccessKey", "file"."webpublicAccessKey", "file"."uri", "file"."src", "file"."folderId", "file"."isSensitive", "file"."maybeSensitive", "file"."maybePorn", "file"."isLink", "file"."requestHeaders", "file"."requestIp" FROM "drive_file" "file" WHERE "file"."userId" = '8zzt247op3' AND "file"."folderId" IS NULL ORDER BY "file"."id" DESC LIMIT 31;"
```

**[ Before ]**
```
QUERY PLAN
Limit  (cost=0.43..6225.04 rows=31 width=1997) (actual time=211.896..51068.746 rows=22 loops=1)
  Buffers: shared hit=1702945 read=701470
  ->  Index Scan Backward using "PK_43ddaaaf18c9e68029b7cbb032e" on drive_file file  (cost=0.43..1091916.33 rows=5438 width=1997) (actual time=211.893..51068.629 rows=22 loops=1)
        Filter: (("folderId" IS NULL) AND (("userId")::text = '8zzt247op3'::text))
        Rows Removed by Filter: 4547669
        Buffers: shared hit=1702945 read=701470
Planning:
  Buffers: shared hit=358 read=68
Planning Time: 19.364 ms
Execution Time: 51068.991 ms
(10 rows)
```

<video src="https://github.com/user-attachments/assets/6fa52988-3bc5-4754-98fa-b24481465a66" controls preload></video>

**[ After ]**
```
QUERY PLAN
Limit  (cost=0.43..118.09 rows=31 width=1997) (actual time=6.629..30.655 rows=22 loops=1)
  Buffers: shared hit=3814
  ->  Index Scan Backward using "IDX_a76118b66adb3228e0ee69c281" on drive_file file  (cost=0.43..25190.51 rows=6637 width=1997) (actual time=6.627..30.648 rows=22 loops=1)
        Index Cond: (("userId")::text = '8zzt247op3'::text)
        Filter: ("folderId" IS NULL)
        Rows Removed by Filter: 4898
        Buffers: shared hit=3814
Planning:
  Buffers: shared hit=463 dirtied=13
Planning Time: 18.885 ms
Execution Time: 30.709 ms
(11 rows)
```

<video src="https://github.com/user-attachments/assets/79a7a283-5726-472b-abfa-c425e4a7e2db" controls preload></video>

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/migration/1756824506000-optimize-drive-files-query.js
+++ b/packages/backend/migration/1756824506000-optimize-drive-files-query.js
@@ -1,0 +1,19 @@
+/*
+ * SPDX-FileCopyrightText: syuilo and misskey-project
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+export class OptimizeDriveFilesQuery1756824506000 {
+	name = 'OptimizeDriveFilesQuery1756824506000'
+
+	async up(queryRunner) {
+		// Create optimized partial index for drive files query performance
+		// This index is specifically designed for the common query pattern:
+		// SELECT * FROM drive_file WHERE userId = ? AND folderId IS NULL ORDER BY id DESC
+		await queryRunner.query(`CREATE INDEX "IDX_drive_file_userid_null_folderid_id_desc" ON "drive_file" ("userId", ("folderId" IS NULL), "id" DESC)`);
+	}
+
+	async down(queryRunner) {
+		await queryRunner.query(`DROP INDEX "IDX_drive_file_userid_null_folderid_id_desc"`);
+	}
+}
--- a/packages/backend/migration/1756824506000-optimize-drive-files-query.js
+++ b/packages/backend/migration/1756824506000-optimize-drive-files-query.js
@@ -4,16 +4,16 @@
  */
 
 export class OptimizeDriveFilesQuery1756824506000 {
-	name = 'OptimizeDriveFilesQuery1756824506000'
+    name = 'OptimizeDriveFilesQuery1756824506000'
 
-	async up(queryRunner) {
-		// Create optimized partial index for drive files query performance
-		// This index is specifically designed for the common query pattern:
-		// SELECT * FROM drive_file WHERE userId = ? AND folderId IS NULL ORDER BY id DESC
-		await queryRunner.query(`CREATE INDEX "IDX_drive_file_userid_null_folderid_id_desc" ON "drive_file" ("userId", ("folderId" IS NULL), "id" DESC)`);
-	}
+    async up(queryRunner) {
+        // Create optimized composite index for drive files query performance
+        await queryRunner.query(`DROP INDEX "IDX_860fa6f6c7df5bb887249fba22"`);
+        await queryRunner.query(`CREATE INDEX "IDX_a76118b66adb3228e0ee69c281" ON "drive_file" ("userId", "id" DESC)`);
+    }
 
-	async down(queryRunner) {
-		await queryRunner.query(`DROP INDEX "IDX_drive_file_userid_null_folderid_id_desc"`);
-	}
+    async down(queryRunner) {
+        await queryRunner.query(`DROP INDEX "IDX_a76118b66adb3228e0ee69c281"`);
+        await queryRunner.query(`CREATE INDEX "IDX_860fa6f6c7df5bb887249fba22" ON "drive_file" ("userId")`);
+    }
 }
--- a/packages/backend/src/models/DriveFile.ts
+++ b/packages/backend/src/models/DriveFile.ts
@@ -10,12 +10,11 @@ import { MiDriveFolder } from './DriveFolder.js';
 
 @Entity('drive_file')
 @Index(['userId', 'folderId', 'id'])
-@Index('IDX_drive_file_userid_null_folderid_id_desc', { synchronize: false })
+@Index(['userId', 'id'])
 export class MiDriveFile {
 	@PrimaryColumn(id())
 	public id: string;
 
-	@Index()
 	@Column({
 		...id(),
 		nullable: true,
--- a/packages/backend/src/models/DriveFile.ts
+++ b/packages/backend/src/models/DriveFile.ts
@@ -10,6 +10,7 @@ import { MiDriveFolder } from './DriveFolder.js';
 
 @Entity('drive_file')
 @Index(['userId', 'folderId', 'id'])
+@Index('IDX_drive_file_userid_null_folderid_id_desc', { synchronize: false })
 export class MiDriveFile {
 	@PrimaryColumn(id())
 	public id: string;


eternal-flame-AD:
Is this because of the ASC/DESC?

It seems the current index is `(folderID ASC, id ASC)` and that might cause bad query plans if the request is id DESC and some awkward filters are applied. Query planner might assume the data is well distributed and underestimate the cost of scanning the whole table.

If the reason is index order we should probably create index for `(..., folderID, id DESC)` instead of partial ones because descending queries on subfolders are legal API calls too.

Or maybe just simply CHANGE IDX_860fa6f6c7df5bb887249fba22 `("userId")` to `("userId", "id")` (which also will help with "list all files for a user in chronological list" case-useful for moderation)? This will 100% be chosen over a PK scan because it has the same depth and is more specific.

caipira113:
Thanks for the comment @eternal-flame-AD
I re‑ran the queries on the problematic user and a control user to validate the hypotheses. Changing the composite index from `(userId, folderId, id)` to use DESC on id did not fix the root‑folder case; on this dataset the planner continued to prefer a backward scan on the primary key and the query still took on the order of tens of seconds. By contrast, replacing the single‑column index on userId with a composite `(userId, id)` consistently made the planner choose that path and brought the worst‑case root‑folder query down from ~43s to ~40ms. The "all files for a user in reverse chronological order" path also benefits and becomes sub‑millisecond, while subfolder queries remain well served by the existing `(userId, folderId, id)` composite.

eternal-flame-AD:
Thanks! I wonder why postgres refuse to use the original index then .. but yes .. I think (userId, id) should do because there is a practical limit on how many files a single user can create without it being abusing the service so the worst case performance of index scan over list of all files owned by a specific user in ID order should be good enough.

Doesn't seem like there is currently a role limit on # of drive files, but I would say it is more useful to add a limit than to try to optimize the query for that case-because that requires the file storage logic to not rely on the file system (which is basically another index) as well.

PR #17222, opened 2026-03-06
feat: add localHost option to allow changing of federation host

### LINKED ISSUE #6724: Accept use domain different from web access
## Summary
In Mastodon we can set LOCAL_DOMAIN for the domain part of ActivityPub users and WEB_DOMAIN for the web access part of it.
This way we can use any domain for access mastodon but the users still match the main domain (like email, for example).
We can have an user@example.com email, but access it from webmail.example.com.

As we can see in https://docs.joinmastodon.org/admin/config/ there are an option for it.

And this doc explain how to do it https://github.com/felx/mastodon-documentation/blob/master/Running-Mastodon/Serving_a_different_domain.md

kenshineto:
## What
Add a `userHost` option in config that allows instances to use a seperate host for their user ids instead of using the host from the instance url.

Fix #6724

## Why
This mirrors a feature that mastodon has with its `LOCAL_URL` and `WEB_URL` options. It allows instances to host their instances on any url without having to compromise their user ids. For example I am running this change on my instance `social.freya.cat` with all user ids `@freya.cat`.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/.config/cypress-devcontainer.yml
+++ b/.config/cypress-devcontainer.yml
@@ -21,6 +21,11 @@ setupPassword: example_password_please_change_this_or_you_will_get_hacked
 # Final accessible URL seen by a user.
 url: 'http://misskey.local'
 
+# Use a seperate hostname for federation. All users on this
+# instance will appear with this host instead of the one with
+# the misskey web url if set.
+#localHost: 'misskey.local'
+
 # ONCE YOU HAVE STARTED THE INSTANCE, DO NOT CHANGE THE
 # URL SETTINGS AFTER THAT!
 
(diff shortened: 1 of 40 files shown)


kenshineto:
I use to have this setup where config.url was changed to `webUrl`, and `localUrl` was the option added. This ended up changing most of the files within the project, which was honestly too many. So this PR as its written now only modifies the places where `userHost` should be used instead.

I manually went though every use of `url`, `host,` and `hostname` and figured out (to by best judgement) which one should be used (url or userHost). Hopefully I got them all :)

kenshineto:
Renamed option form `userHost` to `localHost` since `userHost` is already used in misskey.

PR #17275, opened 2026-04-01
fix(backend): handle array or string in alsoKnownAs

### LINKED ISSUE #17274: Fail to parse an actor with a single `alsoKnownAs` value
### 💡 Summary

In Activity Streams 2.0, a one-element array is required (not optional) to be compacted to a single value. When testing with https://tags.pub/ , I see this error from Misskey:

```
ERR  *    [api]    Internal error occurred in users/show: i.alsoKnownAs.map is not a function {
  ep: 'users/show',
  ps: { username: 'example', host: 'tags.test', i: 'wjX7bMOR6DZJUaJj' },
  e: {
    message: 'i.alsoKnownAs.map is not a function',
    code: 'TypeError',
    stack: 'TypeError: i.alsoKnownAs.map is not a function\n' +
      '    at j.pack (file:///misskey/packages/backend/built/boot/entry.js:2:200523)\n' +
      '    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)\n' +
      '    at async file:///misskey/packages/backend/built/boot/entry.js:137:117057\n' +
      '    at async ds.call (file:///misskey/packages/backend/built/boot/entry.js:137:177315)',
    id: '0e9345d1-1bf9-4336-9ca0-5f53959e8d0e'
  }
}
```

It looks like it's expecting an array. Here's what the value from tags.pub server looks like:

```
  "alsoKnownAs": "acct:example@tags.test",
```

Misskey should accept both arrays and strings in this property.

### 🥰 Expected Behavior

Continue without failing

### 🤬 Actual Behavior

Stack trace in the logs

### 📝 Steps to Reproduce

_No response_

### 💻 Frontend Environment

```markdown
* Browser:
* Server URL:
* Misskey:
```

```markdown
* Misskey:
* Node:
* PostgreSQL:
* Redis:
* OS and Architecture:
```

evanp:
This bug happens if a remote actor has a single value in `alsoKnownAs`. I added a test to check that it does indeed fail. Then, I added a utility function to call with both array or string values. I then check for an array value or a string value and call the utility correctly for each type. As a fallback, it sets the value to `null`.

## Checklist
- [x] (If possible) Add tests


kakkokari-gtyih:
@evanp Could you review the checklist? (Normally it's appeared on the pull request comment section but it's not present, so I added it for you)

evanp:
> @evanp Could you review the checklist? (Normally it's appeared on the pull request comment section but it's not present, so I added it for you)

Thank you so much! I checked off a lot. 

I think there was a recommendation on the issue to use `toArray`, so I will do that. 

I will also add a Changelog line.

evanp:
All right. I changed the code to use the more idiomatic `toArray`. I also updated ApPersonService to handle array or singleton values. I added tests for UserEntityService and ApPersonService. And I added a line to the Changelog (in English -- watashi wa nihongo ga jozu jyanai desu).

syuilo:
Thank you🙏

PR #17340, opened 2026-04-24
fix: change bare activity.actor to getApId(activity.actor) in InboxPr…

### LINKED ISSUE #17338: Embedded object for `actor` in HTTP Signature verification
### 💡 Summary

In [InboxProcessorService.ts](https://github.com/misskey-dev/misskey/blob/3a3057a1b10e6dc04364bd2cff70ee0370408b1d/packages/backend/src/queue/processors/InboxProcessorService.ts#L139), the signing authority's URI is compared to the `actor` property of the activity.

As with many other properties of ActivityPub objects, `actor` can be a string URI, an embedded object, or an array of strings and/or objects.

This comparison should work with any of these constructions; they are all valid ActivityPub activities.

### 🥰 Expected Behavior

The comparison should check the type of `actor`:

- if it's a string, stringwise comparison
- if it's an object, compare against its `id` property (can be a string or array, too!)
- if it's an array, compare against each element

This would be a useful global utility.

### 🤬 Actual Behavior

The stringwise comparison fails when `actor` is an object.

### 📝 Steps to Reproduce

_No response_

### 💻 Frontend Environment

```markdown
* Browser:
* Server URL:
* Misskey:
```

```markdown
* Misskey:
* Node:
* PostgreSQL:
* Redis:
* OS and Architecture:
```

evanp:
## What

Wrapped all uses of `activity.actor` as a string in InboxProcessorService with `getApId()`
Wrapped all uses of `activity.actor` as a string in ApInboxService with `getApId()`

## Why

Misskey silently rejects activities with an `actor` property that is an embedded object. `getApId()` will extract the id whether the property value is an embedded object or a URI.

Several uses of `activity.actor` in these modules already used the wrapper; this change covers the rest. I confirmed with each use that the wrapper was appropriate.

## Additional info (optional)

I was able to confirm that existing tests pass with this change. I manually tested with my own software, which includes `actor` as an embedded object, in a Docker Compose environment. Neither InboxProcessorService nor ApInboxService has a unit test suite.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,10 +10,8 @@
 - Fix: 連合が無効化されたサーバーで一部の設定項目が空欄で表示される問題を修正
 
 ### Server
-
-- Fix: support activity with `actor` as an id string or embedded object in inbox processor and ActivityPub inbox service
 - Enhance: メモリ使用量を削減
-- Enhance: 起動の高速化
+- Enhance: 起動の高速化  
   (Cherry-picked from https://github.com/MisskeyIO/misskey/pull/1410)
 - Enhance: バックエンドの開発モード時の安定性向上
 - Enhance: バックエンドビルド・テスト時に使用する依存関係の整理（swc/esbuild→Rolldown, Jest→Vitest）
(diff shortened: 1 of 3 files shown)


evanp:
I ran this through tests with tags.pub -- it raised some additional problems in ApInboxService.ts. So, I wrapped those comparisons in getApId() also. They are all comparisons with actor.uri.

evanp:
Looking at the failing E2E test. Is there a way to run it again? I'm not sure anything in there touches the inbox, so I'm concerned this might just be a transient test error.

syuilo:
Thank you!

evanp:
Thank *you*!

PR #17356, opened 2026-05-02
fix(backend): Acquire lock of Announce object in announceNote even if it is from a relay actor

tirr-c:
## What
In `announceNote`, acquire lock of `Announce` activity even if it came from a relay actor.

## Why
- Locking `activity` object (instead of `target`) is the correct one because `announceNote` operates on the `Announce` activity.
- Following call to `resolveNote` tries to acquire lock of the target activity, which will always fail for relays currently.

## Additional info (optional)
Tested on mitir.social

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/backend/src/core/activitypub/ApInboxService.ts
+++ b/packages/backend/src/core/activitypub/ApInboxService.ts
@@ -313,7 +313,8 @@ export class ApInboxService {
 		// アナウンス先が許可されているかチェック
 		if (!this.utilityService.isFederationAllowedUri(uri)) return;
 
-		const unlock = await acquireApObjectLock(this.redisClient, uri);
+		const activityUri = getApId(activity);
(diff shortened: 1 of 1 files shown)


packages/backend/src/core/activitypub/ApInboxService.ts
```diff
@@ -313,7 +313,8 @@ export class ApInboxService {
 		// アナウンス先が許可されているかチェック
 		if (!this.utilityService.isFederationAllowedUri(uri)) return;
```
tirr-c, line 314: Not sure if this should also be `activityUri`...

tirr-c:
The test passed locally, it seems flaky...

kakkokari-gtyih:
:eyes:

syuilo:
🙏🏻

PR #17401, opened 2026-05-14
fix(backend): route fastify listen/ready/close errors through logger instead of unhandled rejection

### LINKED ISSUE #17399: ServerService: startup and shutdown failures surface as UnhandledPromiseRejection / SIGKILL instead of via logger
### Summary

`packages/backend/src/server/ServerService.ts` has three error-handling regressions around `fastify.listen()`, `fastify.ready()`, and `fastify.close()` that cause startup and shutdown failures to be reported as uncaught Node exceptions instead of through the misskey logger the code clearly intended to use.

The code at `ServerService.ts:245-264` registers a `fastify.server.on('error', ...)` listener that switches on `EACCES` / `EADDRINUSE` to log a user-friendly message and call `process.exit(1)`. That listener was reachable when misskey was on fastify v4 with callback-style `listen()`, but on fastify v5 the listen errors surface as a rejection of the `listen()`-returned Promise — which is fire-and-forget at `ServerService.ts:276` and callback-with-ignored-`err` at `ServerService.ts:270`. On Node >= 15 with default `--unhandled-rejections=throw`, the rejection terminates the process before `server.on('error')` can run. The handler is reachable today only by accident of timing.

In addition, `dispose()` at `ServerService.ts:283-286` calls `await this.#fastify.close()` with no timeout and no catch. fastify@5's `close()` waits indefinitely for upgraded WebSocket connections to drain. `streamingApiServerService.attach(fastify.server)` adds raw `ws.Server` upgrades that don't get clean close frames; `close()` hangs until PM2/systemd/k8s sends SIGKILL.

### Expected Behavior

1. On `EADDRINUSE`: misskey logs `"Port 3000 is already in use by another process."` and exits with code 1.
2. On `EACCES` on a socket path: misskey logs `"You do not have permission to listen on /var/run/misskey.sock."` and exits with code 1.
3. On plugin-registration timeout (`ERR_AVVIO_PLUGIN_TIMEOUT`) during `await fastify.ready()`: misskey logs the failure through `this.logger.error` and exits cleanly.
4. On graceful shutdown (SIGTERM): `dispose()` resolves within a bounded time even if WebSocket upgrades are leaked, so PM2/systemd records a clean exit instead of timing out and SIGKILLing.

### Actual Behavior

1. `EADDRINUSE`: `UnhandledPromiseRejectionWarning` stack trace + Node terminates the process; the EADDRINUSE switch case in `server.on('error')` is dead code.
2. `EACCES` on socket: `fs.chmodSync('/var/run/misskey.sock', ...)` throws `ENOENT: no such file or directory` (because no socket was created), masking the actual permission error. User sees the chmod error in the log, not the listen error.
3. Plugin timeout: NestJS bootstrap logger emits a generic stack trace; the misskey logger sees nothing useful.
4. Graceful shutdown with leaked WS: `dispose()` hangs forever. PM2's default `kill_timeout` (or systemd's `TimeoutStopSec`) then SIGKILLs the process; logs show "process killed" with no diagnostic.

### Steps to Reproduce

**Repro 1 — TCP `EADDRINUSE`:**

```sh
# Terminal 1
nc -l 3000

# Terminal 2 (Misskey configured with `port: 3000`)
pnpm start
# Observed: UnhandledPromiseRejectionWarning + stack trace, no friendly message
```

**Repro 2 — Unix socket without permission:**

Configure `.config/default.yml` with:

```yaml
socket: /var/run/misskey.sock
chmodSocket: '777'
```

Run as a non-root user. `fastify.listen({ path })` rejects with `EACCES`; the callback's `err` is ignored; then `fs.chmodSync('/var/run/misskey.sock', ...)` throws `ENOENT`. The chmod error is what gets logged.

**Repro 3 — `dispose()` hang:**

1. Open a WebSocket client connection to `wss://<misskey>/streaming` (via `streamingApiServerService`).
2. Force-close the TCP connection without sending a WS close frame (kill the client process).
3. Send SIGTERM to misskey (`pm2 stop misskey` or `systemctl stop misskey`).
4. Observed: `dispose()` hangs; after the supervisor's kill timeout, misskey is SIGKILL'd.

### Frontend Environment

```markdown
* Model and OS of the device(s): N/A — backend-only bug
* Browser: N/A
* Server URL: N/A — affects any misskey instance
* Misskey: 2026.5.2-alpha.0 (verified against develop HEAD)
```

### Backend Environment (for server admin)

```markdown
* Installation Method or Hosting Service: any — reproducible on bare metal, systemd, pm2, docker, k8s
* Misskey: 2026.5.2-alpha.0 (verified against develop @ e233556700)
* Node: 22.15.0 (.node-version) — affects any Node >= 15
* PostgreSQL: N/A — bug is in the fastify lifecycle layer, not the DB layer
* Redis: N/A
* OS and Architecture: reproducible on darwin-arm64 (macOS 25.3.0); platform-agnostic
```

Patch is ready locally (~30 LOC, single file, plus a CHANGELOG bullet). Awaiting triage / design-and-policy review per CONTRIBUTING.md "Recommended discussing before implementation" before opening the PR. Will reference this issue in the PR. Happy to adjust scope based on triage feedback — e.g., if the 5-second `dispose()` timeout should be configurable rather than hardcoded, or if the layering (single helper vs. extracted method) should change.

Related historical issue: #10995 (closed — same shutdown-hang category, different root cause). Not a duplicate of #16651 (different root cause — that issue is about stale dev processes).

calebcgates:
## What

Closes #17399.

`packages/backend/src/server/ServerService.ts` has three error-handling regressions around `fastify.listen()`, `fastify.ready()`, and `fastify.close()` that cause startup and shutdown failures to be reported as uncaught Node exceptions instead of through the misskey logger the code clearly intended to use.

The four changes:

1. `await fastify.listen()` inside a `try/catch` instead of fire-and-forget on the TCP path and callback-with-ignored-`err` on the socket path.
2. Run `fs.chmodSync` on the socket only **after** `listen()` resolves (today the chmod runs even when listen failed, throwing `ENOENT` and masking the original error).
3. Wrap `await fastify.ready()` in the same `try/catch` so plugin-registration timeouts (`ERR_AVVIO_PLUGIN_TIMEOUT`) are reported through `this.logger.error` instead of NestJS's bootstrap stack trace.
4. Cap `fastify.close()` at 5s in `dispose()` so leaked WebSocket upgrades from `streamingApiServerService.attach(fastify.server)` don't hang `OnApplicationShutdown` until PM2/systemd `SIGKILL`s the process.

The intent of the existing `fastify.server.on('error', ...)` block is preserved — moved into a `handleListenError()` helper invoked from the new `try/catch`. The reason this matters: fastify v5's `listen()` rejects its returned Promise when the underlying http.Server hits `EADDRINUSE`/`EACCES`, before (or in a race with) the http.Server emitting `'error'`. On Node >= 15 with default `--unhandled-rejections=throw`, the listen-promise rejection terminates the process with a stack trace before `server.on('error')` can run. The handler is reachable today only by accident of timing.

## Why

Concrete reproductions:

**Repro 1 — TCP port in use:**

```sh
# Terminal 1
nc -l 3000

# Terminal 2 (Misskey configured with `port: 3000`)
pnpm start
# Today: UnhandledPromiseRejectionWarning + stack trace, no friendly message
# After this PR: "Port 3000 is already in use by another process." + clean exit(1)
```

**Repro 2 — Unix socket without permission:**

```yaml
# default.yml
socket: /var/run/misskey.sock
chmodSocket: '777'
```

Run as a non-root user. `fastify.listen({ path })` rejects with `EACCES`, the callback's `err` is ignored, then `fs.chmodSync('/var/run/misskey.sock', ...)` throws `ENOENT: no such file or directory` because no socket was ever created. User sees the chmod error, not the actual permission error. After this PR: `"You do not have permission to listen on /var/run/misskey.sock."` + clean exit(1).

**Repro 3 — `OnApplicationShutdown` hangs:**

A WebSocket client opens a connection via `streamingApiServerService.attach(fastify.server)`, then disconnects abruptly without a clean close frame. PM2 / systemd / k8s sends SIGTERM to misskey. `dispose()` calls `await this.#fastify.close()` which waits indefinitely for the upgraded socket to be reaped. PM2's default `kill_timeout` then SIGKILLs the process; logs show "process killed" with no diagnostic. After this PR: 5-second cap, then `OnApplicationShutdown` resolves cleanly so PM2 records a graceful exit.

These are the only four real bugs in this area. The other fastify call sites in the backend (route handlers throwing in `ClientServerService` / `ApiServerService` / `WellKnownServerService`, `register()` calls deferred to `await fastify.ready()`, `addHook()` callbacks) are correctly handled by the existing `setErrorHandler` (`ClientServerService.ts:924`) plus the now-properly-`try`-wrapped `await fastify.ready()` — they're not part of this PR's scope.

### Behavior preservation

| Scenario | Old | New |
|---|---|---|
| Healthy TCP startup | works | works |
| Healthy socket startup | works | works (chmod still runs after listen) |
| `EADDRINUSE` | unhandled rejection trace | logger.error + exit(1) |
| `EACCES` | unhandled rejection trace | logger.error + exit(1) |
| Cluster-worker listen failure | `process.send('listenFailed')` only when `server.on('error')` wins the race | `process.send('listenFailed')` always |
| Plugin timeout in ready() | NestJS bootstrap trace | logger.error + exit(1) |
| Graceful shutdown w/ idle conns | works | works |
| Graceful shutdown w/ leaked WS | hangs until SIGKILL | resolves at 5s |

### On tests

`packages/backend/test/unit/` has no `ServerService` spec today. Adding one requires mocking the NestJS DI graph for ~18 injected services and stubbing 11 sub-fastify-services. I scoped this PR to the bug-fix and the truth table above; happy to add a vitest unit test for `handleListenError` in a follow-up if useful — please flag if you want it bundled here instead.

### Local verification

- `pnpm install --frozen-lockfile` — no lockfile churn
- `pnpm --filter backend typecheck` (tsgo) — 0 errors
- `pnpm --filter backend exec eslint src/server/ServerService.ts` — 0 errors; warning count net-reduced from 9 to 5 (removed `(err as any)` cast, two unused-param warnings, and the `this.config.socket!` non-null assertion)

## Checklist

- [x] Update CHANGELOG.md
- [ ] (If possible) Add tests — see "On tests" above; offered as follow-up


syuilo:
🙏🏻

PR #17585, opened 2026-06-19
fix: the script contains locale json is prefetched

anatawa12:
## What

This commit upgrades rolldown used by vite to 1.1.0 and set
includeDependenciesRecursively instead of maxSize for
i18n code splitting group.

This the bug prefetches script file includes locale JSON.

## Why

We unexpectedly prefetched the script file includes locale JSON
before this fix because locale inliner did not remove prefetch
for transitive dependency of i18n global variable.

Current locale inliner assumes the file contains i18n global
variable and the file contains i18n global variable are same file
and only removes prefetch for the file for i18n global variable
and leaves dependency files of the file.
However, in the previous fix for rolldown migration regression,
we set `maxSize: 1` for manual chunk of i18n.
This makes the chunk for i18n global variable (@/i18n.js) and
the chunk includes locale JSON (@@/js/locale.js) distinct chunks.
As a result, only prefetch for i18n global is removed and local
JSON remain in the prefetch file name dictionary (__vite__mapDeps).

There is two ways to fix this problem: 1) make rolldown to bundle
i18n related files into one but leave unrelated files separated
module or 2) update locale inliner to remove transitive dependency
of i18n of __vite__mapDeps.
2nd way is prune to rolldown changes, and it's possible by parsing
each .js file to (re)create module graph in inliner, it's complex.
Therefore, this commit fixes this with 1st way with
includeDependenciesRecursively option on `codeSplitting.groups`
newly added in rolldown 1.1.0.

## Additional info (optional)

Since latest vite as of writing (8.0.16) strictly depends on
rolldown 1.0.3, we cannot use rolldown 1.1.0  normally. We use overrides
to work around this problem. As far as I checked the vite
repository, upgrading rolldown to 1.1.x includes no code changes
except for package.json, so I hope this upgrade is safe.

## Checklist
- [ ] (If possible) Add tests

### DIFF
--- a/packages/frontend-embed/vite.config.ts
+++ b/packages/frontend-embed/vite.config.ts
@@ -153,11 +153,10 @@ export function getConfig(): UserConfig {
 							name: 'vue',
 							test: /node_modules[\\/]vue/,
 						}, {
-							// split each i18n related module to each distinct module, deny hoisting
+							// split i18n related module to distinct module
 							name: 'i18n',
-							test: /i18n\.ts/,
-							minSize: 0,
-							maxSize: 1,
+							includeDependenciesRecursively: false,
+							test: /i18n\.ts|locale\.ts/,
 						}],
 					},
 					entryFileNames: `scripts/${localesHash}-[hash:8].js`,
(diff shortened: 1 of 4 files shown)


syuilo:
🙏

PR #17664, opened 2026-07-04
fix(frontend): correctly detect isMe in mention

Johann150:
## What
Mentions are supposed to be differently formatted when the mention is of the logged in user that is reading the note. This is done using the `isMe` variable. This was not working because in the comparison only one part was converted to lower case.

previous behaviour (not working with upper case):
grafik

As you can see only the lower case spelling is slightly highlighted with a different colour. The upper case spelling is shown in the same colour as all other mentions.

## Why
The intention seems obvious from the code but it only works when the mentioned user name in the note is written in lower case.

## Checklist
- [ ] (If possible) Add tests


kakkokari-gtyih:
:pray:

syuilo:
🙏

PR #17927, opened 2026-09-06
fix: make `isVisibleForMe` not require relations

anatawa12:
## What

Actually, `note.reply.userId` check can be covered by `note.mentions` since `NoteCreateService` adds reply user to `note.mentions`, so there is no behavioral changes.
Therefore, I don't think I should add CHANGELOG entry.

## Why

いくつか`getNote`の結果のみを使用している箇所があったため、それに合わせています

## Checklist
- [ ] Read the [contribution guide](https://github.com/misskey-dev/misskey/blob/develop/CONTRIBUTING.md)
- [ ] (If possible) Add tests


syuilo:
👍🏻

PR #17941, opened 2026-09-12
fix: include query string in HTTP Signature (request-target)

### LINKED ISSUE #17734: `(request-target)` value in HTTP Signatures silently drops query string
Misskey incorrectly leaves out part of the URI in HTTP Signatures, only keeping the path and ignoring the query string:
https://github.com/misskey-dev/misskey/blob/0be4d352ec213b1f23977b4169f50a1149d104f2/packages/backend/src/core/activitypub/ApRequestService.ts#L123

This is an issue Mastodon had as well and fixed a while ago (while still temporarily accepting incorrect signatures). Background: https://github.com/mastodon/mastodon/pull/39516

KayanoLiam:
## What
Include the URL query string in the HTTP Signatures `(request-target)` pseudo-header when signing ActivityPub GET/POST requests.

`ApRequestCreator.#genSigningString` previously used only `URL.pathname`, so a request to `/users/alice?page=2` was signed as `/users/alice`.

## Why
Draft HTTP Signatures define `(request-target)` as the HTTP method plus the request-target (path **and** query). Implementations that verify the query string (including Mastodon) reject Misskey's signatures for URLs that have a query.

Fixes #17734

## Additional info (optional)
- Fragment identifiers are still omitted (`URL.search` does not include the hash).
- URLs without a query keep the previous `pathname`-only value (no trailing `?`).
- Unit tests: `NODE_ENV=test pnpm exec vitest --config vitest.config.unit.ts run test/unit/ap-request.ts` → 10 passed.

## Checklist
- [x] (If possible) Add tests

### DIFF
--- a/packages/backend/test/unit/ap-request.ts
+++ b/packages/backend/test/unit/ap-request.ts
@@ -66,6 +66,57 @@ describe('ap-request', () => {
 		assert.deepStrictEqual(result, true);
 	});
 
+	test('createSignedGet includes query string in request-target', async () => {
+		const keypair = await genRsaKeyPair();
+		const key = { keyId: 'x', 'privateKeyPem': keypair.privateKey };
+		const url = 'https://example.com/users/alice?page=2';
+
+		const req = await ApRequestCreator.createSignedGet({ key, url, additionalHeaders: { 'User-Agent': 'UA' } });
+
+		assert.ok(req.signingString.split('\n').includes('(request-target): get /users/alice?page=2'));
+
+		const parsed = buildParsedSignature(req.signingString, req.signature, 'rsa-sha256');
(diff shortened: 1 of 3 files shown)


kakkokari-gtyih (commented):
見た感じ大丈夫そうなんですがちょっと自信が無いので他の方にレビューお願いしたい気持ち

packages/backend/test/unit/ap-request.ts
anatawa12, line 1: Can you add test case with ? as a trailing character without any hash component please?
WHATWG URL removes "?" when hash component is empty, but fetch with string and the RFC 3986 doesn't.

anatawa12 (approved):
LGTM

